PR #200, opened 2015-01-19
Changes fixing issues in GCE

rvansa:

PR #201, opened 2015-01-21
MapReduce and DistExec trait refactored

rvansa:
At this point this is rather a preview PR, as it's based on #200

### DIFF
--- a/core/src/main/java/org/radargun/stages/distributedtask/DistributedTaskStage.java
+++ b/core/src/main/java/org/radargun/stages/distributedtask/DistributedTaskStage.java
@@ -3,6 +3,7 @@ package org.radargun.stages.distributedtask;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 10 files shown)


reporters/reporter-default/src/main/java/org/radargun/reporting/html/TimelineChart.java
```diff
@@ -88,15 +89,30 @@ public void setEvents(List<? extends Timeline.Event> events, int slaveIndex, lon
       plot.setRangeGridlinesVisible(false);
       plot.setRangeZeroBaselineVisible(true);
 
-      long lastTimestamp = Long.MIN_VALUE;
+      Number minValues[] = new Number[MAX_EVENT_VALUES];
+      Number maxValues[] = new Number[MAX_EVENT_VALUES];
+      long minTimestamps[] = new long[MAX_EVENT_VALUES];
+      long maxTimestamps[] = new long[MAX_EVENT_VALUES];
+      //long lastTimestamp = Long.MIN_VALUE;
```
mcimbora: At last something! :-) Remove the comment please.

rvansa:
Rebased on master

alanfx:
@rvansa I sent a PR to your repo: https://github.com/rvansa/radargun/pull/6

There are basically two tricky things in the Map/Reduce implementation.

1) Storing the results of the Map/Reduce task when the trait doesn't support it. There was a bug in the old code. The storeResultInCache flag says that the results should be stored, and the supportsResultCacheName flag says if the trait supports this. I had to kludge this in your code to pass the name of the cache to supportsResultCacheName() method. The name of the cache has to be registered to the service in the trait, so that it can be accessed in the stage later. If you have a better idea, let me know.

2) The other tricky part is printing the results of the Map/Reduce task, if the results are stored in the cache. I decided to implement this using the Iterable trait with an optional dependency, but this will only work in Infinispan 7 and JDG 6.4. The other option would be to add the keySet() to the Cache API, but I'm not sure that's consistent with the JCache API. Once again, let me know what you think.

rvansa:
@alanfx thanks, for the update:

1) Do you need to store the results there at all? Do you reuse them in later stages? If setting result cache is not supported, the stage with this setting on can fail/issue a warning, but I don't see the need to put it there.
I definitely do not like adding the result cache name to supportsX() method - it a) looks as if the support depends on cache name b) has an unexpected side effect. If you can't do cache.put() to a cache that was not defined beforehead, I'd rather add method boolean exists(String cacheName) to BasicOperationsTrait, and let this have the side efect of starting cache (and always return true in Infinispan). And fail the stage in case that exists() returns false.

2) I don't see any problem in using optional trait. Especially when it's due to logging - if you can't enumerate the entries, you can't, period. I would truly object against keySet() method, it was not working properly in previous versions of Infinispan either. And it's conceptually broken, of course.
However, the handling of NoSuchMethodError looks wrong - if the Iterable trait is not supported, Iterable will be set to null and you should handle this. If NoSuchMethodError is thrown, there is a bug in this trait's implementation (or rather you're using Infinispan 7.0 trait and the API looks differently for JDG 6.4, or vice versa).

rvansa:
@alanfx Btw., it would be quite useful if you uploaded benchmark that I could test the changes - the example benchmark-mapreduce .xml is quite minimalistic.

alanfx:
Hey @rvansa 

1) My thought was that it would be possible to additional analysis on the results in the cache. Either running an additional M/R task on the result map or something else. At this point, I have written any benchmarks that take advantage of this yet though. What about adding a registerCache(String cacheName) to the BasicOperationsTrait? Then I could only register the cache if storeResultInCache is true.

2) The Iterable trait is not null in JDG 6.3 where I tried this which is why I added the catch clause for NoSuchMethodError. I'll look into the trait implementation to see what I can find.

I'll also push the benchmark file I used to my branch. It includes configurations for ISPN 6, 7, JDG 6.3, and 6.4.

rvansa:
@alanfx 

1) Let's postpone any such changes until we know what will we try to achieve, and have a better picture what needs to be done. You can always add new stuff, but tidying up hacks is nasty business. WDYT?

2) So you'll need JDG63Iterable, probably.

Thanks

alanfx:
@rvansa 

1) I want to think about this some more. I think it is pretty useful to register a cache to an embedded service without having to add the definition to every configuration file.

2) This is strange. In the infinispan-commons JAR included with 6.4, the iterator method is defined on  CloseableIterable, but that's not the case in 6.3. I need to figure out if it is even possible to create an iterator in 6.3.

Thanks

rvansa:
1) OK. I am also considering whether Lifecycle.startComponent(String componentName) mapping to CacheManager.startCache() wouldn't be a better place rather than BasicOperations which are dedicated to the JCache-like operations themselfves.

rvansa:
And one more note - it would be probably better to issue a PR directly to main RG repo, rather than to my private repo (where is that base PR stored).

alanfx:
Closing this PR in favor of #202

PR #202, opened 2015-01-27
MapReduce and DistExec traits refactored

alanfx:
Started with @rvansa PR, and then made some additional changes


rvansa:
Looks good, integrated.

PR #206, opened 2015-01-30
Fixes for Random data, Map/Reduce, and Distributed Task reporting

alanfx:
DataThroughput
- Fix calculation when small values are used (No more Infinite
Throughput!)
- Store totalBytes to output to reports
- Generate array with only non-zero response times

CsvReporter
- Output DataThroughput values in CSV reports

RandomDataStage
- Report bytes written per node
- Report count of words and target memory per node
- Provide aggregated values

DistributedTaskStage
- Add property to retrieve the total bytes written to the cluster in
case RandomDataStage isn't used
- Report distributed task durations

MapReduceStage
- Add property to retrieve the total bytes written to the cluster in
case RandomDataStage isn't used
- Change aggregated values and result title

DistributedTaskExecutor
- Add operation for statistics gathering

Utils
- Make invokeMethodWithString use generic type instead of Object

Infinispan70EmbeddedService
- Remove unused imports

### DIFF
--- a/core/src/main/java/org/radargun/stats/representation/DataThroughput.java
+++ b/core/src/main/java/org/radargun/stats/representation/DataThroughput.java
@@ -15,12 +15,16 @@ public class DataThroughput {
    public final double maxThroughput;
    public final double meanThroughput;
    public final double deviation;
+   public final long[] responseTimes;
+   public final long totalBytes;
 
-   public DataThroughput(double min, double max, double mean, double deviation) {
+   public DataThroughput(double min, double max, double mean, double deviation, long totalBytes, long[] responseTimes) {
       this.minThroughput = min;
       this.maxThroughput = max;
       this.meanThroughput = mean;
       this.deviation = deviation;
+      this.responseTimes = responseTimes;
+      this.totalBytes =totalBytes;
    }
 
    public static DataThroughput compute(long totalBytes, long[] responseTimes, int requestCount) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/radargun/stats/representation/DataThroughput.java
```diff
@@ -28,24 +32,24 @@ public static DataThroughput compute(long totalBytes, long[] responseTimes, int
...
+         minDurationSecs = toSecs(minDurationSecs);
+         maxDurationSecs = toSecs(maxDurationSecs);
          meanDurationSecs = toSecs(meanDurationSecs / requestCount);
       }
       return new DataThroughput(totalBytes / maxDurationSecs, totalBytes / minDurationSecs, totalBytes
-            / meanDurationSecs, getDeviation(throughputs));
-   }
-
-   private static double toSecs(long nanos) {
-      return nanos / TimeUnit.SECONDS.toNanos(1);
```
rvansa: Nanoseconds are converted to seconds rather with TimeUnit.NANOSECONDS.toSeconds(nanos) :)

core/src/main/java/org/radargun/utils/Utils.java
```diff
@@ -442,7 +442,7 @@ public static void createOutputFile(String fileName, String fileContent, boolean
     *           a Map where the public method name is the key, and the String parameter is the value
     * @return the modified Object, or <code>null</code> if the field can't be changed
     */
-   public static Object invokeMethodWithString(Object object, Map<String, String> properties) {
+   public static <T> T invokeMethodWithString(T object, Map<String, String> properties) {
```
rvansa: Is there any reason for returning the object at all? Moreover, I think that the name of the method is rather misleading - it should be invokeMethod*s*WithString.

rvansa:
Rebased & Integrated, thanks.

PR #207, opened 2015-02-02
Code cleaning - removed redundant variables, unused imports, etc.

mcimbora:


core/src/main/java/org/radargun/config/ConfigSchemaGenerator.java
```diff
@@ -63,7 +63,6 @@ protected void generate() {
       addAttribute(masterComplex, ATTR_PORT, intType, null, false);
 
       Element clustersChoice = createChoice(benchmarkSequence, 0, 1);
-      Element localComplex = createComplexElement(clustersChoice, ELEMENT_LOCAL, 0, 1);
```
rvansa: createComplexElement has a side-effect. You probably want just to remove the unused variable. The same for scenarioType.

core/src/main/java/org/radargun/config/Stage.java
```diff
@@ -15,7 +15,7 @@
    static final String NO_DEPRECATED_NAME = "__no_deprecated_name__";
    String name() default CLASS_NAME_WITHOUT_STAGE;
    String deprecatedName() default NO_DEPRECATED_NAME;
-   String doc();
+   String doc() default "";
```
rvansa: Not having a default value is intended, in order to force you to write some documention.

core/src/main/java/org/radargun/stages/AbstractDistStage.java
```diff
@@ -26,7 +26,7 @@
  *
  * @author Mircea Markus &lt;Mircea.Markus@jboss.com&gt;
  */
-@Stage(doc = "")
```
rvansa: The documentation for this stage won't be ever seen in schema, therefore it's empty.

core/src/main/java/org/radargun/stages/cache/IsolationLevelCheckStage.java
```diff
@@ -81,6 +81,7 @@ public DistStageAck executeOnSlave() {
       try {
          Thread.sleep(duration);
       } catch (InterruptedException e) {
+         log.error("Thread has been interrupted", e);
```
rvansa: You should probably restore interrupt status as well.

plugins/infinispan51/src/main/java/org/radargun/service/Infinispan51Operations.java
```diff
@@ -74,7 +74,7 @@ public boolean remove(K key) {
       @Override
       public V getAndRemove(K key) {
          boolean startedTx = startTxAndLock(key);
-         boolean retval = super.remove(key);
```
rvansa: Seems like you found a bug - you should not call both super.remove and super.getAndRemove. Could you fix that? Also, I think that stopTx should be called in finally block.

rvansa:
And this will need a rebase, too.

mcimbora:
@rvansa updated

PR #210, opened 2015-02-13
Add jdg65 plugin

mcimbora:


mcimbora:
@alanfx Fixed

alanfx:

rvansa:
@alanfx @mcimbora Hi guys, please don't set SNAPSHOT or any non-publicly available version of JDG as the default version for plugin. Plugins with no public release should have their own profile in aggregator pom.xml, and this profile should be activated by setting the version with which you want to compile the plugin.

PR #211, opened 2015-02-26
Add ability to set JMX attributes and use clustered script to start server service

alanfx:


core/src/main/java/org/radargun/stages/JMXInvocationStage.java
```diff
@@ -107,6 +113,11 @@ public DistStageAck executeOnSlave() {
             } else if (operationType == OperationType.GET_ATTRIBUTE_VALUE) {
                log.trace("Getting value of attribute " + targetName);
                result = connection.getAttribute(objectInstance.getObjectName(), targetName);
+            } else if (operationType == OperationType.SET_ATTRIBUTE_VALUE) {
+               log.trace("Setting value of attribute " + targetName + " to " + methodParameters.get(0).getElementValue().toString());
```
mcimbora: I believe this will end up with NPE when 'methodParameters' is not specified. Could you please add a check for the list size (!= 0) when setting attribute value?

plugins/infinispan52/src/main/java/org/radargun/service/InfinispanServerService.java
```diff
@@ -159,7 +159,7 @@ public InfinispanServerClustered getClustered() {
    @Override
    protected List<String> getCommand() {
       ArrayList<String> command = new ArrayList<String>();
-      command.add(FileSystems.getDefault().getPath(home, "bin", "standalone." + (windows ? "bat" : "sh")).toString());
+      command.add(FileSystems.getDefault().getPath(home, "bin", "clustered." + (windows ? "bat" : "sh")).toString());
```
rvansa: Could you comment why are you changing this?

alanfx: Sure @rvansa, I can see that the standalone and clustered scripts are basically identical, but I think it is important that the servers use the default clustered configuration if the configuration file is not specified in the benchmark.

core/src/main/java/org/radargun/stages/JMXInvocationStage.java
alanfx: Good catch @mcimbora. I just added a null check for methodParameters

mcimbora: @alanfx Unfortunately this still leads to exception being thrown :-) Fixed locally

mcimbora:
Integrated, thanks!

core/src/main/java/org/radargun/stages/JMXInvocationStage.java
alanfx: Thanks @mcimbora and sorry for my denseness!

PR #212, opened 2015-03-19
Include remote listener trait provider into JDG

vjuranek:


vjuranek:
PR updated, thanks @rvansa  for point out that previous one was wrong

rvansa:
Thanks, integrated!

PR #213, opened 2015-03-23
Fix backward incompatible JGroups DISCARD.setDiscardAll invocation

mcimbora:


alanfx:
Integrated. Thanks, @mcimbora !

PR #214, opened 2015-03-26
Allow measure latency between sending and receiving cache events

vjuranek:


mcimbora:
@vjuranek Integrated, thanks!

PR #215, opened 2015-03-27
Fixed bug causing infinite loops

rvansa:


mcimbora:
@rvansa Works fine, thanks. However, if 'name' attribute is not specified in the outer loop, NPE is thrown in the last iteration.

See https://gist.github.com/mcimbora/b1e237a9938377f647cd

rvansa:
@mcimbora Does not work fine, then :) Updated.

core/src/main/java/org/radargun/stages/control/RepeatEndStage.java
```diff
@@ -26,17 +26,19 @@ private void updateState(StateBase state) {
...
       state.remove(counterName);
       Stack<String> repeatNames = (Stack<String>) state.get(RepeatStage.REPEAT_NAMES);
-      if (repeatNames == null || repeatNames.isEmpty() || repeatNames.peek() != name) {
+      if (repeatNames == null || repeatNames.isEmpty() || !equals(repeatNames.peek(), name)) {
          throw new IllegalStateException("Cannot unwind repeat: " + repeatNames);
       }
       repeatNames.pop();
    }
 
+   private static boolean equals(String s1, String s2) {
```
mcimbora: Wait, this is from Objects.equals, why not using it right away when we support 1.7? :-)

mcimbora: Wait, this is from Objects.equals, why not using it right away when we support 1.7? :-)

mcimbora:
@rvansa Integrated, thanks!

PR #216, opened 2015-03-27
Fix remote listener stats input - should be in nanos

vjuranek:


mcimbora:
@vjuranek Integrated, thanks!

PR #217, opened 2015-03-31
Fix cluster formation issue caused by stale member list

mcimbora:


alanfx:

PR #218, opened 2015-03-31
Add support for near caching

vjuranek:
Near caching was added in ISPN 7.1, therefore adding whole new plugin `infinispan71`


vjuranek:
PR updated with near caching support in JDG 6.5

rvansa:
Thanks, integrated!

alanfx:
Hey @vjuranek and @rvansa, wouldn't it make more sense to bump the version of the infinispan70 plugin to 7.1.1.Final instead of adding a infinispan71 plugin?  Sorry I didn't ask about this earlier, but just trying to understand the thought behind this.

rvansa:
We have plugins for each minor version of Infinispan and JDG, I don't see why this should be different. We want to gradually add features and be able to test 7.0 vs 7.x. However, we've already discussed with @mcimbora that we could merge older versions (Infinspan 5.x) in order to reduce build times and distribution size.
@alanfx do you think that we should merge any minor versions of JDG, too?
We could also just add an option to pack only the most recent version of each project/product line (then we would have all those modules and have to build them, but the zip file wouldn't be as fat).

alanfx:
Yes, @rvansa I can see that new plugins for minor versions has been the case in the past. I think I was just confused because infinispan60 never got a minor version release. I think this makes sense, but I'll just have to move some code I was working on!

I was also talking to @mcimbora about merging versions, but I'm not sure it is worth the effort. The builds run pretty fast at this point, and keeping track of which versions you have or don't have seems like unnecessary overhead.

PR #220, opened 2015-04-02
Allow register listener stage to reset statistics

vjuranek:


mcimbora:
@vjuranek Looks good, integrating. Thanks!

PR #222, opened 2015-04-08
Background ops fixes, improve trace logging

mcimbora:
* Avoid loading caches if the service is not running
* Allow to scan for ignored keys also for non-zero operation ids (when StressorRecordPool survives service restart)
* Improve trace logging


alanfx:
Integrated. Thanks, @mcimbora

Issue #223, opened 2015-04-09
NPE when reading config file


vjuranek:
RG tries to read original server config file and fails with NPE while doing so:
```
07:58:02,532 ERROR [org.radargun.service.ServerConfigurationProvider] (sc-main) Error while reading configuration file (/standalone/configuration/st-indexed-sync.xml)
java.lang.NullPointerException
        at org.radargun.utils.Utils.readAsBytes(Utils.java:189)
        at org.radargun.service.AbstractConfigurationProvider.loadConfigFile(AbstractConfigurationProvider.java:39)
        at org.radargun.service.AbstractConfigurationProvider.getOriginalConfigs(AbstractConfigurationProvider.java:28)
        at org.radargun.stages.lifecycle.ServiceStartStage.executeOnSlave(ServiceStartStage.java:82)
        at org.radargun.SlaveBase.scenarioLoop(SlaveBase.java:86)
        at org.radargun.SlaveBase$ScenarioRunner.run(SlaveBase.java:150)
```

vjuranek:
Should be fixed by cfafc21ce76221a5bf89

PR #224, opened 2015-04-09
#223 Read config file directly from filesystem when not present as resou...

vjuranek:
...rce


alanfx:
Integrated. Thanks @vjuranek

PR #225, opened 2015-04-09
Update code to detect cache rehashing in Infinispan Server Service

alanfx:
This change checks the pendingViewAsString attribute of the RpcManager
component associated with each cache using JMX. If pendingViewAsString
equals "null", no rehash is in progress.

### DIFF
--- a/plugins/infinispan60/src/main/java/org/radargun/service/Infinispan60ServerTopologyHistory.java
+++ b/plugins/infinispan60/src/main/java/org/radargun/service/Infinispan60ServerTopologyHistory.java
@@ -1,77 +1,58 @@
 package org.radargun.service;
 
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
 
-/**
- */
+import javax.management.MBeanServerConnection;
+import javax.management.ObjectName;
+
 public class Infinispan60ServerTopologyHistory extends AbstractTopologyHistory {
-   protected final static String ALL_CACHES = "__all_caches__";
-
    protected final InfinispanServerService service;
 
-   /**
-    * Requires logging:
-    * TRACE on org.infinispan.statetransfer.StateConsumerImpl,
-    * TRACE on org.infinispan.statetransfer.StateTransferManagerImpl
-    * DEBUG on org.infinispan.transaction.impl.TransactionTable
-    */
+   protected final static String ALL_CACHES = "__all_caches__";
+
+   protected final static String JMX_CACHE_COMPONENT = "%s:type=Cache,name=\"*\",manager=\"clustered\",component=%s";
+   // Topology change strings
+   protected final static String JMX_RPC_MANAGER = "RpcManager";
+   protected final static String JMX_PENDING_VIEW_ATTR = "pendingViewAsString";
+   // Rehash strings
+   protected final static String JMX_STATE_TRANSFER_MANAGER = "StateTransferManager";
+   protected final static String JMX_STATE_TRANSFER_IN_PROGRESS_ATTR = "stateTransferInProgress";
+
+   protected final Map<String, CacheStatus> cacheChangesOngoing = new HashMap<>();
+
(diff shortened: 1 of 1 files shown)


mcimbora:
@alanfx Looks good to me. If there are no additional comments today, I'll go ahead and integrate it.

rvansa:
Yep, this could work, though I miss any such handling for topology. Also, check StateTransferManager.isStateTransferInProgress. Maybe you could issue Infinispan PR with managed attributes on StateTransferManager that would expose topology information, too?

alanfx:
Thanks, @mcimbora. @rvansa yes, this PR only handles rehashing. I have been talking to @danberindei about this. He is talking about exposing the "highest topology id received from the coordinator" and "highest topology id that it finished installing" as JMX attributes, but this will have to happen in the Infinispan 7.1 or 7.2 timeframe. Those two attributes could be used to determine if a topology change is in progress.

alanfx:
The new commit adds per cache booleans as well as keeping track of a rehash happening on any cache

alanfx:
OK, don't integrate this PR. The topology check is not correct right now, because the number of log messages for the start of a topology change is not balanced by the same number of log messages for the end of a topology change. I think I have an idea for using some other JMX attributes that I am talking to Dan about.

alanfx:
@mcimbora and @rvansa, ok this new commit uses JMX to detect topology changes and rehashes. I decided that the pending view attribute on the RpcManager was a better signal for a topology change, and then I found some attributes on the StateTransferManager that made sense to detect a rehash. I am waiting to hear if @danberindei agrees with me, and I am running more tests. Feel free to have a look and let me know what you think. It seems better to me to use existing attributes that we can write bugs against, instead of adding new attributes.

plugins/infinispan60/src/main/java/org/radargun/service/Infinispan60ServerTopologyHistory.java
```diff
@@ -80,4 +88,155 @@ public void afterStop(boolean graceful) {
...
+            List<Attribute> result = connection.getAttributes(managerName, JMX_REHASH_ATTRS).asList();
+            if (result.get(0).getValue().equals(false) || result.get(1).getValue().equals(true)) {
+               log.info(String.format("Rehash in progress. %s = %s and %s = %s", JMX_JOIN_COMPLETE_ATTR, result.get(0)
+                     .getValue(), JMX_STATE_TRANSFER_IN_PROGRESS_ATTR, result.get(1).getValue()));
+               rehashMap.put(rehashCacheNameMap.get(managerName.toString()), true);
+               rehashesInProgress++;
+            } else {
+               log.debug("No rehash in progress");
+               rehashMap.put(rehashCacheNameMap.get(managerName.toString()), false);
+            }
```
mcimbora: Looks like topologyChangeInProcess & rehashInProcess contain some amount of duplicate code. Maybe you could find a way how to refactor it? (e.g. introducing a common template and plugging in specific behavior via dedicated object)

alanfx: Yes, this could certainly benefit from lambdas, but first I want to see if this will work at all! I'll try to see how much I can parameterize it.

alanfx:
OK, this latest commit includes an important fix to update the map as changes come in. I also collapsed the code into a single Runnable. There is still some code duplication, but i thought this made sense in this circumstance. I have verified that the resilience tests aren't failing on the WaitForTopologySettleStage with these changes.

rvansa:
@mcimbora @alanfx Speaking generally... since we can't have unit tests for most of this stuff, I'd recommend to be rather reluctant to file PRs with fresh code, unless someone needs it. Let's work on private devel branches for a while (e.g. 2 weeks), and file PRs after you've tested it in more scenarios (and apply fixes through interactive rebase). This saves time to both reviewer and committer.

rvansa:
I've created a label 'preview' for PRs which you want to show for comments but don't feel certain about it.

alanfx:
@rvansa I think that approach makes sense to me. Maybe add another label that says "Ready to integrate" or "Tested"?

rvansa:
@alanfx Let's make it default that if PR is waiting, it's tested and ready to be integrated :)

alanfx:
@rvansa OK, so removing the "Preview" label would indicate the same thing. Works for me!

mcimbora:
@alanfx Integrated, thanks!

PR #226, opened 2015-04-09
Add version.infinispan.public property to parent POM

alanfx:
This version of the infinispan-distexec-demo JAR is used as a dependency
in the JDG plugins
Also added the infinispan-distexec-demo profile to the infinispan71
plugin

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -250,5 +250,7 @@
    <properties>
       <!-- By default don't run unit tests for the fwk! -->
       <maven.test.skip>true</maven.test.skip>
+      <!-- Use this version of infinispan-distexec-demo in the JDG plugins -->
+      <version.infinispan.public>7.1.1.Final</version.infinispan.public>
    </properties>
 </project>
--- a/plugins/infinispan71/pom.xml
+++ b/plugins/infinispan71/pom.xml
@@ -119,6 +119,28 @@
             <version.infinispan>7.1.1.Final</version.infinispan>
          </properties>
       </profile>
+      <profile>
+         <id>demos</id>
+         <activation>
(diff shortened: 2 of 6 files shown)


rvansa:
Why is that universally in parent/pom.xml? It seems to me that this property should stay in each plugin, since the version of demo can be non-compatible.

alanfx:
Actually, the same version of the demo code is compatible, because the mapper, reducer, collator, and executor signatures have not changed. This change just allows me to use the latest version of the demo code by changing the value in one place. The infinispan## plugins will still use the version of the demo code that they shipped with, though.

rvansa:
I am still reluctant to integrate this, even though it 'works', changing parent/pom.xml is not the right thing to do. If you want to override that version often for all modules, add the default version as profile, and update your build stript to define the version (as we do e.g. for build from snapshots).

alanfx:
Not sure why there are scare quotes around works. It seems to me that adding a property in parent/pom.xml is exactly the place to define a property used by multiple sub-projects and to centralize maintenance. I also don't see how adding a profile that is required to build is helpful either, so I'll just send a pull request to update all of the POMs separately.

rvansa:
I was talking about profile since I haven't found a way to define default value for properties in Maven (the ${foo:default-foo} syntax does not work in Maven), and therefore I had to work that around using profile that's activated when -Dfoo=... is not set.
I see your point in centralizing the maintenance (though I think that compatibility across major versions is rather a coincidence) - what do you think about creating an infinispan-parent module in plugins/ and moving Infinispan under that? (and, in order to have plugins symmetrical, do the same for other plugins as well?) Or, am I overengineering?

PR #227, opened 2015-04-11
Add JCache (JSR-107) plugin

mcimbora:


core/src/main/java/org/radargun/utils/Utils.java
```diff
@@ -249,15 +249,24 @@ public static Properties getPluginProperties(String plugin) {
...
+         log.error(String.format("Exception while loading plugin specific properties file %s.", PROPERTIES_FILE), e);
+      }
+      return properties;
+   }
+
+   public static Properties loadProperties(Properties properties, File file) throws IOException {
+      return loadProperties(properties, new FileInputStream(file));
+   }
+
+   public static Properties loadProperties(Properties properties, InputStream inputStream) throws IOException {
```
rvansa: I think that closing resource in different scope than where it is allocated is an anti-pattern. RG is baselined on Java 7, use try-with-resources instead, please.

plugins/jcache/src/main/java/org/radargun/service/JCacheOperations.java
```diff
@@ -0,0 +1,90 @@
...
+public class JCacheOperations implements BasicOperations, ConditionalOperations {
+
+   private JCacheService service;
+
+   public JCacheOperations(JCacheService service) {
+      this.service = service;
+   }
+
+   @Override
+   public <K, V> Cache<K, V> getCache(String cacheName) {
```
rvansa: So, you're silently ignoring the requested cacheName here? That's not correct. JCache CacheManager seems to support multiple caches.

pom.xml
```diff
@@ -450,7 +464,7 @@
                         </path>
 
                         <echo message="Packaging plugins: " />
-                        <ac:for list="${plugins.chm},${plugins.ehcache},${plugins.hazelcast},${plugins.jbosscache},${plugins.jgroups},${plugins.infinispan},${plugins.jdg},${plugins.coherence},${plugins.process},${plugins.spymemcached}" param="plugin" xmlns:ac="antlib:net.sf.antcontrib">
+                        <ac:for list="${plugins.chm},${plugins.ehcache},${plugins.hazelcast},${plugins.jbosscache},${plugins.jgroups},${plugins.infinispan},${plugins.jdg},${plugins.coherence},${plugins.process},${plugins.spymemcached},${plugins.jcache}" param="plugin" xmlns:ac="antlib:net.sf.antcontrib">
```
rvansa: I don't think you need to add it here, since the plugin will always be used as a dependency to another plugins and does not need to be packaged separately.

plugins/jcache/src/main/java/org/radargun/service/JCacheOperations.java
mcimbora: Oh, good catch

pom.xml
```diff
@@ -289,6 +289,20 @@
             <plugins.spymemcached>spymemcached</plugins.spymemcached>
          </properties>
       </profile>
+      <profile>
```
rvansa: It would be nice if you could trigger plugin compilation by the fact that another plugin that has this as a dependency is compiled. But I don't know if that could be done in any easy way (I think that properties defined in profile cannot activate another profiles).

core/src/main/java/org/radargun/utils/Utils.java
mcimbora: Makes sense. As for the second comment, I'd prefer all-or-nothing strategy.

mcimbora: i.e. either we use it everywhere or nowhere.

plugins/jcache/src/main/java/org/radargun/service/JCacheLifecycle.java
```diff
@@ -0,0 +1,69 @@
...
+      if (service.propertiesFile != null) {
+         configProperties = new Properties();
+         try {
+            Utils.loadProperties(configProperties, JCacheLifecycle.class.getClassLoader().getResourceAsStream(service.propertiesFile));
+         } catch (IOException e) {
+            log.error(String.format("Exception while loading configuration properties file %s.", service.propertiesFile));
+            throw new IllegalStateException(e);
+         }
+      }
+      cacheManager = Caching.getCachingProvider().getCacheManager(configFileUri, JCacheLifecycle.class.getClassLoader(), configProperties);
```
rvansa: Recent Infinispan versions bring JCacheProvider for both embedded and remote, and since there can (and should!) be both of them on classpath, you should add property for selection of the provider.

mcimbora: That's not correct, only one CachingProvider can be present on the classpath at the same time. Otherwise you get "Multiple CachingProviders have been configured when only a single CachingProvider is expected". This will require separate profiles for embedded/remote modules.

pom.xml
mcimbora: Yes, that's remaint of my attempt to make the plugin configurable just by adding provider dependency.

pom.xml
mcimbora: Challenge accepted!

plugins/jcache/src/main/java/org/radargun/service/JCacheLifecycle.java
```diff
@@ -0,0 +1,69 @@
...
+   public JCacheLifecycle(JCacheService service) {
+      this.service = service;
+   }
+
+   @Override
+   public void start() {
+      URI configFileUri = null;
+      try {
+         if (service.configFile != null) {
+            URL configFile = JCacheLifecycle.class.getClassLoader().getResource(service.configFile);
```
rvansa: I don't think that the URI should point to any configuration file - that's what properties are for. URI should identify the cache manager itself.

plugins/jcache/src/main/java/org/radargun/service/JCacheLifecycle.java
rvansa: How foolish to think that! :) That exception is thrown only when you don't specify the provider classname, since it really cannot choose in the no-arg method.

plugins/jcache/src/main/java/org/radargun/service/JCacheLifecycle.java
mcimbora: Then you should send a PR to change this in ISPN :-) (see embedded JCacheManager for reference). It's true different caching providers may specify path to configuration file in a different manner, I'll make it more configurable.

plugins/jcache/src/main/java/org/radargun/service/JCacheLifecycle.java
mcimbora: Ok, you win. This time!

mcimbora:
@rvansa Comments have been addressed. As to https://github.com/radargun/radargun/pull/227#discussion_r28786473, I added TODO for now, as this will require more research.

core/src/main/java/org/radargun/utils/Utils.java
```diff
@@ -249,16 +249,21 @@ public static Properties getPluginProperties(String plugin) {
...
-         throw new RuntimeException(e);
-      } finally {
-         close(inStream);
+         log.error(String.format("Exception while loading plugin specific properties file %s.", PROPERTIES_FILE), e);
       }
+      return properties;
+   }
+
+   public static Properties loadProperties(Properties properties, File file) throws IOException {
+      return loadProperties(properties, new FileInputStream(file));
```
rvansa: you're still not closing the input stream anywhere

mcimbora: Oh, I see. Overlooked that FIS creation :-)

mcimbora:
@rvansa Fixed

rvansa:
I've removed the JDG plugin from benchmark-jcache.xml since users should be able to run this test without productized plugins (with default mvn install).

Integrated.

PR #228, opened 2015-04-16
Add null checks to remove and replace operations

alanfx:


mcimbora:
@alanfx Integrated, thanks!

PR #230, opened 2015-04-21
Reporting updates

rvansa:


alanfx:
Hey @rvansa, can you provide a benchmark file that generates a percentiles chart? I haven't been able to make one yet.

rvansa:
@alanfx just use histogram statistics, percentiles are generated from the same data.

reporters/reporter-default/src/main/java/org/radargun/reporting/html/LogTimeAxis.java
```diff
@@ -0,0 +1,129 @@
...
+         if (labelBounds != null && labelBounds.getWidth()
+               + insets.getTop() + insets.getBottom() > maxHeight) {
+            maxHeight = labelBounds.getWidth()
+                  + insets.getTop() + insets.getBottom();
+         }
+      }
+      return maxHeight;
+   }
+
+   protected double findMaximumTickLabelWidth(List ticks, Graphics2D g2, Rectangle2D drawArea, boolean vertical) {
```
mcimbora: findMaximumTickLabelWidth & findMaximumTickLabelHeight are identical to PercentileLogAxis.findMaximumTickLabelWidth & PercentileLogAxis.findMaximumTickLabelHeight. Perhaps they could be extracted to some other class (e.g. AxisUtils)?

mcimbora:
@rvansa Looks fine. I'm thinking of using freemarker templates for generating HTML pages, which will provide better separation of view & model. I'll have a look at it.

alanfx:
@rvansa I think it would be be better to have an example of the usage of this chart in one of the benchmarks provided with the PR. As I said, I haven't been able to create one myself, so I'd appreciate if you could supply one for my testing! :-)

rvansa:
@mcimbora created ChartsHelper with those two methods
@alanfx added histogram statistics to benchmark-dist.xml, works like charm out-of-the box :-P

mcimbora:
@rvansa Integrated, good job!

PR #231, opened 2015-05-12
Failover testing - fix wrong slave indices and addition of confirmations

mcimbora:
Fix
* faulty slave indices when using multiple groups
* adding confirmations to StressorRecord


alanfx:
I've been running jobs with these fixes for a while. Integrated. Thanks, @mcimbora

PR #232, opened 2015-05-12
Include partition status in topology history

mcimbora:


plugins/infinispan70/src/main/java/org/radargun/service/Infinispan70TopologyHistory.java
```diff
@@ -0,0 +1,51 @@
...
+   }
+
+   @Listener
+   public class Infinispan70TopologyAwareListener extends TopologyAwareListener {
+
+      public Infinispan70TopologyAwareListener(String cacheName) {
+         super(cacheName);
+      }
+
+      @TopologyChanged
```
rvansa: It's not necessary to override these methods.

rvansa:
Generally ok. @alanfx could you check how this will interact with your planned changes?

alanfx:
This change bypasses my changes completely, @rvansa!

@mcimbora, can this also be implemented for servers using JMX?

mcimbora:
@alanfx CacheImpl exposes cache availability via getCacheAvailability().

mcimbora:
@rvansa Updated.

mcimbora:
@rvansa Do not integrate yet, requires implementation for CS mode.

mcimbora:
@alanfx Updated.

plugins/infinispan60/src/main/java/org/radargun/service/Infinispan60ServerTopologyHistory.java
```diff
@@ -162,6 +176,50 @@ protected void processCacheStatus() {
...
+         cacheLoop : for (Map.Entry<String, Object> entry : cacheAttributes.entrySet()) {
+            status = statusMap.get(entry.getKey());
+            // Comparison with previous cache availability is required
+            CacheStatus oldStatus = cacheChangesOngoing.get(entry.getKey());
+            // init
+            if (oldStatus == null) {
+               oldStatus = status;
+            }
+            CacheAvailability cacheAvailability = CacheAvailability.valueOf(entry.getValue().toString());
+            switch (cacheAvailability) {
```
rvansa: can't you simplify this into:
```
if (!oldStatus.prevCacheAvailability.equals(cacheAvailability)) {
   status.cacheAvailabilityChanged = true;
   cacheAvailabilityChanges++;
   log.debugf("Availability status change for cache %s: %s -> %s", entry.getKey(), oldStatus.prevCacheAvailability, cacheAvailability);
}
status.prevCacheAvailability = cacheAvailability;
```

mcimbora:
@alanfx @rvansa Fixed

mcimbora:
@alanfx @rvansa Tests are pushed. This PR depends on https://github.com/radargun/radargun/pull/233. 

As to Alan's previous comments (missing) - I did some reseach and it seems AvailabilityMode enum was added to ISPN in 7.x, it wouldn't work with 6.x branch.
- Cache availability (https://github.com/mcimbora/radargun/blob/d78e346483794d55cab06a457f62fec69dd69daf/plugins/infinispan60/src/main/java/org/radargun/service/Infinispan60ServerTopologyHistory.java#L96) - I'm adding a single event with pre=false, meaning this event is already finished (we care only about the latest changes in availability, there's no before/after event in this case)
- I tried to refactor CacheStatus to contain latest raw results (not boolean values), however this seems to complicate matters a bit. I suggest to leave it as is now (tested & working) and deal with it in 2.2 if necessary.

alanfx:
@mcimbora the presence of Cache availability enum in ISPN 7, implies that this code should be in a Infinispan70ServerTopologyHistory class. I'm fine with not refactoring CacheStatus right now, but I am still confused by the code only adding end events. I think the rest of the code expects begin and end events, so I worry about it getting stuck in a loop waiting for partition availability to change. I have seen this in the past with other events that were not balanced with matching start and end events.

mcimbora:
@alanfx Well, it was there before, however it was located in PartitionHandlingManagerImpl. As to the events, have a look at how WaitForTopologySettle handles this now - it makes sure all events that have started (with Event.finished == null) have their counterpart (with Event.finished != null). If there are more "finished" events, than "non-finished" ones, it's ok. If this is not sufficient, I can think of introducing special kind of event which holds only one time information (no start/end). Test were added to cover this - org/radargun/stages/topologyWaitForTopologySettleStageTest.java. IMO the way WaitForTopologySettle stage worked before wasn't correct - it just checked the last item in the list, ignoring the rest of the history.

alanfx:
@mcimbora, ok it looks good, but I am concerned about the removal of the statusMap.put statements in Infinispan60ServerTopologyHistory. The rest looks good to me.

rvansa:
@mcimbora please rebase

mcimbora:
@alanfx @rvansa Updated & rebased.

mcimbora:
@alanfx @rvansa Fixed, have a look at https://github.com/mcimbora/radargun/commit/a7d06a9eb4871d5bf44ce01aa1f0ed171abd6288 (not a best place to add all those changes, but it was quite difficult to perform a rebase at this point). Events now have time attribute and their type (start/end/single). Single is a new event type, which represents a single point of time, therefore we don't expect such events to be matched when goint through the history. Adjusted tests accordingly, updated also CheckTopologyStage & WaitForTopologyEventStage, including tests.

mcimbora:
@alanfx @rvansa Updated. Added (https://github.com/mcimbora/radargun/commit/5d6742b001e425b6e661c8bd87bec0e42caf9570#diff-c4af248750d8c9c894c10cafd3dc8540R95) directly to TopologyHistory interface & promoted reusability of EventType, utilizing EnumSet as suggested. Modified DefaultConverter accordingly.

rvansa:
Integrated, thanks.

PR #233, opened 2015-05-12
Failover testing - further refactoring

mcimbora:
* Introduce checking for failed transactions and delayed removes
* Handling false stale reads on the same key updated within a single transaction
* Moved a part of logic into dedicated FailureHolder & StressorRecordNotifier classes
* Renaming

This is still quite fresh code and will require some more testing. Pushing it here for review of if anyone wants to try it. The next step will be to simplify BackgroundOperationsManager, which still contains plenty of logic that could be delegated.

Includes fixes from https://github.com/radargun/radargun/pull/231


mcimbora:
@alanfx @rvansa Ready for review. I've been using this version for quite a while so at least it shouldn't explode right on startup :-)

alanfx:
Thanks, @mcimbora

core/src/main/java/org/radargun/stages/cache/background/AbstractLogLogic.java
```diff
@@ -219,28 +230,50 @@ protected void afterRollback() {
...
 
-   protected void afterCommit() {
+   protected boolean afterCommit() {
       boolean inTransaction = false;
       try {
+         Map<Long, Integer> delayedRemoveAttemptsMap = new HashMap<>();
          while (!stressor.isTerminated()) {
             try {
-               if (inTransaction) {
+               if (inTransaction || ongoingTx != null) {
```
rvansa: Looks like we don't need the inTransaction variable at all if ongoingTx != null signalizes that we really are in tx. Or is there any reasonable case when inTransaction == false && ongoingTx != null?

core/src/main/java/org/radargun/stages/cache/background/AbstractLogLogic.java
```diff
@@ -219,28 +230,50 @@ protected void afterRollback() {
...
                   } catch (Exception e) {
-                     log.error("Failed to rollback ongoing transaction", e);
+                     log.error("Failed to rollback ongoing transaction.", e);
                   }
                }
                startTransaction();
                inTransaction = true;
                for (DelayedRemove delayedRemove : delayedRemoves.values()) {
-                  checkedRemoveValue(delayedRemove.keyId, delayedRemove.oldValue);
+                  try {
```
rvansa: Delayed remove can fail during commit, too, and you don't count commit failures. Besides that, it seems to me that counting this for individual keys is rather overkill, I'd use single counter for the whole transaction (if one remove fails, whole transaction needs to be rolled back anyway).

core/src/main/java/org/radargun/stages/cache/background/LogChecker.java
```diff
@@ -110,32 +108,36 @@ public void run() {
...
             if (trace) {
                log.tracef("Checking operation %d for thread %d on key %d (%s)",
-                     record.getOperationId(), record.getThreadId(), record.getKeyId(), keyGenerator.generateKey(record.getKeyId()));
+                          record.getOperationId(), record.getThreadId(), record.getKeyId(), keyGenerator.generateKey(record.getKeyId()));
             }
             boolean notification = record.hasNotification(record.getOperationId());
             Object value = findValue(record);
             boolean contains = containsOperation(value, record);
+            // TODO move to PrivateLogLogic?
+            // If check is unsuccessful, verify whether complementary key (k') containing current record operation does exist (every 10 unsuccessful
```
rvansa: Operation should not be confirmed until delayed removes are executed as well, am I missing something?

core/src/main/java/org/radargun/stages/cache/background/LogChecker.java
```diff
@@ -143,60 +145,55 @@ public void run() {
...
-                     Object ignored = basicCache.get(ignoredKey(slaveIndex, record.getThreadId()));
-                     if (ignored != null && record.getOperationId() <= (Long) ignored) {
-                        log.debugf("Operations %d - %d for thread %d are ignored.", record.getOperationId(), ignored, record.getThreadId());
-                        while (record.getOperationId() <= (Long) ignored) {
-                           record.next();
-                        }
-                        continue;
-                     }
+                     && (logLogicConfiguration.writeApplyMaxDelay <= 0 || System.currentTimeMillis() > confirmationTimestamp + logLogicConfiguration.writeApplyMaxDelay)) {
+                  // Check one more time to be sure the record should not be ignored
```
rvansa: the comment is a bit misleading - you're not checking 'one more time', you're just retrieving the ignored key

core/src/main/java/org/radargun/stages/cache/background/PrivateLogLogic.java
```diff
@@ -48,12 +51,13 @@ protected boolean invokeLogic(long keyId) throws Exception {
          if (prevValue == null || !prevValue.contains(prevOperation.operationId)) {
             // non-cleaned old value or stale read, try backup
             backupValue = checkedGetValue(~keyId);
-            if (backupValue == null || !backupValue.contains(prevOperation.operationId)) {
+            // Modifying the same key within a single transaction may cause false stale reads, avoid it by checking txModificationKeyIds
```
rvansa: I would reconsider other options to get around the stale reads, since having multiple modifications of the same entry within single transaction can reveal bugs, too.

core/src/main/java/org/radargun/stages/cache/background/StressorRecord.java
```diff
@@ -23,9 +23,10 @@
    protected final int threadId;
    protected long currentKeyId;
    protected volatile long currentOp = -1;
-   protected final List<StressorConfirmation> confirmations = new LinkedList<>();
+   protected final List<StressorConfirmation> confirmations = Collections.synchronizedList(new LinkedList<StressorConfirmation>());
```
rvansa: It seems that most of the accesses is synchronized anyway, but getLastConfirmedOperationId() which is not and synchronizedList does not help, since you need to retrieve last element anyway (btw., confirmations.getLast() should be used there).

core/src/main/java/org/radargun/stages/cache/background/StressorRecordNofifier.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+
+/**
+ * Listener handling {@link org.radargun.stages.cache.background.StressorRecord} notifications.
+ *
+ * @author Matej Cimbora
+ * @author Radim Vansa
+ */
+public class StressorRecordNofifier implements CacheListeners.UpdatedListener, CacheListeners.CreatedListener {
```
rvansa: Nofifier

core/src/main/java/org/radargun/stages/cache/background/StressorRecordPool.java
```diff
@@ -26,69 +26,33 @@
...
    private final AtomicReferenceArray<StressorRecord> allRecords;
    // Represents current state of pool
-   private final ConcurrentLinkedQueue<StressorRecord> availableRecords = new ConcurrentLinkedQueue<StressorRecord>();
-   private final BackgroundOpsManager manager;
-   // TODO move the following attributes to a separate class (or elsewhere)
-   private final AtomicLong missingOperations = new AtomicLong();
-   private final AtomicLong missingNotifications = new AtomicLong();
-   private final AtomicLong staleReads = new AtomicLong();
+   private final ConcurrentLinkedQueue<StressorRecord> availableRecords = new ConcurrentLinkedQueue<>();
+   private final StressorRecordNofifier stressorRecordNofifier;
```
rvansa: nofifier

rvansa:
@mcimbora review complete

mcimbora:
@rvansa Thanks, I'll have a look at it.

core/src/main/java/org/radargun/stages/cache/background/AbstractLogLogic.java
mcimbora: Makes sense, fixed.

core/src/main/java/org/radargun/stages/cache/background/AbstractLogLogic.java
mcimbora: Thinking about it now, such key differentiation really brings little added value, fixed.

core/src/main/java/org/radargun/stages/cache/background/PrivateLogLogic.java
mcimbora: As far as I remember the was related to reaching max size & triming checked operations from value. I was hitting the issue regularly, therefore I would prefer to keep current solution, which doesn't spoil test runs. I can add TODO so that we make sure the problem can be handled inthe future when we have better test coverage.

core/src/main/java/org/radargun/stages/cache/background/LogChecker.java
mcimbora: Well, the last sentence is probably a bit misleading, as this was might have been fixed as a part of https://github.com/mcimbora/radargun/commit/ad5ea131fac80b3d130fc87372c343e5e3ce1a1b, skipping timeout check when using ignoreDeadCheckers. Nevertheless this change is valid with other scenarios as well - if both key (e.g. [1,2]) and its complement ([1,2,3]) exist in the cache, we always return the first value (1,2), ignoring the fact value 3 is already present in the cache. In this case I believe checker should proceed and check operation 3, refreshing the timestamps.

mcimbora:
@rvansa I fixed the comments locally. This will require some test runs to verify nothing exploded. I'll add a comment when the PR is updated.

core/src/main/java/org/radargun/stages/cache/background/LogChecker.java
rvansa: Yes, but if you read [1,2] and [1,2,3] is in complement, it implies that operation 3 was not confirmed by writing stressor_xxx and therefore no failure should occur, you just don't progress further. So, if you allow the progress on non-confirmed operations, you could go a bit further (up to the number of operations that can be non-confirmed, but usually much less), but besides being able to truncate values a bit more, there's no point in that.

mcimbora:
@rvansa Comments have been addressed. I ran a sample test without any problems.

mcimbora:
Let's wait a bit with integrating this one until tests are implemented.

mcimbora:
@rvansa @alanfx Added unit tests & bunch of fixes (see https://github.com/mcimbora/radargun/commit/2e1b83b428e9a3efccec01647443e7a22d7925a4). I ran a bunch of tests (elasticity/split-brain), works fine. I still need to add tests for SharedLogLogic/SharedLogValue/SharedLogChecker, however this can be targeted for 2.2 if we want to release in a near future (those components weren't affected too much by refactoring anyway). wdyt?

alanfx:
@mcimbora I am fine with targeting more testing for 2.2. I am interested to see the tests you have already written!

mcimbora:
@alanfx Sure, they're included in the latest commit (https://github.com/mcimbora/radargun/commit/2e1b83b428e9a3efccec01647443e7a22d7925a4).

alanfx:
Since @rvansa has already reviewed this and @mcimbora has addressed the comments, I am going to concentrate on reviewing #232. If that makes sense, then i'll integrate this PR.

core/src/main/java/org/radargun/stages/JMXInvocationStage.java
```diff
@@ -66,7 +66,7 @@
 
    @Override
    public DistStageAck executeOnSlave() {
-      if (!shouldExecute()) {
+      if (!shouldExecute() || !isServiceRunning()) {
```
rvansa: shouldExecute() does not need to be called as it's called from StateBase

core/src/main/java/org/radargun/stages/cache/background/PrivateLogLogic.java
```diff
@@ -31,6 +33,7 @@
    // Keys modified during current transaction, should be recorded to timestamps when
    // the transaction is committed
    private final Collection<KeyOperationPair> txModifications = new ArrayList<>(Math.max(0, transactionSize));
+   private final Map<Long, Long> maxPrunedOperationIds = new HashMap<>(2);
```
rvansa: Please check how you use this with non-tx cache.

mcimbora: Fixed, restricted only to transactional scenarios (for non-tx it doesn't make sense as timestamps are updated after each write/remove).

mcimbora:
@rvansa Fixed. Reverted BreakTx removal, as it seems to risky.

rvansa:
Cogito ergo integrating.
@mcimbora maybe we should consider having the fixes in separate commit so that it can be reviewed more easily (can be amended by committer).

PR #235, opened 2015-05-26
Add GaussianKeysSelector

jmarkos:
Frequency table for numEntries=200 000, mean=100 000, and 10million samples:
alt text

### DIFF
--- a/core/src/main/java/org/radargun/stages/cache/test/GaussianKeysSelector.java
+++ b/core/src/main/java/org/radargun/stages/cache/test/GaussianKeysSelector.java
@@ -1,0 +1,63 @@
+package org.radargun.stages.cache.test;
+
+import java.util.Random;
+
+import org.radargun.config.DefinitionElement;
+import org.radargun.config.Property;
+import org.radargun.config.PropertyHelper;
+
+/**
+ * @author Jakub Markos &lt;jmarkos@redhat.com&gt;
+ */
+public class GaussianKeysSelector implements KeySelector {
+    private final Random random;
+    private final long numEntries;
+    private final long mean;
+    private final long standardDeviation;
+
+    public GaussianKeysSelector(Random random, long numEntries, long mean, long standardDeviation) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/radargun/stages/cache/test/GaussianKeysSelector.java
```diff
@@ -0,0 +1,63 @@
...
+    public GaussianKeysSelector(Random random, long numEntries, long mean, long standardDeviation) {
+        this.random = random;
+        this.numEntries = numEntries;
+        this.mean = mean;
+        this.standardDeviation = standardDeviation;
+    }
+
+    @Override
+    public long next() {
+        long result = (long) (random.nextGaussian() * standardDeviation + mean);
```
mcimbora: This might be a good fit for do-while cycle.

core/src/main/java/org/radargun/stages/cache/test/GaussianKeysSelector.java
```diff
@@ -0,0 +1,63 @@
...
+
+        @Property(doc = "Mean value of indices. Default is numEntries/2.")
+        private long mean = -1;
+
+        @Property(doc = "Standard deviation. Default is numEntries/8.")
+        private long standardDeviation = -1;
+
+        @Override
+        public KeySelector newInstance(CacheOperationsTestStage stage, Random random, int globalThreadId, int threadId) {
+            if (mean == -1) {
```
mcimbora: I think you can initialize the instance variables with their values directly in declaration. If you specify property value, it has a precedence over default value.

mcimbora: Aha, I see the point. Ignore my previous comment :-)

jmarkos:
Updated, thanks for the review.

mcimbora:
Itegrated, thanks!

PR #236, opened 2015-06-11
New (and more correct!) way to run tests

rvansa:
To let you guys know what to expect as soon as my other PRs are merged.


mcimbora:
Shall I start reviewing now to be finished by the end of the year? :-)

rvansa:
@mcimbora: yes

alanfx:
@rvansa and @mcimbora is this RadarGun 3.0?! I'm also wondering if you can explain how this approach is more correct?

rvansa:
@alanfx check out for example the 'It's broken? Coordinated WOT?' part:
http://psy-lob-saw.blogspot.cz/2015/03/fixing-ycsb-coordinated-omission.html

rvansa:
@mcimbora I think these are all changes I've done (added one commit fixing test schema, ff-mergeable).

mcimbora:
@rvansa Good, let me see

mcimbora:
Globally - please use TimeService to get current time.

rvansa:
@mcimbora rebased, dropped commits related to reporting... we'll sort that out later.

mcimbora:
@rvansa Looks like all comments have been lost after the rebase (at least here), I'll continue with the review tomorrow.

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+
+   public void setMinThreadCreationDelay(long minThreadCreationDelay) {
+      this.minThreadCreationDelay = minThreadCreationDelay;
+   }
+
+   /**
+    * Add operation executed when the test is stopped.
+    * @param runnable
+    */
+   public synchronized void addStopListener(Runnable runnable) {
```
mcimbora: Unused?

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+   public void addStressor(boolean failSilently) {
+      int threadIndex = getNextThreadId(maxThreads);
+      if (threadIndex < 0) {
+         if (failSilently) {
+            return;
+         }
+         reachedMax = true;
+         log.warnf("Attempt to create more than %d threads!", maxThreads);
+         Utils.threadDump();
+         return;
```
mcimbora: Shouldn't this throw an exception instead? Otherwise this fails silently even if failSilently == false.

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+         }
+      }
+      synchronized (this) {
+         for (Runnable listener : stopListeners) {
+            listener.run();
+         }
+      }
+   }
+
+   public void addStressor(boolean failSilently) {
```
mcimbora: Method name is a bit misleading as this starts the stressor as well.

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+      interruptStressors();
+      List<Stressor> stressors;
+      synchronized (this) {
+         stressors = new ArrayList<>(this.stressors);
+      }
+      for (Stressor stressor : stressors) {
+         try {
+            stressor.join();
+         } catch (InterruptedException e) {
+            throw new IllegalStateException(e);
```
mcimbora: You may want to restore interrupted status of the thread before rethrowing the exception.

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+      ArrayList<Statistics> statistics = new ArrayList<>();
+      while (!terminated && statistics.size() < statisticsCounter.get()) {
+         Statistics stats = statisticsQueue.poll();
+         if (stats != null) {
+            statistics.add(stats);
+         } else {
+            try {
+               Thread.sleep(1000);
+            } catch (InterruptedException e) {
+               throw new IllegalStateException(e);
```
mcimbora: Interrupted handling.

core/src/main/java/org/radargun/stages/test/Stressor.java
```diff
@@ -10,112 +11,102 @@
...
-
-   private void runInternal() {
-      int i = 0;
-      while (completion.moreToRun(i)) {
-         Object result = null;
-         try {
-            if (useTransactions && txRemainingOperations <= 0) {
-               ongoingTx = stage.transactional.getTransaction();
-               logic.transactionStarted();
+         while (!test.isFinished()) {
```
mcimbora: Shouldn't the loops check interrupted status of the stressor as well? We had problems due to this little detail in BackgroundOperationsManager some time ago.

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+      stressor.start();
+   }
+
+   public int getUsedThreads() {
+      return threadCounter.get();
+   }
+
+   private int getNextThreadId(int maxThreads) {
+      int threadIndex;
+      do {
```
mcimbora: Wouldn't it be easier (to understand) to introduce synchronized method which would increment the thread counter? That would avoid assignment + increment atomicity problem if the loop was introduced for this reason.

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+   private final AtomicLong lastCreated = new AtomicLong(Long.MIN_VALUE);
+   private final ConcurrentMap<Conversation, AtomicInteger> actuallyExecuting = new ConcurrentHashMap<>();
+   private final ArrayList<Runnable> stopListeners = new ArrayList<>();
+
+   private Statistics statisticsPrototype;
+   private AtomicInteger waitingThreads = new AtomicInteger();
+   private AtomicInteger threadCounter = new AtomicInteger();
+   private AtomicInteger statisticsCounter = new AtomicInteger();
+
+   private volatile boolean steadyState = false;
```
mcimbora: The class is quite complex, it would be worth to at least add comments to describe the fields (or find better names which will avoid the need of having those comments).

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+      do {
+         threadIndex = threadCounter.get();
+         if (threadIndex >= maxThreads) {reachedMax = true;
+            return -1;
+         }
+      } while (!threadCounter.compareAndSet(threadIndex, threadIndex + 1));
+      return threadIndex;
+   }
+
+   public void updateSelector(SchedulingSelector<Conversation> schedulingSelector) {
```
mcimbora: This should be standard setter instead?

core/src/main/java/org/radargun/stages/test/RunningTest.java
```diff
@@ -0,0 +1,290 @@
...
+   class ThreadPoolingSelector implements ConversationSelector {
+      private final SchedulingSelector<Conversation> schedulingSelector;
+
+      ThreadPoolingSelector(SchedulingSelector<Conversation> schedulingSelector) {
+         this.schedulingSelector = schedulingSelector;
+      }
+
+      @Override
+      public Conversation next() throws InterruptedException {
+         Conversation previous = lastConversation.get();
```
mcimbora: Could you please document this piece of code better?

core/src/main/java/org/radargun/stages/test/SchedulingSelector.java
```diff
@@ -0,0 +1,131 @@
...
+import org.radargun.utils.TimeService;
+
+/**
+ * Based on provided frequency, returns matching invocation from {@link #next()} or blocks the thread calling it.
+ *
+ * @author Radim Vansa &lt;rvansa@redhat.com&gt;
+ */
+public class SchedulingSelector<T> {
+   private final static Log log = LogFactory.getLog(SchedulingSelector.class);
+   private final static boolean trace = log.isTraceEnabled();
```
mcimbora: Unused

core/src/main/java/org/radargun/stages/test/SchedulingSelector.java
```diff
@@ -0,0 +1,131 @@
...
+   private final int[] invocations;
+   private final long[] intervals;
+   private final T[] operations;
+   private final int numOperations;
```
mcimbora: Is it worth introducing separate variable for numOperations when it can be derived from operations array?

core/src/main/java/org/radargun/stages/test/RunningTest.java
rvansa: Not in this PR, but it will be :)

core/src/main/java/org/radargun/stages/test/RunningTest.java
rvansa: Since this is called from another stressor, it does not make much sense to throw an exception and just catch it elsewhere, logging an error. The test will be later declared failed anyway.

core/src/main/java/org/radargun/stages/test/SchedulingSelector.java
```diff
@@ -0,0 +1,131 @@
...
+      if (operations.length != invocations.length) throw new IllegalArgumentException();
+      this.operations = operations;
+      this.invocations = invocations;
+      this.intervals = intervals;
+      numOperations = operations.length;
+      lastIntervals = new AtomicLongArray(numOperations);
+      todoInvocations = new AtomicIntegerArray(numOperations);
+   }
+
+   /**
```
mcimbora: Documentation

core/src/main/java/org/radargun/stages/test/SchedulingSelector.java
```diff
@@ -0,0 +1,131 @@
...
+    * @return
+    * @throws InterruptedException
+    */
+   public T next() throws InterruptedException {
+      WAIT_LOOP: for (;;) {
```
mcimbora: What about adding a timeout here?

core/src/main/java/org/radargun/stages/test/SchedulingSelector.java
```diff
@@ -0,0 +1,131 @@
...
+         long now = TimeService.currentTimeMillis();
+         int myOffset = offset;
+         INVOCATIONS_LOOP: for (int i = 0; i < numOperations; ++i) {
+            int operationIndex = (i + myOffset) % numOperations;
+
+            long myInterval = intervals[operationIndex];
+            long currentInterval = now / myInterval;
+
+            long lastInterval;
+            boolean set = false;
```
mcimbora: Can you rename this? Can't immediately see what is set.

core/src/main/java/org/radargun/stages/test/Stressor.java
rvansa: I'd prefer to to as little checks as possible, though, it should not incur too much overhead. These threads should get interrupted only after the finished is set to true.

core/src/main/java/org/radargun/stages/test/RunningTest.java
rvansa: The less synchronization the better. This code does not have any race conditions, since the reachedMax should flip only in the false -> true direction.

core/src/main/java/org/radargun/stages/test/RunningTest.java
mcimbora: I disagree. According to log message this is undesired state so I see no point in pretending nothing happened. IMO letting the client decide what to do is more proper way.

core/src/main/java/org/radargun/stages/test/RunningTest.java
rvansa: getSelector() returns ConversationSelector while updateSelector accepts SchedulingSelector - those methods are not symmetric. As ThreadPoolingSelector is an inner class, it cannot be created outside (moving the instance creation to the caller).

core/src/main/java/org/radargun/stages/test/SchedulingSelector.java
rvansa: Why?

core/src/main/java/org/radargun/stages/test/RunningTest.java
mcimbora: On the one hand such optimizations may be beneficial from performance point of view, but what are the costs? The code becomes unmaintainable. Premature optimizations are the root of all evil :-)

core/src/main/java/org/radargun/stages/test/SchedulingSelector.java
mcimbora: Can you guarantee this won't loop indefinitely?

core/src/main/java/org/radargun/stages/test/RunningTest.java
rvansa: I disagree, too. The new thread is added as a *side effect* of retrieving new operation from the selector, it does not relate to what the client wants to do. I would accept raising this from warn to error. The fact that something got broken is not lost, but the caller is not the one who should deal with that.

core/src/main/java/org/radargun/stages/test/RunningTest.java
rvansa: I think that using atomic counters and volatile boolean variable is far from 'unmaintainable'.

mcimbora: Think of it from global perspective. You add one counter here, one counter there and you end up with class having 20 fields.

core/src/main/java/org/radargun/stages/test/SchedulingSelector.java
rvansa: It is desired to have the thread waiting here until it gets the operation - this is the part that should keep threads blocked. It is perfectly valid to have the thread waiting here for the whole duration of test. The only undesired property is that the threads have to wake up every 1 ms too see that the time has advanced and there may be more work to do - this is a place for optimization.
If we want to forcefully stop the thread, we interrupt it - and I can guarantee that the thread reacts to interruption.

rvansa: (unless it is stuck in the do while loop with atomic updates, but it's unlikely that it will be stuck there for too long).

core/src/main/java/org/radargun/stages/test/RunningTest.java
rvansa: Do you argue about number of counters or the synchronization? Maybe I could restructure that to keep the threadCounter increasing and omit reachedMax, but this way seemed more readable to me.

core/src/main/java/org/radargun/stages/test/Stressor.java
```diff
@@ -10,112 +11,102 @@
...
+                  } catch (InterruptedException e) {
+                     // the test is interrupted
+                  } catch (RequestException e) {
+                     // the exception was already logged in makeRequest
+                  }
+               }
+            } finally {
+               stats.end();
+               test.recordStatistics(stats);
+               stats = null; // let's throw NPE if we try to record now
```
mcimbora: If this may ever occur, let's handle the situation instead.

core/src/main/java/org/radargun/stages/test/TestSetupStage.java
```diff
@@ -0,0 +1,137 @@
...
+         return errorResponse("Ramp-up has not stabilized within timeout");
+      } else if (runningTest.isReachedMax()) {
+         return errorResponse("Max thread count reached during ramp-up");
+      }
+      return successfulResponse();
+   }
+
+   protected abstract SchedulingSelector<Conversation> createSelector();
+
+   protected void prepare() {
```
mcimbora: Looks like this could be abstract.

core/src/main/java/org/radargun/stages/test/TestStage.java
```diff
@@ -1,425 +1,164 @@
...
+            }
+        }
+        if (checkRepeatCondition(aggregated)) {
+            return StageResult.SUCCESS;
+        } else {
+            return StageResult.BREAK;
+        }
+    }
+
+    protected void destroy() {
```
mcimbora: Could be removed

core/src/main/java/org/radargun/stages/test/TestSetupStage.java
rvansa: Abstract methods need to be defined. This provides a point which the stage can use optionally.

core/src/main/java/org/radargun/stages/test/legacy/AbstractCompletion.java
```diff
@@ -0,0 +1,39 @@
...
+   @Override
+   public synchronized void start() {
+      if (!started) {
+         startTime = TimeService.nanoTime();
+         started = true;
+      }
+   }
+
+   @Override
+   public synchronized void setCompletionHandler(Runnable handler) {
```
mcimbora: I'd rather opt for introducing CompletionHandler interface.

core/src/main/java/org/radargun/stages/test/Stressor.java
rvansa: I think that it's handled in the end, accesses of stats check in recording() method.

core/src/main/java/org/radargun/stages/test/TestStage.java
rvansa: Nope, this is an extension point that I am about to use.

mcimbora: How I'm supposed to know that? You can add it once it is used.

core/src/main/java/org/radargun/stages/test/TestSetupStage.java
mcimbora: All subclasses seem to implement the method anyway.

core/src/main/java/org/radargun/stages/test/legacy/AbstractCompletion.java
rvansa: I think that lambda expression will be used for handling this now as we are based on Java 8 now, so introducing special interface would not have the effect. I'll fix the usage.

core/src/main/java/org/radargun/stages/test/TestStage.java
rvansa: Okay, I'll document the prepare and destroy methods :) So little clairvoyance in the crowd...

core/src/main/java/org/radargun/stages/test/Stressor.java
mcimbora: So is the comment still valid? Causing intentional NPE anywhere is not acceptable.

core/src/main/java/org/radargun/stages/test/legacy/OperationLogic.java
```diff
@@ -7,28 +10,34 @@
...
     */
-   public void init(Stressor stressor) {
+   public void init(LegacyStressor stressor) {
       this.stressor = stressor;
    }
 
    /**
+    * Release resources held by this logic.
+    */
+   public void destroy() {}
```
mcimbora: No implementation found.

mcimbora:
@rvansa Review complete, however I might add some other comments. As usual, any tests are welcomed. A design doc would be nice as well.

core/src/main/java/org/radargun/stages/test/LoadStage.java
```diff
@@ -0,0 +1,129 @@
...
+
+   @Property(doc = "The number of threads that should load the entries on one slave. Default is 10.")
+   protected int numThreads = 10;
+
+   @Property(doc = "Seed used for initialization of random generators - with same seed (and other arguments)," +
+         " the stage guarantees same entries added to the cache. By default the seed is not set.")
+   protected Long seed;
+
+   @Property(doc = "During loading phase, if the insert fails, try it again. This is the maximum number of attempts. Default is 10.")
+   protected int maxLoadAttempts = 10;
```
mcimbora: Both maxLoadAttempts and waitOnError are unused on this level..

mcimbora:
Please make sure example configs work, e.g. query, jcache are broken.

core/src/main/java/org/radargun/stages/test/legacy/OperationLogic.java
rvansa: You must really hate providing extension points. Though, in this particular case, I have to admit that I haven't used that anywhere (usage got lost during refactoring).

core/src/main/java/org/radargun/stages/test/LoadStage.java
rvansa: Well spotted, I forgot to remove them from LoadDataStage (I'll probably rename that to LoadCacheStage)

rvansa:
@mcimbora I would appreciate review of the updates & integration.

mcimbora:
@rvansa This is to be discussed today with the rest of RG team. Anyway as long as there's no design doc describing new test mechanism and improved test coverage, I won't integrate the PR.

mcimbora:
@rvansa Thanks, these tests are a good starting point. @alanfx @mgencur @vjuranek are you planning to add some more comments? This might be getting ready for integration.

vjuranek:
@mcimbora I haven't time to dive into it in near future, so if you want to integrate it soon, don't wait for any comments from me :-)

alanfx:
@rvansa can you look at the comment I had on ServiceStartStage? I will try to look through more of this today as well.

rvansa:
@mcimbora @alanfx Dropped the controversial change for expected num of slaves, and rebased + fixed benchmarks that haven't been properly updated.

mcimbora:
@rvansa Integrated, thanks.

rvansa:
:triumph: :sweat_smile:

rvansa:
@mcimbora Looking into master, the last commit d5787bd2dfde8d714a4d843a0e4b9c41da798321 was not integrated. Please, correct this.

mcimbora:
@rvansa Fixed

PR #237, opened 2015-06-11
3 small enhancements!

alanfx:
Make requestTerminate() static and the terminate variable static and
volatile in LogChecker and Stressor classes, and change the call site in
BackgroundOpsManager

If multiple groups are used with ServiceStart, calculate the
expectNumSlaves based on getExecutingSlaves().size().

Don't add end events for events that the code didn't start in
Infinispan60ServerTopologyHistory

### DIFF
--- a/core/src/main/java/org/radargun/stages/cache/background/BackgroundOpsManager.java
+++ b/core/src/main/java/org/radargun/stages/cache/background/BackgroundOpsManager.java
@@ -397,14 +397,10 @@ public class BackgroundOpsManager extends ServiceListenerAdapter {
       // interrupt all threads
       log.debug("Stopping stressors");
       if (stressors && stressorThreads != null) {
-         for (int i = 0; i < stressorThreads.length; i++) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/radargun/stages/cache/background/BackgroundOpsManager.java
```diff
@@ -397,14 +397,10 @@ private synchronized void stopBackgroundThreads(boolean stressors, boolean check
       // interrupt all threads
       log.debug("Stopping stressors");
       if (stressors && stressorThreads != null) {
-         for (int i = 0; i < stressorThreads.length; i++) {
-            stressorThreads[i].requestTerminate();
-         }
+         Stressor.requestTerminate();
```
rvansa: This does not fix the problem, it's just a nasty workaround, and breaks the case when you have multiple BackgroundOpsManagers.

alanfx:
Closing for now to look for a better solution to stopping threads...

PR #239, opened 2015-06-11
2 small enhancements!

alanfx:
If multiple groups are used with ServiceStart, calculate the
expectNumSlaves based on getExecutingSlaves().size().

Don't add end events for events that the code didn't start in
Infinispan60ServerTopologyHistory

### DIFF
--- a/core/src/main/java/org/radargun/stages/lifecycle/ServiceStartStage.java
+++ b/core/src/main/java/org/radargun/stages/lifecycle/ServiceStartStage.java
@@ -41,9 +41,10 @@ public class ServiceStartStage extends AbstractServiceStartStage {
    @Property(doc = "Collect configuration files and properties for the service, and pass those to reporters. Default is true.")
    private boolean dumpConfig = true;
 
-   @Property(doc = "The number of slaves that should be up after all slaves are started. Applicable only with " +
-         "validateCluster=true. Default is all slaves in the cluster (in the same site in case of multi-site configuration).")
-   private Integer expectNumSlaves;
+   @Property(doc = "The number of slaves that should be up after all slaves are started. Applicable only with "
+         + "validateCluster=true. Default is all slaves in the cluster where this stage will be executed (in the "
+         + "same site in case of multi-site configuration).")
+   private Integer expectNumSlaves = -1;
 
    @Property(doc = "Set of slaves that should be reachable to the newly spawned slaves (see Partitionable feature for details). Default is all slaves.")
    private Set<Integer> reachable = null;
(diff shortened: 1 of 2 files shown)


alanfx:
After adding some logging, all of the threads were accounted for and stopped within 15 minutes, so I removed that part of the PR.

alanfx:
And I just ran a job that wouldn't stop after an hour, so I'll look at it again

rvansa:
Do you have logs from the stuck run with the logging?

alanfx:
The following Jenkins job has logs. edg-perf14-16 are running the clients. Build 76 has been running for 14 hours, so it probably has some good logs too! :-)

https://jenkins.mw.lab.eng.bos.redhat.com/hudson/user/afield@REDHAT.COM/my-views/view/afield's%20jobs/job/jdg-perf-cs-kill-coordinator-many-caches-rg2/75/

mcimbora:
@alanfx In my view the problem is with LifecycleHelper invoking beforeServiceStop callback in a synchronous fashion. From the logs it's obvious the threads are indeed getting stopped (see e.g. 15:42:44,129 DEBUG [org.radargun.stages.cache.background.BackgroundOpsManager] (sc-main) All threads have joined), however due to retries HotRod client does, it may take quite a while until the threads stop. The fact you're using 500 BackgroundOpsManager instances makes the problem more obvious. How about making the invocation asynchronous by wrapping it in a separate thread (possibly using executor with limited capacity)?

mcimbora:
@alanfx Added better handling of interrupts https://github.com/mcimbora/radargun/commit/8e3858b1472a0384895af75dd7514691af32141b

alanfx:
Those are exactly the changes I want to try @mcimbora!

alanfx:
Added a reusable method to retrieve JMX attributes from MBeans associated with caches

mcimbora:
@alanfx Integrated, thanks! As for bg ops problem, could you please create an issue so that we can track it?

rvansa:
@alanfx I see what's the problem there now - with each BOM taking several seconds to terminate, and synchronously waiting for each one, the sum is too high. So there are two options:
a) make the threads finish faster (e.g. through interrupts)
b) stop them in parallel
As for b), I don't think that the best way is to run all the listeners simply as tasks in threadpool. The best way would be to split it into two phases - in first phase you'd set the terminate flag and in second you'd wait for the finish. Regrettably, I don't know how to solve that in an elegant way - let's discuss suggestions.

alanfx:
@rvansa I think making the threads interruptible will be my first course of action. I think that is the right way to do it as well. To me, setting the terminate flag is a graceful attempt to stop the thread, and interrupting them is the "kill -9" equivalent. I'll try using @mcimbora's code and let you know if it helps.

rvansa:
Regrettably we don't have any kill -9 equivalent, interrupt is rather like regular kill and Thread.stop() is not 100% reliable either. However, if interrupts work for you, it is a good news.

alanfx:
@rvansa ok, 'kill -9' is the wrong analogy, but I think making the threads interruptible will probably make this workable. I ran a job with the PR @mcimbora sent, and both configurations failed and the threads were stopped. 

https://jenkins.mw.lab.eng.bos.redhat.com/hudson/job/jdg-perf-cs-kill-coordinator-many-caches-rg2/78/

PR #240, opened 2015-06-16
JGroups probe wrapper

mcimbora:
* Enable JGroups tuning via probe. Just a preview, may implement additional features (e.g. scheduling specified query task - similar to JGroupsDumper, but more general).
* Added custom infinispan configuration, as enable_diagnostics property of TP needs to be enabled

### DIFF
--- a/core/src/main/java/org/radargun/stages/JGroupsProbeStage.java
+++ b/core/src/main/java/org/radargun/stages/JGroupsProbeStage.java
@@ -1,0 +1,117 @@
+package org.radargun.stages;
+
+import org.radargun.DistStageAck;
+import org.radargun.config.DefinitionElement;
+import org.radargun.config.Property;
+import org.radargun.config.Stage;
+import org.radargun.utils.ReflexiveConverters;
+import org.radargun.utils.TimeConverter;
+
+import java.io.IOException;
+import java.net.DatagramPacket;
+import java.net.DatagramSocket;
+import java.net.InetAddress;
+import java.net.MulticastSocket;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Inspired by org.jgroups.tests.Probe.
+ *
+ * @author Matej Cimbora
+ */
+@Stage(doc = "Allows to invoke JGroups probe queries. For details on probe usage see org.jgroups.tests.Probe.")
+public class JGroupsProbeStage extends AbstractDistStage {
+
+   @Property(doc = "List of queries to be performed.", optional = false, complexConverter = ListConverter.class)
+   private List<Query> queries = new ArrayList<>();
+
+   @Property(doc = "Diagnostic address to send queries to. Default is 224.0.75.75.")
+   private String address = "224.0.75.75";
+
+   @Property(doc = "Diagnostic port. Default is 7500.")
+   private int port = 7500;
+
+   @Property(doc = "Maximum time to wait for query responses. Default is 60 seconds. Valid only when used in conjunction " +
+         "with expectedResponseCount parameter.", converter = TimeConverter.class)
+   private long timeout = 60_000;
+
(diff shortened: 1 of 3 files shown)


alanfx:
@mcimbora it seems like this stage could apply to both the infinispan and jgroups plugins, so it makes sense to me to have it as a stage. Does that make sense to you, @rvansa?

rvansa:
Generally, I find having a stage specific to JGroups a bit confusing, but I guess it makes sense here. So be it.

However, rather than another JGroups config, I would welcome service parameter that would set the `enable_diagnostics` programmatically.

Do we want to have this still as a preview? It's been around for a few months.

alanfx:
I agree that this would be more useful if it did not have a separate configuration and just programmatically changed the `enable_diagnostics` property.

mcimbora:
@alanfx @rvansa Makes sense, fixed.

rvansa:
I think that in pre-7.0 versions, the enableDiagnostics property should not be at all, rather than check configuration and throw exception. That gives it rather confusing meaning - documentation for 5.1 says what happens in future versions until 7.0, and 7.0 is not in the property description at all.

I would just put it into 7.0 and further. Do you have different opinion, @alanfx ?

alanfx:
Hey @rvansa I think that probe will still work in pre-7.0 versions. The difference is that in versions >= 7.0 `enable_diagnostics` can be enabled programmatically. In earlier versions `enable_diagnostics` has to be set to `true` in the configuration.

rvansa:
@alanfx Nobody prohibits you from using the probe for pre-7.0, the `enable_diagnostics` does not need to be there to let you use the probe.

alanfx:
@rvansa ok, then I am confused! If probe can be used when `enable_diagnostics` is `false`, then we don't need the code to enable it programmatically either. The JGroups documentation implies that `enable_diagnostics` needs to be set to `true`. @mcimbora?

rvansa:
@alanfx Sorry about the confusion - probe can be used if `InfinispanXXService.enable_diagnostics` is not set, but `TP.enable_diagnostics` is set to true. The programmatic change in Service is useful in order to reduce number of JGroups configurations. For older versions you need the configuration with this enabled anyway. So the only value is early failure when benchmarks (explicitly) expects the configuration to be have this enabled, and I think that the loss of elegance is not worth it (as user specifies the configurations in benchmark anyway).

alanfx:
IMO failing to start the service if `TP.enable_diagnostics` is `false` is worthwhile. However, it is true that the probe stage will work even if `InfinispanXXService.enable_diagnostics` isn't set and `TP.enable_diagnostics` is set. Ideally the check could be made directly in the stage, but it doesn't have access to the CacheManager or the JGroupsTransport to retrieve the value of `TP.enable_diagnostics`. I'm not sure what's the best solution here...

mcimbora:
What about just letting the user run the stage even without diagnostics enabled? This way the stage will fail after a timeout, which is a valid result. It's user's responsibility to set up the JGroups configuration in a correct way.

alanfx:
IMO the current implementation of this PR best possible at this point. The implementation isn't perfect, but if the benchmark file is written "correctly" and includes the `InfinispanXXService.enable_diagnostics` then there is a benefit over just failing by timing out in the stage.

rvansa:
If you want to keep the attribute in 5.1 (why not even 4.0?), I would suggest to change the docs to a simpler version, just 'Opens JGroups diagnostics port', and in the old version throw exception that it was not possible. Let's not bother users of new versions with any trickery about versions.

mcimbora:
@rvansa Updated. TP.enable_diagnostics property configurable from infinispan70 onwards (removed from previous plugin versions as it made little sense to keep it there - the field wasn't configurable on TP level).

rvansa:
Integrated, thanks.

PR #241, opened 2015-07-10
Fix reporter-perfrepo schema symlink

mcimbora:


mcimbora:
Integrating based on discussion with @rvansa

Issue #242, opened 2015-07-21
Allow expanding env variables without `env.` prefix


vjuranek:
Currently the variables in scenarios are expanded only when passed as java properties or when starts with `env.` prefix. It would be nice to expand env. variables directly, without needs to add `env.` prefix.

mcimbora:
Discussed & explained in https://github.com/radargun/radargun/pull/244.

PR #243, opened 2015-07-21
#242 Use also env. veriable for evaluting variables

vjuranek:


alanfx:
Integrated! Thanks, @vjuranek!

PR #244, opened 2015-07-22
Revert "#242 Use also env. veriable for evaluting variables"

vjuranek:
This reverts commit c5554156969cac77123229a2e0c665e9dedb1f1a.

After further discussion with Radim, previous commit was a result
of misunderstanding of our prior conversation and enviroment
variables should always have previx `env.`. The only correct way
how to pass some variables to the scenario without `env.` prefix
is to use Java properties.

### DIFF
--- a/core/src/main/java/org/radargun/config/Evaluator.java
+++ b/core/src/main/java/org/radargun/config/Evaluator.java
@@ -149,8 +149,6 @@ public class Evaluator {
       String strValue = System.getProperty(property);
       if (strValue != null && !strValue.isEmpty()) {
          value = new Value(strValue.trim());
-      } else if (System.getenv().containsKey(property)) {
-         value = new Value(System.getenv(property).trim());
       } else {
          if (property.startsWith("env.")) {
             String env = System.getenv(property.substring(4));


vjuranek:
Sorry for that (I need to make conclusions crystal clear next time before sending PR)

mcimbora:
Integrated, thanks

alanfx:
Hey @vjuranek and @rvansa, just so I am clear about this, you can only resolve a configuration option to a System property and not an environment variable? Why is this the case?

rvansa:
@alanfx You can resolve env var HOME from ${env.HOME}, but @vjuranek thought that it should be accessible from ${HOME}, too, but this was a misunderstanding. SmartFrog was involved in the discussion and that makes things dirty :)

alanfx:
@rvansa OK, I think I get it now, and sorry I integrated it so quickly. (Using the plugin in Hyperion right now) SmartFrog *does* make things dirty! :-)

rvansa:
@vjuranek will be delighted to help you with migrating to https://github.com/vjuranek/radargun-plugin

alanfx:
@vjuranek is already in deep with me on this! I think there is some work to be done on the startup scripts to set Java properties, though.

vjuranek:
>  think there is some work to be done on the startup scripts to set Java properties, though.

yes, what was passed as `env key=value master.sh` needs to be changed to `master.sh -J '-Dkey=value'`, working on it right now, but still have some issues with parenthesis which for some reason disappers

alanfx:
Cool, thanks @vjuranek! Let me know if you need any help with the scripts or verification. I have made progress running in Hyperion but setting System properties is definitely needed to run older versions of the Infinispan plugins.

PR #247, opened 2015-07-28
RadarGun 2.1.0.Final

mcimbora:
Preparing for 2.1.0 release
- Upgrading version to 2.1.0.Final
- Removed old schema versions
- Replaced schema symlinks by xsd files


alanfx:
Looks good besides the one missing schema file

mcimbora:
@alanfx Thanks for the review. The schema is embedded in xsd files of plugins, which use jcache module as a dependency, e.g. jdg65, infinispan70, hazelcast3, which reminds me of the fact that infinispan71, infinispan72 and infinispan80 should use jcache dependency as well. I'll issue a PR with fix.

mcimbora:
Hmm didn't realize they use infinispan70 as a dependency, this solves the issue and no PR is necessary :-)

core/src/main/resources/benchmark-jcache.xml
```diff
@@ -1,5 +1,5 @@
-<!-- RadarGun 2.0 benchmark -->
-<benchmark xmlns="urn:radargun:benchmark:2.0">
+<!-- RadarGun 2.1 benchmark -->
+<benchmark xmlns="urn:radargun:benchmark:2.1">
```
alanfx: So this schema reference works based on the schema being embedded in the main plugin schema?

alanfx:
Integrated. Thanks @mcimbora! On to RadarGun 3.0!!!

rvansa:
Guys, release commit should not go into master branch! The commit should be just tagged with 2.1.0.Final. Then, master should be updated with another commit bumping 2.1.0-SNAPSHOT to 3.0.0-SNAPSHOT.

rvansa:
@mcimbora will sort it out, no more pushing into master until he starts 3.0.0-SNAPSHOT there, please.

alanfx:
Sorry, @rvansa I guess we need a document for the release process?

rvansa:
@alanfx Yes, it should... I've just used +- the same process as Infinispan does.

alanfx:
@rvansa ok, I haven't done one of those releases either, so I apologize for my ignorance!

mcimbora:
@alanfx @rvansa I think we might want to introduce a new 2.x branch for future fixes. Wdyt?

rvansa:
@mcimbora sure, although I doubt there will be much fixes...

alanfx:
@mcimbora I think it is worth having a branch though, just in case

PR #253, opened 2015-08-12
Improve input stream handling

mcimbora:
* Make sure input streams get closed
* Use try-with-resources where possible

PR #255, opened 2015-08-13
Serialized reporter fixes

mcimbora:
Depends on https://github.com/radargun/radargun/pull/253.

Fixes problem of ReporterHelper initialization - static initializer threw exception as it attempted to register the same reporers multiple times. As bin/report.sh adds contents of reporter-default on classpath, wrong plugin.properties file was read when registering reporter-perfrepo (the one included in reporter-default).

### DIFF
--- a/core/src/main/java/org/radargun/reporting/ReporterHelper.java
+++ b/core/src/main/java/org/radargun/reporting/ReporterHelper.java
@@ -53,12 +53,13 @@ public class ReporterHelper {
    }
 
    public static void loadReporters(File reporterDir, Map<String, Class<? extends Reporter>> reporters) {
+      InputStream stream = null;
       try {
          if (!reporterDir.isDirectory()) {
             log.warn(reporterDir + " is not a directory");
             return;
          }
-         List<URL> urls = new ArrayList<URL>();
+         List<URL> urls = new ArrayList<>();
          for (File jar : reporterDir.listFiles(new Utils.JarFilenameFilter())) {
             urls.add(jar.toURI().toURL());
          }
@@ -67,7 +68,7 @@ public class ReporterHelper {
             return;
          }
          ClassLoader classLoader = new URLClassLoader(urls.toArray(new URL[urls.size()]), ReporterHelper.class.getClassLoader());
-         InputStream stream = classLoader.getResourceAsStream("plugin.properties");
+         stream = classLoader.getResourceAsStream("plugin.properties");
          if (stream == null) {
             log.warn("No JAR in " + reporterDir + " contains properties file");
             return;
@@ -95,6 +96,8 @@ public class ReporterHelper {
          throw new RuntimeException(e);
       } catch (IOException e) {
          throw new RuntimeException("Failed to load properties from " + reporterDir, e);
+      } finally {
+         Utils.close(stream);
       }
    }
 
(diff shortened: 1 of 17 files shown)


mcimbora:
Added option to specify multiple result directories.

alanfx:
Hey @mcimbora, do you want these changes on master, branch_2.x, or both? I guess we should figure out a process for that...

mcimbora:
@alanfx I agree. In my view it would be best to create 2 separate PRs, each of them having its specific target milestone (2.2.0/3.0.0). This makes reviewer's life a bit easier as some commits require further changes before integrating (e.g. make sure right version is used when new pom.xml is added, fixing merge conflicts, etc.). This one should go to both branches. @alanfx wdyt?

rvansa:
@mcimbora @alanfx There should be 2 PRs, but if the PR is non-trivial, PR for master should come first and for 2.x branch only after the reviewer is confident that there'll be no more modifications needed.

alanfx:
@mcimbora I agree with the process described by @rvansa. The handle_pull_request.py will need to be modified to handle multiple branches, though. Unless I am the only one still using this script!

core/src/main/java/org/radargun/reporting/ReporterHelper.java
```diff
@@ -83,7 +85,8 @@ public static void loadReporters(File reporterDir, Map<String, Class<? extends R
                } else {
                   log.debug("Registering reporter '" + type + "' using " + clazzName);
                   try {
-                     Class<? extends Reporter> reporterClazz = (Class<? extends Reporter>) classLoader.loadClass(clazzName);
+                     ClassLoader reporterHelperClassLoader = new URLClassLoader(urls.toArray(new URL[urls.size()]), ReporterHelper.class.getClassLoader());
```
rvansa: Why do you create another ClassLoader and don't use the reporterSpecificClassloader?

plugins/infinispan52/src/main/java/org/radargun/service/AbstractConfigurationProvider.java
```diff
@@ -36,14 +36,17 @@
 
    private void loadConfigFile(String filename, Map<String, byte[]> configs) {
       if (filename != null) {
+         InputStream is = null;
          try {
-            InputStream is = getClass().getClassLoader().getResourceAsStream(filename);
+            is = getClass().getClassLoader().getResourceAsStream(filename);
```
rvansa: Any reason to not use try-with-resources here?

rvansa:
It seems to me that adding the reporters directly on classpath in reporter.sh is a mistake. We should rather add id to REPORTER_PATHS as the default value.

rvansa:
@rmacor This PR is stalled here for a while, will you react?

rvansa:
@rmacor oops, sorry, this was meant for @mcimbora

mcimbora:
@rvansa Sorry, other priorities ATM. We can close this & I'll reopen when ready if that makes sense to you.

rvansa:
@mcimbora Since it's not a new feature but bugfix, I'd prefer to keep it open. But it would be best if you found some time to take a look sometime in next few weeks.

mcimbora:
Closing, fixed by https://github.com/radargun/radargun/commit/2931c23a17212a57223978b970bb8653f1928bc4

PR #257, opened 2015-08-25
Add support for list.get & size operation

mcimbora:


core/src/main/java/org/radargun/config/Evaluator.java
```diff
@@ -422,6 +451,15 @@ private Value(String string) {
             } catch (NumberFormatException e2) {
             }
          }
+         if (string.contains(",") && t == ValueType.STRING) {
+            t = ValueType.LIST;
+            String[] rawValues = string.split(",");
+            List<Value> processedValues = new ArrayList<>(rawValues.length);
+            for (String rawValue : rawValues) {
+               processedValues.add(new Value(rawValue.trim()));
```
rvansa: If you eventually find out that you need to use the value as string, the trimming will lose the whitespaces (you can't re-join the elements). Also, single element lists wouldn't be detected.
It seems to me that you should rather add a method that will try to convert the value to list when appropriate.

core/src/main/java/org/radargun/config/Evaluator.java
```diff
@@ -363,6 +364,34 @@ private static Value abs(Value value) {
       }
    }
 
+   private static Value listGet(Value first, Value second) {
+      if (!second.type.canBeLong()) {
```
rvansa: If the second argument is a list, you can treat this as a multi select, e.g. (1, 2, 3, 4).get(1, 3) -> (2, 4)

mcimbora:
Updated.

rvansa:
Integrated, thx

PR #265, opened 2015-09-04
Support bulk data loading

mcimbora:
Please do not integrate yet, the purpose of this PR is to get an early preview - i.e. wasn't refactored yet

As to batch + tx combo, currently one batch == one unit of transaction. We can discuss whether this should be changed to per key approach.

### DIFF
--- a/core/src/main/java/org/radargun/stages/cache/test/LoadDataStage.java
+++ b/core/src/main/java/org/radargun/stages/cache/test/LoadDataStage.java
@@ -1,7 +1,9 @@
 package org.radargun.stages.cache.test;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


rvansa:
@alanfx Actually the IGNORE_RETURN_VALUES should be used for all BasicOperations.Cache.put() calls - the version without this flag should be getAndPut(), according to JCache semantics. And putAll already returns void. But that should not be related to this PR, it's plugin implementation stuff.

rvansa:
@mcimbora Two concerns: 
I) this will conflict with LoadDataStage -> LoadStage separation in my other PR (I assume you want this in 3.0 branch, too)
II) could loadEntry share the code more between bulk/non-bulk loaders? And it should be probably renamed, when it loads multiple entries...

alanfx:
@rvansa so you are saying that the BasicOperations trait for Infinispan should do this? I don't see it in the code now, and it seems like it should be parameterized for the stages that do want to check the return value. Thanks for reviewing this PR as well. It's a little time-critical for a test next week.

rvansa:
@alanfx Yes, there should be no custom flags, just the trait. InfinispanOperations.Cache uses ignoreReturnValueImpl for implementing the put() call, which has the flag set.

alanfx:
@rvansa Thanks, I see it now

rvansa:
@mcimbora Is this still preview?

mcimbora:
@rvansa I'm working on refactoring the stage a bit & adding some tests. I'll let you know when I'm done (ongoing tasks with higher priority).

mcimbora:
@rvansa Updated

rvansa:
Integrated, thanks.

PR #267, opened 2015-09-08
Infinispan 8.x fixes - 2.x branch

mcimbora:
- org.infinispan.commons.util.concurrent.jdk8backported.ForkJoinPool no longer present in ISPN 8
-- call @Destroy only once, if annotated method is overriden
- Update log4j2 & jboss logging versions

### DIFF
--- a/core/src/main/java/org/radargun/config/InitHelper.java
+++ b/core/src/main/java/org/radargun/config/InitHelper.java
@@ -1,47 +1,110 @@
 package org.radargun.config;
 
+import org.radargun.logging.Log;
+import org.radargun.logging.LogFactory;
+
 import java.lang.reflect.Method;
-import java.util.Stack;
+import java.util.Arrays;
+import java.util.Iterator;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/radargun/config/InitHelper.java
```diff
@@ -1,20 +1,46 @@
...
+
    public static void init(Object target) {
+      processAnnotatedMethods(target, Init.class);
+   }
+
+   public static void destroy(Object target) {
+      processAnnotatedMethods(target, Destroy.class);
+   }
+
+   private static void processAnnotatedMethods(Object target, Class annotationClass) {
```
rvansa: This is not correct - destroy methods must be called in inverse order than inits!

mcimbora:
Fixed.

vjuranek:
Btw: this needs to be pushed also into master branch

rvansa:
Merged through #271

PR #271, opened 2015-09-21
Continuous query support - 2.x

vjuranek:


vjuranek:
This PR assumes, that PR #267 is merged (i.e. cover also these changes)

mgencur:
Isn't it the right moment to add jdg-66 plugin as part of this PR? We need to run continuous query tests with jdg-66.

vjuranek:
@rvansa all your comments should be addressed now (btw: it's better to comment on given files in PR than on commit as after force push your comments are not visible any more)

PR #273, opened 2015-09-23
Query aggregations - preview

jmarkos:
This is a preview - hazelcast plugin still needs to be updated, plus merging with @vjuranek CQ support will be needed (and for 3.x branch also with @rvansa changes).
ProjectionConverter is just a converter for a list of AttributeConverters, but I didn't figure out if they can be somehow merged.
The QueryBuilder interface doesn't have any new having() methods - the filtering that happens before .groupBy() is the WHERE clause, and filtering after groupBy is meant as the HAVING clause.

Example query:
```html
       <query test-name="lala"
              duration="5s"
              num-threads-per-node="1"
              projection="integerValue,sum(integerValue),max(integerValue)"
              group-by="integerValue" 
              query-object-class="org.radargun.query.NumberObject">
           <conditions>
               <le path="integerValue" value="999999"/>
           </conditions>
           <having>
              <gt path="sum(integerValue)" value="50000" />
           </having>
       </query>
```

### DIFF
--- a/core/src/main/java/org/radargun/stages/cache/test/QueryStage.java
+++ b/core/src/main/java/org/radargun/stages/cache/test/QueryStage.java
@@ -30,6 +30,10 @@ import org.radargun.utils.ReflexiveConverters;
 /**
  * Executes Queries using Infinispan-Query API against the cache.
  *
+ * The conditions field are the standard conditions that you would expect in a WHERE clause.
+ * Having refers to the conditions which follow the GROUP BY clause, and can make use
+ * of aggregations.
+ *
  * @author Anna Manukyan
  */
 @Stage(doc = "Stage which executes a Query using Infinispan-query API against all keys in the cache.")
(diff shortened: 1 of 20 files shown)


core/src/main/java/org/radargun/stages/cache/test/QueryStage.java
```diff
@@ -40,14 +44,22 @@
    @Property(optional = false, doc = "Conditions used in the query", complexConverter = ConditionConverter.class)
    private List<Condition> conditions;
 
-   @Property(doc = "Use projection instead of returning full object. Default is without projection.")
-   private String[] projection;
+   @Property(doc = "Having conditions used in the query", complexConverter = ConditionConverter.class)
```
rvansa: doc = "Conditions applied on aggregated columns', or something like that.

jmarkos:
Updated and rebased version of this with Hazelcast support (for 2.x branch) is https://github.com/radargun/radargun/pull/283.

PR #272, opened 2015-09-22
Add plugin for JDG 6.6

vjuranek:

PR #274, opened 2015-09-29
Continuous query support

vjuranek:
Covers also previous related PR #272

### DIFF
--- a/core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
+++ b/core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
@@ -1,0 +1,272 @@
+package org.radargun.stages.cache.test;
+
+import org.radargun.config.Converter;
+import org.radargun.config.DefinitionElement;
+import org.radargun.config.Property;
+import org.radargun.config.PropertyHelper;
+import org.radargun.stages.test.OperationLogic;
+import org.radargun.stages.test.TestStage;
+import org.radargun.traits.InjectTrait;
+import org.radargun.traits.Queryable;
+import org.radargun.utils.NumberConverter;
+import org.radargun.utils.ObjectConverter;
+import org.radargun.utils.ReflexiveConverters;
+
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Base class for stages which contain queries
+ */
+public abstract class AbstractQueryStage extends TestStage {
+
+    @Property(optional = false, doc = "Full class name of the object that should be queried. Mandatory.")
+    protected String queryObjectClass;
+
+    @Property(optional = false, doc = "Conditions used in the query", complexConverter = ConditionConverter.class)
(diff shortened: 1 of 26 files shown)


rvansa:
Needs rebase. The stage should also allow registering multiple similar queries; one of the largest advantage of CQ compared to simple listeners is that the predicate trees for different queries are merged, reducing the number of comparisons needed to decide whether certain modification fits to the filtered set.

rvansa:
And you should apply review comments from #271, too

vjuranek:
PR rebased and updated with support for JDG 6.6. Support for multiple CQ is not easy as I originally thought. I created #281 not to forgot on it and will address it in another PR.

vjuranek:
PR updated with support for continuous query in Hazelcast.

vjuranek:
Added support for CQ in CS mode.

Can you guys please review this? Thanks!

core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
```diff
@@ -0,0 +1,272 @@
...
+    protected List<SortElement> orderBy;
+
+    @Property(doc = "Offset in the results. Default is none.")
+    protected long offset = -1;
+
+    @Property(doc = "Maximum number of the results. Default is none.")
+    protected long limit = -1;
+
+    @Property(doc = "Check whether all slaves got the same result, and fail if not. Default is false.")
+    protected boolean checkSameResult = false;
```
mcimbora: I think 'checkSameResult' and 'checkSameThreadResult' fields shouldn't be pulled up, as they're only used in QueryStage. This way XSD schema will include those properties in ContinuousQueryStage, however their usage will have no effect.

core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
```diff
@@ -0,0 +1,272 @@
...
+
+        private SortElement(String attribute, boolean asc) {
+            this.attribute = attribute;
+            this.asc = asc;
+        }
+    }
+
+    protected static class SortConverter implements Converter<List<SortElement>> {
+        @Override
+        public List<SortElement> convert(String string, Type type) {
```
mcimbora: This might be a good candidate for unit test.

plugins/infinispan60/src/main/java/org/radargun/service/Infinispan60HotrodService.java
```diff
@@ -116,6 +116,12 @@ protected void registerMarshallers(SerializationContext context) {
    }
 
    @ProvidesTrait
+   public Queryable getQueryable() {
+      return new InfinispanHotrodQueryable(this);
+   }
+
+
```
mcimbora: Extra space.

mcimbora:
@vjuranek There are some minor comments, looks good to me in overall.

vjuranek:
@mcimbora thanks for review. I finally get down to this. PR updated and all your comments should be addressed.

mcimbora:
Integrated, thanks!

Issue #275, opened 2015-09-29
Incorrect expected slave count


vjuranek:
In client-server mode RG fails to start as it waits for more servers in the cluster than actually configured. It seems it counts also client into cluster size, e.g. for this configuration
```
<cluster name="1client">
  <group name="server" size="2" />
  <group name="client" size="1" />
</cluster>
```
if expects cluster of size 3:
```
[org.radargun.stages.lifecycle.LifecycleHelper] (sc-main) Number of members=[slave1/clustered(local=true, coord=true), slave0/clustered(local=false, coord=false)] is not the one expected: 3

```

alanfx:
Hey @vjuranek, in your benchmark, are you starting all of the services at once? (i.e. the clients and the servers?) I think it would work if you had two `service-start` elements that each had the `groups` property set:

`<service-start groups="server" />
<service-start groups="client" />`

But I agree that this isn't ideal either. I'm not sure I know of a way to handle this situation, and the one where you specify multiple groups.

`<service-start groups="server1,server2" />`

vjuranek:
Thanks @alanfx 
```
<service-start groups="server"/>
<service-start groups="client"/>
```
seems to work and I'm fine with it. I originally thought that this was some unintended regression and I wanted to fix it. After your explanation I have no problem with it :smiley:

alanfx:
It would be nicer if a single `<service-start/>` element would work correctly in all situations, but I think the stage would have to know about group memberships and service configurations to get this right in all situations. I'll try to think of a better solution, but I'm glad this works, @vjuranek!

rvansa:
I am not sure if the current HotRod service would keep trying to connect if the clients were started too soon (before servers), so this is another risk when having single `<service-start />`

alanfx:
Yeah, that's a good point, @rvansa. The ServiceStartStage code would have to know that the server service should be started prior to the client service. Service dependencies anyone?!

PR #276, opened 2015-09-29
#275 Don't set expected num. of slaves if not explicitly set, default…

vjuranek:
… should be group size


alanfx:
Hey @vjuranek, I made this change to handle the issue you raised. IMO, you should set the expectNumSlaves to the correct value or the group size.

vjuranek:
Groups size doesn't take any effect as it's taken into account only when `expectedSlaves` is not set (see [LifecycleHelper.java#L64](https://github.com/radargun/radargun/blob/master/core/src/main/java/org/radargun/stages/lifecycle/LifecycleHelper.java#L64)). 

In most of the cases IMHO the `expectedSlaves` is equal to the group size so using `null` as default (and therefore effectively using group size) seems to me more user friendly.

alanfx:
I think the problem is with a benchmark that increases the number of services running on each iteration. In that case the group size is the wrong value for `expectedSlaves`. The other situation where this fails is when you specify multiple group names in the `groups` property.

vjuranek:
OK, makes sense to me, thanks for explanation @alanfx ! Closing.

alanfx:
Thanks, @vjuranek. I'm still not sure this is a `perfect` solution for this, but I don't know if that exists!

PR #278, opened 2015-10-02
Add support for operations with lifespan and maxIdle and expiration e…

mgencur:
…vents

* makes it possible to call operations with lifespan and maxIdle


mgencur:
I will create a PR for 2.x branch once this is integrated. It would be very useful to integrate the PR with jdg66 plugin so that we can all leverage it in our tests. Currently my tests run only with Infinispan 8.0

rvansa:
Integrated, thanks.

PR #279, opened 2015-10-07
Add support for operations with lifespan and maxIdle and expiration e…

mgencur:
…vents

This is the same commit that was integrated in master branch. There are no conflicts so it can be directly merged to 2.x branch. Thanks!


jmarkos:
Done, thanks!

PR #282, opened 2015-10-09
Adding new Stream stage which uses DistributedCacheStream API

rmacor:
Adds new Map/Reduce stage (currently only Word Count example) that uses DistributedCacheStream API.

### DIFF
--- a/core/src/main/java/org/radargun/stages/stream/StreamStage.java
+++ b/core/src/main/java/org/radargun/stages/stream/StreamStage.java
@@ -1,0 +1,230 @@
+package org.radargun.stages.stream;
+
+import org.radargun.DistStageAck;
+import org.radargun.StageResult;
+import org.radargun.config.Property;
+import org.radargun.config.Stage;
+import org.radargun.reporting.Report;
+import org.radargun.stages.AbstractDistStage;
+import org.radargun.stages.cache.RandomDataStage;
+import org.radargun.state.SlaveState;
+import org.radargun.stats.DataOperationStats;
+import org.radargun.stats.DefaultStatistics;
+import org.radargun.stats.Statistics;
+import org.radargun.traits.CacheInformation;
+import org.radargun.traits.Clustered;
+import org.radargun.traits.InjectTrait;
+import org.radargun.traits.Streamable;
+import org.radargun.utils.Projections;
+import org.radargun.utils.TimeService;
+import org.radargun.utils.Utils;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * @author Roman Macor &lt;rmacor@redhat.com&gt;
+ */
+@Stage(doc = "Stage which executes a specified stream Task against all keys in the cache.")
+public class StreamStage extends AbstractDistStage {
+
+   @Property(doc = "Name of the test as used for reporting. Default is 'Stream_Stage'.")
+   protected String testName = "Stream_Stage";
+
+   @Property(doc = "Boolean value that determines if the "
+           + "final results of the stream are written to the log of the "
+           + "first slave node. The default is false.")
+   private boolean printResult = false;
+
+   @Property(doc = "The number of times to execute the stream task. The default is 10.")
+   private int numExecutions = 10;
+
+   @Property(doc = "Name of the cache where stream task should be executed. Default is the default cache.")
+   private String cacheName;
+
+   @Property(doc = "The name of the key in the MasterState object that returns the total number of "
+           + "bytes processed by the stream task. The default is RandomDataStage.RANDOMDATA_TOTALBYTES_KEY.")
+   private String totalBytesKey = RandomDataStage.RANDOMDATA_TOTALBYTES_KEY;
+
+   @Property(optional = false, doc = "Fully qualified class name of the StreamConsumer implementation.")
+   private String streamOperationClass;
+
+   @Property(doc = "Boolean value that determines if the parallelStream is used ")
+   private boolean parallelStream = false;
+
+   @Property(name = "statistics", doc = "Type of gathered statistics. Default are the 'dataOperation' statistics " +
+           "(fixed size memory footprint for each operation).", complexConverter = Statistics.Converter.class)
+   protected Statistics statisticsPrototype = new DefaultStatistics(new DataOperationStats());
+
+   @InjectTrait
+   private Streamable streamable;
+
+   @InjectTrait()
+   private Clustered clustered;
+
+   @InjectTrait()
+   private CacheInformation cacheInformation;
+
+   private StreamFunction function;
+
+   private Object streamTaskResult;
+
+   @Override
+   public StageResult processAckOnMaster(List<DistStageAck> acks) {
+      StageResult result = super.processAckOnMaster(acks);
+      if (result.isError())
+         return result;
+
+      Report report = masterState.getReport();
+      Report.Test test = report.createTest(testName, null, true);
+      int testIteration = test.getIterations().size();
+
+      Map<Integer, Report.SlaveResult> numberOfResultKeysResult = new HashMap<>();
+      Map<Integer, Report.SlaveResult> durationsResult = new HashMap<>();
+
+      for (StreamStageAck ack : Projections.instancesOf(acks, StreamStageAck.class)) {
+         if (ack.stats != null) {
+            DataOperationStats opStats = (DataOperationStats) ack.stats.getOperationsStats().get(Streamable.STREAMABLE.name);
+
+            opStats.setTotalBytes((Long) masterState.get(totalBytesKey));
+            test.addStatistics(testIteration, ack.getSlaveIndex(), Collections.singletonList(ack.stats));
+            durationsResult.put(ack.getSlaveIndex(), new Report.SlaveResult(opStats.getResponseTimes(), false));
+            test.addResult(testIteration, new Report.TestResult("Stream task durations on slave" + ack.getSlaveIndex(), durationsResult, "", false));
+         }
+         if (ack.numberOfResultKeys != null) {
+            numberOfResultKeysResult.put(ack.getSlaveIndex(), new Report.SlaveResult(ack.numberOfResultKeys, false));
+            test.addResult(testIteration, new Report.TestResult("Key count in stream result map on slave" + ack.getSlaveIndex(),
+                    numberOfResultKeysResult, "", false));
+         }
+      }
+      return StageResult.SUCCESS;
+   }
+
+   @Override
+   public DistStageAck executeOnSlave() {
+      if (!isServiceRunning()) {
+         return errorResponse("Service is not runnning", null);
+      }
+
+      DistStageAck result = null;
+      Object prevResult = null;
+
+      Statistics stats = statisticsPrototype.copy();
+      stats.begin();
+
+      for (int i = 0; i < numExecutions; i++) {
+         prevResult = streamTaskResult;
+
+         result = executeStreamTask(stats);
+
+         if (prevResult != null) {
+            if (prevResult.equals(streamTaskResult)) {
+               log.info(i + ": Got the same results for two Stream tasks runs");
+            } else {
+               log.error(i + ": Did not get the same results for two Stream tasks runs");
+               break;
+            }
+         }
+         if (result.isError() && exitOnFailure) {
+            break;
+         }
+      }
+      stats.end();
+
+      log.info("Executed stream stage on slave");
+      return result;
+   }
+
+   private void configureStreamTask() {
+      try {
+         function = Utils.instantiate(streamOperationClass);
+      } catch (Exception e) {
+         throw new IllegalArgumentException("Could not instantiate StreamConsumer implementation class: " + streamOperationClass, e);
+      }
+   }
+
+   private DistStageAck executeStreamTask(Statistics stats) {
+      long durationNanos;
+      long start;
+
+      configureStreamTask();
+
+      StreamStageAck ack = new StreamStageAck(slaveState);
+      start = TimeService.nanoTime();
+      try {
+
+         if (parallelStream) {
+            streamTaskResult = function.apply(streamable.parallelStream(cacheName));
+         } else {
+            streamTaskResult = function.apply(streamable.stream(cacheName));
+         }
+
+         durationNanos = TimeService.nanoTime() - start;
+         stats.registerRequest(durationNanos, Streamable.STREAMABLE);
+         log.info("Stream task completed in "
+                 + Utils.prettyPrintTime(durationNanos, TimeUnit.NANOSECONDS));
+
+         long resultCount = function.getResultCount();
+
+         log.info("Result map contains '" + resultCount + "' keys.");
+         ack.setNumberOfResultKeys(resultCount);
+
+         ack.setStats(stats);
+
+         if (printResult) {
+            String printableResult = function.getPrintableResult();
+            if (printableResult != null) {
+               log.info("Stream task result:" + printableResult);
+            }
+         }
+      } catch (Exception e) {
+         durationNanos = TimeService.nanoTime() - start;
+         stats.registerError(durationNanos, Streamable.STREAMABLE);
+         ack.error("executeStreamTask() threw an exception", e);
+         log.error("executeStreamTask() returned an exception", e);
+      }
+
+      if (clustered != null && cacheInformation != null){
+         log.infof("%d nodes were used. %d entries on this node", clustered.getMembers().size(), cacheInformation
+                 .getCache(cacheName).getLocallyStoredSize());
+      }
+      log.info("--------------------");
+
+      return ack;
+   }
+
+   public interface StreamFunction extends Function<Stream, Object> {
+
+      /**
+       * @return result of the stream task in form of String
+       */
+      String getPrintableResult();
+
+      /**
+       * @return number of results produced by stream task
+       */
+      long getResultCount();
+   }
+
+   private static class StreamStageAck extends DistStageAck {
+      private Statistics stats;
+      private String numberOfResultKeys;
+
+      private StreamStageAck(SlaveState slaveState) {
+         super(slaveState);
+      }
+
+      public void setStats(Statistics stats) {
+         this.stats = stats;
+      }
+
+      public void setNumberOfResultKeys(long numberOfResultKeys) {
+         this.numberOfResultKeys = Long.toString(numberOfResultKeys);
+      }
+   }
+}
--- a/core/src/main/java/org/radargun/traits/Streamable.java
+++ b/core/src/main/java/org/radargun/traits/Streamable.java
@@ -1,0 +1,19 @@
+package org.radargun.traits;
+
+import org.radargun.Operation;
+
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * @author Roman Macor &lt;rmacor@redhat.com&gt;
+ */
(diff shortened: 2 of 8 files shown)


rmacor:
PR updated according to @alanfx comment.

Provides a way to specify map, flatMap and collect function. 
WordCount is provided as an example implementation.

Please review, I'm not sure if this is the right way to do this.

plugins/infinispan80/src/main/resources/dist-sync.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<infinispan
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="urn:infinispan:config:7.0 http://www.infinispan.org/schemas/infinispan-config-7.0.xsd"
```
rvansa: In infinispan80 plugin, use 8.0 config schema

plugins/infinispan80/src/main/resources/dist-sync.xml
```diff
@@ -0,0 +1,23 @@
...
+        xmlns="urn:infinispan:config:7.0">
+    <jgroups>
+        <stack-file name="jgroupsStack" path="default-configs/default-jgroups-udp.xml"/>
```
rvansa: Isn't this the default?

rvansa:
I have doubts about the usefulness of `MapReducerStream`, as it hardwires execution of map(...).flatMap(...).collect(...). This was hardwired into former Infinispan's Map/Reduce which was a basis for MapReducer trait, but Streams can do much more. I would like the interface look like 
```
interface Streamable {
Stream stream(String resourceName);
Stream parallelStream(String resourceName);
}
```
and wrap/delegate (if needed) full Stream API. Then, you would not need the `MapReduceInterface`, you would just use `Consumer<Stream>`

Also, I think that we should have more examples than variant flavours of word count, to prove that the interface is capable enough. For example, the current interface does not give you the option to use filter() - what if you want to count just the words starting with 'f'? And I don't want you just to add the `filterFnc`, I want all the expressiveness of streams.

alanfx:
Hey @rmacor, I agree with @rvansa about this. This stage should be implemented so that all of the features of the distributed cache stream API are supported. The fact that it can also be used as a comparison to the previous Map/Reduce implementation is a secondary benefit.

rvansa:
I don't think we need *all* features as this is quite a big task, and methods from `CacheStream` are specific to Infinispan, too. But the provided interface should be subset/aligned to it (e.g. throwing UnsupportedOperationException from the non-implemented stubs) so that if there's a need for another feature, we could just add it without any significant changes to the existing code.

alanfx:
@rvansa and @rmacor, :+1: Sorry, the stream API is large, so I agree that implementing it all would be too big a task.

rmacor:
PR updated.

alanfx:
@rmacor and I talked about the need to remove the rest of the Map/Reduce-isms in the code!

core/src/main/java/org/radargun/traits/Streamable.java
```diff
@@ -0,0 +1,24 @@
...
+@Trait(doc = "Provides interface for executing distributed tasks on streams")
+public interface Streamable {
+   String TRAIT = Streamable.class.getSimpleName();
+   Operation STREAMABLE = Operation.register(TRAIT + ".Streamable");
+
+   Stream stream(String resourceName);
+
+   Stream parallelStream(String resourceName);
+
+   interface StreamConsumer extends Consumer<Stream> {
```
rvansa: If you're expecting a result, it's not just a consumer, it's https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html

core/src/main/java/org/radargun/stages/mapreducestream/MapReduceStreamStage.java
```diff
@@ -0,0 +1,213 @@
...
+      log.infof("%d nodes were used. %d entries on this node", clustered.getMembers().size(), cacheInformation
+              .getCache(cacheName).getLocallyStoredSize());
+      log.info("--------------------");
+
+      return ack;
+   }
+
+   private static class MapReduceStreamAck extends DistStageAck {
+      private Statistics stats;
+      private String numberOfResultKeys;
```
rvansa: Why is this a String?

core/src/main/java/org/radargun/stages/mapreducestream/MapReduceStreamStage.java
```diff
@@ -0,0 +1,213 @@
...
+      configureMapReduceTask();
+
+      MapReduceStreamAck ack = new MapReduceStreamAck(slaveState);
+      try {
+         start = TimeService.nanoTime();
+
+         consumer.accept(streamable.stream(cacheName));
+
+         Object result = consumer.getResult();
+         if (result instanceof Map) {
```
rvansa: Not really elegant. If you want to result be just reported, it should be the consumer (or rather the Function) that processes it to something presentable.

rvansa: I have to correct myself. The computation of 'presentation' should be in different method than the measured call.

core/src/main/java/org/radargun/stages/mapreducestream/MapReduceStreamStage.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * @author Roman Macor &lt;rmacor@redhat.com&gt;
+ */
+@Stage(doc = "Stage which executes a MapReduce Task against all keys in the cache.")
+public class MapReduceStreamStage extends AbstractDistStage {
+
```
alanfx: MapReduceStreamStage -> StreamStage, and the doc string needs to be updated

rmacor:
PR updated.

plugins/infinispan80/src/main/java/org/radargun/stream/example/WordCount.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * @author Roman Macor &lt;rmacor@redhat.com&gt;
+ */
+public class WordCount implements Streamable.StreamFunction {
+
+   private Object result;
+
+   @Override
+   public Object apply(Stream stream) {
+      result = stream.map((Serializable & Function<Map.Entry<String, String>, String[]>) e -> e.getValue().split("[\\p{Punct}\\s&&[^'-]]+")).
```
wburns: The payload here can be reduced significantly by using Infinispan AdvancedExternalizer [1].

This can be done to the object passed in via map, the one to flatMap and the collect argument.

[1] http://infinispan.org/docs/8.1.x/user_guide/user_guide.html#_advanced_externalizers

rvansa: @wburns But you can't add advanced externalizer on the fly (once cache manager is built), which makes the use rather complicated. Should you add all possible externalizers to global configuration for all RG use cases?
Is the serialization cost for this lambda worth it?

core/src/main/java/org/radargun/stages/stream/StreamStage.java
```diff
@@ -0,0 +1,214 @@
...
+   private Object streamTaskResult;
+
+   @Override
+   public StageResult processAckOnMaster(List<DistStageAck> acks) {
+      StageResult result = super.processAckOnMaster(acks);
+      if (result.isError())
+         return result;
+
+      Report report = masterState.getReport();
+      Report.Test test = report.createTest("Stream_Stage", null, true);
```
rvansa: Please make the test name property (`testName` to be consistent with other stages)

core/src/main/java/org/radargun/stages/stream/StreamStage.java
```diff
@@ -0,0 +1,214 @@
...
+   public DistStageAck executeOnSlave() {
+      if (!isServiceRunning()) {
+         return errorResponse("Service is not runnning", null);
+      }
+
+      if (slaveState.getSlaveIndex() == 0) {
+         DistStageAck result = null;
+         Object prevResult = null;
+
+         Statistics stats = new DefaultStatistics(new DataOperationStats());
```
rvansa: Stats should be configurable, too.

core/src/main/java/org/radargun/stages/stream/StreamStage.java
```diff
@@ -0,0 +1,214 @@
...
+ */
+@Stage(doc = "Stage which executes a specified stream Task against all keys in the cache.")
+public class StreamStage extends AbstractDistStage {
+
+   public static final String STREAM_RESULT_KEY = "streamResult";
+
+   @Property(optional = true, doc = "Boolean value that determines if the "
+           + "final results of the stream task are stored in the cache. "
+           + "The result object will be stored at key STREAM_RESULT_KEY. The default is false.")
+   private boolean storeResultInCache = false;
```
rvansa: What is this good for? I can't find any stage reading the result.

core/src/main/java/org/radargun/stages/stream/StreamStage.java
```diff
@@ -0,0 +1,214 @@
...
+
+   @Property(optional = true, doc = "Boolean value that determines if the "
```
rvansa: `optional = true` is the default

core/src/main/java/org/radargun/stages/stream/StreamStage.java
alanfx: This is something the MapReduceStage does as well to allow post processing the result

rvansa: So is this post-processed anywhere? StreamStage shouldn't be a direct copy. And it would be better if we could avoid having dependency on BasicOperations trait.

rvansa: IIRC it was implemented for M/R because M/R allowed to directly store the result in cache - Streams don't allow that.

alanfx: You are correct. That is the reason it was added, and I am ok with removing this feature and the dependency on BasicOperations trait.

core/src/main/java/org/radargun/stages/stream/StreamStage.java
```diff
@@ -0,0 +1,214 @@
...
+      return StageResult.SUCCESS;
+   }
+
+   @Override
+   public DistStageAck executeOnSlave() {
+      if (!isServiceRunning()) {
+         return errorResponse("Service is not runnning", null);
+      }
+
+      if (slaveState.getSlaveIndex() == 0) {
```
mcimbora: @alanfx @rvansa @rmacor Is there any reason to make the slave index hardcoded? This is what I've recently observed in MapReduce stage as well. If it's supposed to run only on one slave, it should be documented in stage documentation and configured by "slaves" attribute.

core/src/main/java/org/radargun/stages/stream/StreamStage.java
```diff
@@ -0,0 +1,214 @@
...
+         log.info("Stream task completed in "
+                 + Utils.prettyPrintTime(durationNanos, TimeUnit.NANOSECONDS));
+
+         long resultCount = function.getResultCount();
+
+         log.info("Result map contains '" + resultCount + "' keys.");
+         ack.setNumberOfResultKeys(resultCount);
+
+         ack.setStats(stats);
+         String printableResult = function.getPrintableResult();
```
mcimbora: function.getPrintableResult() can be skipped if printResult is false.

plugins/infinispan80/src/main/java/org/radargun/stream/example/WordCount.java
```diff
@@ -0,0 +1,44 @@
...
+              flatMap((Serializable & Function<String[], Stream<String>>) Arrays::stream).
+              collect(CacheCollectors.serializableCollector(() -> Collectors.groupingBy(Function.identity(), Collectors.counting())));
+      return result;
+   }
+
+   @Override
+   public String getPrintableResult() {
+      Map<Object, Object> resultMap = (Map<Object, Object>) result;
```
mcimbora: Are you sure 'result' will always be of Map type? If so, why not just make 'result' be a Map type to avoid those casts?

core/src/main/java/org/radargun/stages/stream/StreamStage.java
alanfx: Yes, the stream and map/reduce operation is initiated from one node in the cluster. I don't know if you can make the `slaves` property required, so I think the stage should execute on the node with `slaveState.getSlaveIndex() == 0` if the property isn't set.

plugins/infinispan80/src/main/java/org/radargun/stream/example/WordCount.java
```diff
@@ -0,0 +1,44 @@
...
+      StringBuilder sb = new StringBuilder();
+      for (Map.Entry<Object, Object> entry : resultMap.entrySet()) {
+         sb.append(System.getProperty("line.separator"));
```
mcimbora: flip those 2 lines? Even better: iterator + sb.append(System.getProperty("line.separator") only on hasNext).

core/pom.xml
```diff
@@ -114,6 +114,15 @@
                </execution>
             </executions>
          </plugin>
+         <plugin>
```
mcimbora: Isn't it enough to update plugin in parent pom?

core/src/main/java/org/radargun/stages/stream/StreamStage.java
mcimbora: The problem that I experienced was connected with using multiple groups - this enforces you to specify predefined order of groups (with MR-performing group as first, which is not mentioned in the doc - OK I can find it in a code, but it can cause some confusion). I can think of using getIndexInGroup == 0 + getExecutingSlaves == 1, or at least add a documentation snippet wdyt?

alanfx: I'm not sure what the behavior would be if the `slaves` property is used with multiple `groups` specified? I'm not sure how you want to change the conditional or where the documentation would go though.

mcimbora: Sounds like a good test candidate.
> I'm not sure what the behavior would be if the slaves property is used with multiple groups specified?

getIndexInGroup == 0 && getExecutingSlaves == 1 instead of getSlaveIndex == 0
> I'm not sure how you want to change the conditional

Stage documentation.
> or where the documentation would go though

mcimbora: https://github.com/radargun/radargun/issues/309

rvansa: Good spot, I haven't noticed this check. @alanfx ,
> the stream and map/reduce operation is initiated from one node in the cluster.
Parallel execution of the same task on from multiple nodes is perfectly valid usecase, there's no reason to require the `slaves` property to be explicitly defined, and I think that we should be consistent with other stages behavior - the stage should be ran on all slaves (in parallel) unless specified otherwise using the `slaves` property.

rvansa: > I'm not sure what the behavior would be if the slaves property is used with multiple groups specified

It should be common subset (intersection) of specified groups and slaves.

core/src/main/java/org/radargun/stages/stream/StreamStage.java
rmacor: Alright, I'm removing storeResultInCache functionality.

core/src/main/java/org/radargun/stages/stream/StreamStage.java
alanfx: @rvansa I can see that using the `slaves` property to select the node to run the task on is reasonable, but I am not sure I agree that running the same MapReduceTask on every node in the cluster is a valid use case. If different MapReduceTasks were executed at the same time, I think that might make sense, but that is not implemented in this class or `MapReduceStage`. I also expect that this scenario will cause an OOM due to the RAM usage of the intermediate caches.

plugins/infinispan80/src/main/java/org/radargun/stream/example/WordCount.java
rmacor: @mcimbora and I discussed it, the output looks better in logs if it starts with an empty line.

core/src/main/java/org/radargun/stages/stream/StreamStage.java
rvansa: Do you think that users will never run concurrent tasks on different nodes in production? WordCount is just an example stressing the cluster, we don't need to complicate it with different tasks ATM.
Seeing what happens when you run the same task on multiple nodes in parallel is also valuable, even though the result might be OOME.

rvansa: Anyway, are we settled on removing the check and relying on manual `slaves="0"` setting in config when appropriate?

alanfx: I'm ok with making that change. The MapReduceTask would need more work to handle this change due to storing the results in the cache, but i can work on a PR next week.

rmacor:
PR updated.

alanfx:
This looks good to me, @rmacor. @wburns, @mcimbora, @rvansa, you guys have anymore comments?

rvansa:
`StreamFunction` should go to StreamStage, it does not have really anything to do with the Stream trait itself.
Do you need to put `WordCount` into plugin (as oposed to core)? That should be reusable but the `CacheCollectors.serializableCollectors` call, and this is just kind of convenience method (we can do it in WordCount directly. Let's make the test as portable as possible.
Minor rant: CacheInformation and Clustered traits shouldn't be mandatory here, since these are used only for logging. Do the null check and allow them to be optional.

rmacor:
PR updated according to @rvansa's comment.
I also added parameters to stream and map-reduce benchmarks and extra log to Random data stage so that we can easily compare number of generated words with count calculated by WordCount.

alanfx:
@rmacor looks good to me. I think you have addressed the comments from @rvansa

mcimbora:
@rvansa Any other comments? If not I'll integrate the PR.

mgencur:
I'll just add that using the Java 8 API for map/reduce is totally awful: https://github.com/rmacor/radargun/commit/3f56de895010e80149419dc0dac77c3663c52e50#diff-92c0e0fddf012f72d8e89117775e48e6R26  But that's a problem of Java itself:)

wburns:
@mgencur There are a couple things to highlight with that statement.

1. This isn't map/reduce.  It is allows for much clearer declaration of operations.  Word count is just a trivialized example.  Also map/reduce would require significantly more lines of code to do the same thing.
2. Unfortunately lambdas are not serializable, which is part of the problem you are probably thinking.  If we didn't require serialization the code would be as such:

```
Object result = stream.map(e -> e.getValue().split("[\\p{Punct}\\s&&[^'-]]+")
                      .flatMap(Arrays::stream)
                      .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
```

mgencur:
@wburns , I meant conceptually it is still map/reduce. Yes, Java has a different api for the reduce phase but conceptually it's still m/r. Even the javadocs for "collect" method starts with: "Performs a mutable 
reduction....". Thanks for your points but I didn't change my opinion:)

wburns:
@mgencur Conceptually, yes it is map/reduce.  But map/reduce at a conceptual level is very basic.  I was talking about the map/reduce API pioneered by google.  Everyone can have their opinion.  Like I was trying to allude to, when we start more interesting operations it may help sway you maybe :)

rvansa:
Integrated, thanks.

PR #283, opened 2015-10-14
Query aggregations

jmarkos:
Aggregations in Infinispan 8 use Java8 features, so I moved the whole support only to JDG plugin for 2.x branch.

Not using structured expressions leads to this structure:
```html
<having>
    <gt value="50000">
        <aggregated-path>
            <sum path="integerValue"/>
        </aggregated-path>
    </gt>
</having>
```

Hazelcast requires a class (implementing PropertyExtractor) which extracts the aggregated property from cache entries. We use reflection for this in ReflexivePropertyExtractor. According to my local tests, it wasn't slower than using a special PropertyExtractor (like NumberObjectPropertyExtractor) which would know about the entries used in the test.

### DIFF
--- a/core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
+++ b/core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
@@ -84,12 +84,11 @@ public abstract class AbstractQueryStage extends TestStage {
         }
         if (orderBy != null) {
             for (SortElement se : orderBy) {
-                builder.orderBy(new Queryable.SelectExpression(se.attribute), se.asc ? Queryable.SortOrder.ASCENDING : Queryable.SortOrder.DESCENDING);
+                builder.orderBy(new Queryable.SelectExpression(se.attribute, se.asc));
             }
         } else if (orderByAggregatedColumns != null) {
             for (OrderedSelectExpressionElement orderByAggregatedColumn : orderByAggregatedColumns) {
-                builder.orderBy(orderByAggregatedColumn.toSelectExpression(), orderByAggregatedColumn.toSelectExpression().asc ?
-                      Queryable.SortOrder.ASCENDING : Queryable.SortOrder.DESCENDING);
+                builder.orderBy(orderByAggregatedColumn.toSelectExpression());
             }
         }
         if (projection != null) {
--- a/core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
+++ b/core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
@@ -19,22 +19,36 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Base class for stages which contain queries
+ *
+ * The conditions field are the standard conditions that you would expect in a WHERE clause.
+ * Having refers to the conditions which follow the GROUP BY clause, and can make use
+ * of aggregations
  */
 public abstract class AbstractQueryStage extends TestStage {
 
     @Property(optional = false, doc = "Full class name of the object that should be queried. Mandatory.")
     protected String queryObjectClass;
 
-    @Property(optional = false, doc = "Conditions used in the query", complexConverter = ConditionConverter.class)
+    @Property(optional = false, doc = "Conditions used in the query.", complexConverter = ConditionConverter.class)
     protected List<Condition> conditions;
+
+    @Property(doc = "Conditions applied to groups when using group-by, can use aggregations.", complexConverter = ConditionConverter.class)
+    protected List<Condition> having;
 
     @Property(doc = "Use projection instead of returning full object. Default is without projection.")
     protected String[] projection;
+
+    @Property(doc = "Projection, possibly with aggregations.",
+            complexConverter = ProjectionConverter.class)
+    protected List<SelectExpressionElement>  aggregatedProjection;
 
     @Property(doc = "Use sorting order, in form [attribute[:(ASC|DESC)]][,attribute[:(ASC|DESC)]]*. " +
             "Without specifying ASC or DESC the sort order defaults to ASC. Default is unordereded.",
             converter = SortConverter.class)
     protected List<SortElement> orderBy;
+
+    @Property(doc = "Use grouping, in form [attribute][,attribute]*. Default is without grouping.")
+    protected String[] groupBy;
 
     @Property(doc = "Offset in the results. Default is none.")
     protected long offset = -1;
@@ -61,8 +75,25 @@ public abstract class AbstractQueryStage extends TestStage {
     }
 
     protected static abstract class PathCondition extends Condition {
-        @Property(doc = "Target path (field on the queried object or path through embedded objects)", optional = false)
+        @Property(doc = "Target path (field on the queried object or path through embedded objects)")
         public String path;
+
+        @Property(doc = "Target path, can be aggregated.", complexConverter = SelectExpressionConverter.class)
+        public SelectExpressionElement aggregatedPath;
+
+        Queryable.SelectExpression resolvedPath;
+
+        public void apply(Queryable.QueryBuilder builder) {
+            if (path != null) {
+                resolvedPath = new Queryable.SelectExpression(path);
+            }
+            if (aggregatedPath != null) {
+                resolvedPath = aggregatedPath.toSelectExpression();
+            }
+            if (resolvedPath == null) {
+                throw new IllegalStateException("You need to specify a target path for the condition!");
+            }
+        }
     }
 
     @DefinitionElement(name = "eq", doc = "Target is equal to value")
@@ -72,7 +103,8 @@ public abstract class AbstractQueryStage extends TestStage {
 
         @Override
         public void apply(Queryable.QueryBuilder builder) {
-            builder.eq(path, value);
+            super.apply(builder);
+            builder.eq(resolvedPath, value);
         }
     }
 
@@ -86,7 +118,8 @@ public abstract class AbstractQueryStage extends TestStage {
     protected static class Lt extends PathNumberCondition {
         @Override
         public void apply(Queryable.QueryBuilder builder) {
-            builder.lt(path, value);
+            super.apply(builder);
+            builder.lt(resolvedPath, value);
         }
     }
 
@@ -94,7 +127,8 @@ public abstract class AbstractQueryStage extends TestStage {
     protected static class Le extends PathNumberCondition {
         @Override
         public void apply(Queryable.QueryBuilder builder) {
-            builder.le(path, value);
+            super.apply(builder);
+            builder.le(resolvedPath, value);
         }
     }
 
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
```diff
@@ -61,8 +75,25 @@ public String toString() {
...
+        @Property(doc = "Target path (field on the queried object or path through embedded objects)")
         public String path;
+
+        @Property(doc = "Target path, can be aggregated.", complexConverter = SelectExpressionConverter.class)
+        public SelectExpressionElement aggregatedPath;
+
+        Queryable.SelectExpression resolvedPath;
+
+        public void apply(Queryable.QueryBuilder builder) {
+            if (path != null) {
```
rvansa: You should enforce that only one of path/aggregatedPath is used.

core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
```diff
@@ -72,7 +103,8 @@ public String toString() {
 
         @Override
         public void apply(Queryable.QueryBuilder builder) {
-            builder.eq(path, value);
+            super.apply(builder);
```
rvansa: Declaring another protected abstract void apply(QueryBuilder, SelecExpression) would be probably more elegant than requiring to call super.apply(builder), but that's rather a matter of taste.

core/src/main/java/org/radargun/stages/cache/test/ContinuousQueryStage.java
```diff
@@ -127,12 +127,22 @@ private void registerCQ(SlaveState slaveState) {
...
         if (orderBy != null) {
             for (SortElement se : orderBy) {
-                builder.orderBy(se.attribute, se.asc ? Queryable.SortOrder.ASCENDING : Queryable.SortOrder.DESCENDING);
+                builder.orderBy(new Queryable.SelectExpression(se.attribute), se.asc ? Queryable.SortOrder.ASCENDING : Queryable.SortOrder.DESCENDING);
             }
         }
-        if (projection != null) {
-            builder.projection(projection);
-        }
+       if (projection != null) {
```
rvansa: Move this logic to AbstractQueryStage?

plugins/hazelcast3/src/main/java/org/radargun/service/HazelcastQueryable.java
```diff
@@ -206,6 +238,70 @@ public QueryResult execute() {
...
+
+         Supplier reflexiveSupplier = Supplier.all(propertyExtractor);
+         if (predicate != null) {
+            supplier = Supplier.fromPredicate(predicate, reflexiveSupplier);
+         } else {
+            supplier = reflexiveSupplier;
+         }
+
+         AggregationFunction function = aggregatedAttribute.function;
+         String type = accessor.getReturnType().getSimpleName();
```
rvansa: Seriously? Why don't you use `Class<?> type` and `type == int.class` etc.?

plugins/hazelcast3/src/main/java/org/radargun/service/HazelcastQueryable.java
```diff
@@ -206,6 +238,70 @@ public QueryResult execute() {
...
+
+         aggregation = null;
+         if (function == AggregationFunction.NONE)
```
rvansa: Two-level switch would look better here, IMO; though I am not sure if switch works on classes (rather not, I suppose).

plugins/hazelcast3/src/main/resources/dist-sync-object-format.xml
```diff
@@ -22,10 +22,9 @@
         </join>
     </network>
 
-    <executor-service>
```
rvansa: Why have you removed the other executor service?

plugins/jdg66/src/main/java/org/radargun/service/JDG66EmbeddedQueryable.java
```diff
@@ -0,0 +1,94 @@
...
+import org.infinispan.query.dsl.Expression;
+import org.infinispan.query.dsl.FilterConditionEndContext;
+import org.infinispan.query.dsl.QueryFactory;
+
+/**
+ * Adds support for aggregations.
+ *
+ * @author Jakub Markos &lt;jmarkos@redhat.com&gt;
+ */
+public class JDG66EmbeddedQueryable extends Infinispan70EmbeddedQueryable {
```
rvansa: Isn't there a community version of Infinispan that would suport aggregations?

plugins/hazelcast3/src/main/resources/dist-sync-object-format.xml
mcimbora: I think the configuration changed a bit in one of the previous versions.

plugins/infinispan60/src/main/java/org/radargun/service/AbstractInfinispanQueryable.java
```diff
@@ -142,21 +142,30 @@ public QueryBuilder any(QueryBuilder... subqueries) {
...
          return this;
       }
 
       @Override
-      public QueryBuilder projection(String... attributes) {
+      public QueryBuilder projection(SelectExpression... selectExpressions) {
          if (builder == null) throw new IllegalArgumentException("You have to call projection() on root query builder!");
-         builder.setProjection(attributes);
+         String[] stringAttributes = new String[selectExpressions.length];
+         for (int i = 0; i < selectExpressions.length; i++) {
```
rvansa: throw exception if selectExpression.function != NONE

core/src/main/java/org/radargun/traits/Queryable.java
```diff
@@ -27,27 +27,52 @@
...
+      QueryBuilder ge(SelectExpression selectExpression, Object value);
+      QueryBuilder between(SelectExpression selectExpression, Object lowerBound, boolean lowerInclusive, Object upperBound, boolean upperInclusive);
+      QueryBuilder isNull(SelectExpression selectExpression);
+      QueryBuilder like(SelectExpression selectExpression, String pattern);
+      QueryBuilder contains(SelectExpression selectExpression, Object value);
       QueryBuilder not(QueryBuilder subquery);
       QueryBuilder any(QueryBuilder... subqueries);
-      QueryBuilder orderBy(String attribute, SortOrder order);
-      QueryBuilder projection(String... attribute);
+      QueryBuilder orderBy(SelectExpression selectExpression, SortOrder order);
```
rvansa: Can you do `GROUP BY foo ORDER BY SUM(bar)`? Is this legal without the group by clause? JPQL does not allow that http://openjpa.apache.org/builds/1.2.3/apache-openjpa/docs/jpa_langref.html#jpa_langref_orderby (SQL-2003 seems to do but I don't know if that's supposed to work).

rvansa:
Example use in a benchmark would be nice, too.

plugins/hazelcast3/src/main/java/org/radargun/service/HazelcastQueryable.java
jmarkos: Then you'll also need to do (type == int.class || type == Integer.class) for double and long, instead of just type.equalsIgnoreCase("double"). I just saw a way to save a few chars so I went for it.

plugins/hazelcast3/src/main/resources/dist-sync-object-format.xml
jmarkos: The configuration changed, but I didn't have to remove it completely, so I'll put it back.

core/src/main/java/org/radargun/traits/Queryable.java
jmarkos: Infinispan doesn't allow you to do that, on the other hand it doesn't really follow any specification so 'legal' doesn't mean much.
For example I was just able to run
.select(Expression.property("accountId"), Expression.avg("amount"))
.groupBy("accountId")
.orderBy(Expression.max("amount"))
 but the result was not correct (at least compared to what I'd expect), and I don't know if SQL/JPQL allows this.

jmarkos: But I forgot to add support for ordering by aggregated columns. Personally I'd just omit that, doesn't seem too useful.

rvansa: Could you issue a JIRA for that to Infinispan?
Anyway, I'd like to see a clean semanthics in the RadarGun API, and I don't know what this is supposed to do. So, either comment and define it, or limit it to arguments with obvious meaning (I know that there are not comments on the other methods in this class, but these are IMO quite obvious).

plugins/jdg66/src/main/java/org/radargun/service/JDG66EmbeddedQueryable.java
jmarkos: There is, but it requires Java 8. I thought that the 2.x branch was meant to stay on 7, am I wrong?

rvansa: Makes sense. Could you add the Infinispan 8 community implementation, too?

jmarkos:
Updated. This will require changes to @mgencur 's work, because we can no longer depend on infinispan80 plugin, which uses Java 8.

plugins/infinispan80/src/main/resources/query-aggregations.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2013 Red Hat, Inc. and/or its affiliates.
```
mcimbora: Should this header be removed?

jmarkos: I suppose it's not necessary so I removed all occurences.

jmarkos:
While doing the rebase, I noticed that @vjuranek CQ also depends on the infinispan80 plugin.

vjuranek:
not sure what does it mean @jmarkos , but feel free to do any changes you want, when working on PR #290 recently, I noticed CQ is broken, so I'll have to re-do it anyway. Or was it meant that I should fix it first and you will adjust your PR accordingly?

jmarkos:
@vjuranek sorry I should've been more clear - it means that jdg66 plugin cannot have infinispan80 as a dependency, because jdg66 needs to be java 7 compatible, and infinispan80 will use java 8 features if this PR is integrated (in this class https://github.com/jmarkos/radargun/blob/88bcb82e9b09436170684d3a2cf4c1ffede57de0/plugins/infinispan80/src/main/java/org/radargun/service/Infinispan80EmbeddedQueryable.java ).
Ideally you'd fix it by dropping the dependency on infinispan80 - probably just by duplicating everything in both plugins - that's what Radim suggested I do in this PR.

vjuranek:
Should be fixed in PR #318. Sorry for the delay @jmarkos !

rvansa:
Still expecting the rebase...

jmarkos:
This needs https://github.com/radargun/radargun/pull/318

rvansa:
@jmarkos you're good to go (rebase).

rvansa:
Closing as abandoned. Feel free to reopen when you'll have the rebase.

PR #284, opened 2015-10-19
Fix NPE when unregistering remote listeners

mgencur:
I've tested this fix in a real test and the NPE is gone.

### DIFF
--- a/plugins/infinispan70/src/main/java/org/radargun/service/InfinispanClientListeners.java
+++ b/plugins/infinispan70/src/main/java/org/radargun/service/InfinispanClientListeners.java
@@ -92,8 +92,8 @@ public class InfinispanClientListeners extends AbstractInfinispanListeners {
    public void removeListener(String cacheName, boolean sync) {
       GenericClientListener listener = getListenerOrThrow(cacheName, sync);
       if (listener.isEmpty()) {
-         service.getRemoteManager(false).getCache(cacheName).removeClientListener(listener);
-         service.getRemoteManager(true).getCache(cacheName).removeClientListener(listener);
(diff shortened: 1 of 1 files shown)


plugins/infinispan70/src/main/java/org/radargun/service/InfinispanClientListeners.java
```diff
@@ -92,8 +92,8 @@ public void removeExpiredListener(String cacheName, ExpiredListener listener, bo
    public void removeListener(String cacheName, boolean sync) {
       GenericClientListener listener = getListenerOrThrow(cacheName, sync);
       if (listener.isEmpty()) {
-         service.getRemoteManager(false).getCache(cacheName).removeClientListener(listener);
-         service.getRemoteManager(true).getCache(cacheName).removeClientListener(listener);
+         service.getRemoteManager(false).getCache(cacheName == null ? "" : cacheName).removeClientListener(listener);
```
mcimbora: Does it mean that RCM.getCache("") should return a default cache? I went through RCM class and documentation doesn't mention anything about this. I'd double check whether the method doesn't create cache with empty name. What about using just getCache() when 'cacheName' param is empty?

mgencur: getCache() has the same meaning as get.Cache(""). You can find it in RCM source code. I can change it but that's just a cosmetic change with twice as many lines of code.

mcimbora: I can see it now, integrating. Thanks

PR #288, opened 2015-10-26
Enable to configure backup clusters for HR client

mcimbora:


plugins/infinispan52/src/main/java/org/radargun/service/InfinispanServerService.java
```diff
@@ -57,6 +57,9 @@
    @Property(doc = "Is this executed on Windows (should we use *.bat instead of *.sh)? Default is false.")
    private boolean windows = false;
 
+   @Property(doc = "Name of the script used to start the server. Default is 'clustered'.")
+   protected String startScriptName = "clustered";
```
rvansa: If this is a property, you'll have to use it for all scripts for Infinispan 8+. I am introducing similar change in the another PR.

mcimbora:
@rvansa Updated, added isnp81 plugin

rvansa:
Besides the need for rebase, I think this can be integrated.

mcimbora:
@rvansa @alanfx Rebased. I also removed jdk18 profile as we'll be using version 8 by default.

PR #290, opened 2015-11-02
Add example benchmark for continuous query

vjuranek:


core/src/main/resources/benchmark-continuous-query.xml
```diff
@@ -0,0 +1,90 @@
+<benchmark xmlns="urn:radargun:benchmark:2.2">
```
mcimbora: Could you please fix namespace version (3.0)? This will require some renaming as https://github.com/radargun/radargun/pull/236 has been integrated.

vjuranek:
looks like completely broken by recent changes, will take some time....

mcimbora:
@vjuranek Can you be more specific? Was this caused by latest rebases of query module or by https://github.com/radargun/radargun/pull/236?

vjuranek:
@mcimbora by PR #236, but I don't blame @rvansa he asked me for example benchmark to verify that he doesn't break it, but I failed to send it in a timely manner. I haven't looked much into it yet, but I expect fix should be quite straightforward. Will send fix by beginning of next week probably (as time allows:-)

vjuranek:
PR updated. And here I have to apologize to @rvansa as he actually didn't break anything (except one minor fix), it was my wrong conclusion after skimming through recent changes :-( Sorry, I'll be more careful next time before blaming anybody.

vjuranek:
Covered in PR #315, closing

PR #291, opened 2015-11-02
Continuous query support for 2.x branch

vjuranek:
Basically backport of PR #274


core/src/main/java/org/radargun/stages/cache/test/AbstractQueryStage.java
```diff
@@ -258,7 +255,7 @@ public String convertToString(List<SortElement> value) {
             for (SortElement e : value) {
                 sb.append(e.attribute).append(':').append(e.asc ? "ASC" : "DESC").append(", ");
             }
-            return sb.toString();
+            return sb.toString().substring(0, sb.length() - 2);
```
mcimbora: Just a matter of taste, this workaround could be fixed by using iterator.

mcimbora:
Integrated, thanks.

PR #299, opened 2015-11-06
tests.sh - add fast mode, dist.sh -J param fix

mcimbora:


rvansa:
The -J option is meant to pass only single next argument, why are you changing that? The current behaviour is consistent with e.g. jconsole and other JVM tools.

mcimbora:
Current behavior might be consistent with e.g. jconsole and other JVM tools, but it's not consistent with other scripts we have (master.sh & slave.sh), where multiple params can be passed within a single -J option. Additionally, the old behavior is maintained, i.e. you can still use multiple -J opts.

rvansa:
Integrated, thanks.

PR #303, opened 2015-11-10
Added JGroups-3.6 plugin (https://github.com/radargun/radargun/issues…

belaban:
JGroups 3.6 plugin

### DIFF
--- a/plugins/jgroups30/src/main/java/org/radargun/service/JGroupsService.java
+++ b/plugins/jgroups30/src/main/java/org/radargun/service/JGroupsService.java
@@ -1,22 +1,7 @@
 package org.radargun.service;
 
-import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.concurrent.ThreadLocalRandom;
-
-import org.jgroups.Address;
-import org.jgroups.JChannel;
-import org.jgroups.Message;
-import org.jgroups.ReceiverAdapter;
-import org.jgroups.View;
-import org.jgroups.blocks.MethodCall;
-import org.jgroups.blocks.MethodLookup;
-import org.jgroups.blocks.RequestOptions;
-import org.jgroups.blocks.ResponseMode;
-import org.jgroups.blocks.RpcDispatcher;
+import org.jgroups.*;
+import org.jgroups.blocks.*;
 import org.jgroups.logging.Log;
 import org.jgroups.logging.LogFactory;
 import org.jgroups.util.RspList;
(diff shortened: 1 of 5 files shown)


plugins/jgroups30/src/main/java/org/radargun/service/JGroupsService.java
```diff
@@ -77,7 +69,7 @@
    @Property(doc = "If true, reads are executed on all owners using ResponseMode GET_FIRST. Otherwise it just randomly picks one node for reading. Default is false.")
    private boolean getFirst;
    @Property(doc = "Controls use of the OOB flag. Default is false.")
```
rvansa: Please update the default value in doc text as well.

rvansa:
Simulation of the put method is not using the same pattern as Infinispan; Infinispan does RPC to primary owner and then the primary owner does RPC(s) to backup owners. If primary owner == originator, it's one RPC shorter. On the other hand, JGroups simulates this by doing mcast RPCs to both owners. Could you fix that?

pom.xml
```diff
@@ -193,9 +195,11 @@
             <module>plugins/jgroups32</module>
             <module>plugins/jgroups33</module>
             <module>plugins/jgroups34</module>
+            <module>plugins/jgroups35</module>
+            <module>plugins/jgroups36</module>
          </modules>
          <properties>
-            <plugins.jgroups>jgroups30,jgroups32,jgroups33,jgroups34</plugins.jgroups>
+            <plugins.jgroups>jgroups30,jgroups32,jgroups33,jgroups34,jgroups35,jgroups36</plugins.jgroups>
```
rvansa: JGroups 3.5 has its own profile few lines below. If you're including 3.5 to the default build, please remove that profile and set JGroups dependency in jgroups35 plugin to latest release from 3.5 branch.

belaban:
Actually, I'm rewriting the plugin. As you said, the PUT and GET behave differently in Infinispan, so the change you suggested above should make things faster.
I'm also thinking of creating a standalone 36 plugin, which doesn't depend on 35.
You can scrap this pull request.

rvansa:
And what will be different in the new plugin so that it does not apply to older versions?

belaban:
The behavior:
* A get is sent to the all owners and the call returns after the *first* response has been received. A get which has the local node included returns immediately (mimicking a local read) and no RPC is sent.
* A put is sent to the primary owner, which synchronously updates the backup owner(s) and then returns. The time the caller is blocked is therefore the initial sync put() to the primary and the sync updates to the backups. A put to target set S which includes the local node removes the local node from S and sends the put() RPC to all elements of S minus the local node

rvansa:
Yes, that's the behaviour I'd like to see. But why do you want it to be a standalone and not just test this logic with older versions of JGroups, too?

belaban:
I suggest I issue a PR (today) for 3.6 and add that new plugin. After that, we can have a discussion how to best backport (not sure everything is backportable) to 3.0.
It would be ideal to have just a single plugin and select the JGroups version via maven...

rvansa:
No, single plugin + maven selection is not how this works in RG. We want to be able to test different versions of JGroups against each other without the need to recompile it (in single benchmark). But since JGroups API is quite stable across 3.x, you can have single Service and reuse that in plugins.

OK, please show us the new version and I can help you with the backport if needed.

belaban:
Hi Radim,

I pushed branch rg-295 to my clone of RG, take a look. This does not yet include the PUT changes we discussed yesterday on the mailing list. I won't be able to get much work done next week, too (Berlin).

PR #305, opened 2015-11-10
Namespaces introduction

rvansa:
My attempt to introduce namespaces into benchmark configuration, first stage. I am open to suggestions.
One thing that I am not 100% happy about is that when you have a property with multiple definition elements, those elements need to be specified with the namespace of that property. This happens because the property has type 'property-type'+'converter-type', and it is this in this type where we enumerate the possible inner elements. The alternative would be to use <any> here, but we would lose the code-completion ability then.

Note that d5787bd2dfde8d714a4d843a0e4b9c41da798321 belongs to #236, that should be integrated separately.


rvansa:
I've found some shortcomings yet, but those are rather details - feel free to review & comment.

rvansa:
Fixed the issues, and added refactoring of cache- and query- stages to different modules.

mcimbora:
@rvansa Thanks for introducing this, I'll check soon

mcimbora:
First observation - Number of tests executed for core module: 53 vs 107 in the previous version?

mcimbora:
I see, new modules have been added (e.g. cache). However the tests are not executed there (as maven.test.skip flag is set to true)

core/src/main/java/org/radargun/config/ClasspathScanner.java
```diff
@@ -0,0 +1,103 @@
...
+   protected static <TClass, TAnnotation extends Annotation> void scanClass(String className,
+         Class<TClass> superClass, Class<TAnnotation> annotationClass, String requirePackage, Consumer<Class<? extends TClass>> consumer) {
+      if (requirePackage != null && !className.startsWith(requirePackage)) return;
+      Class<?> clazz;
+      try {
+         System.setErr(NULL_PRINT_STREAM); // suppress any error output
+         clazz = Class.forName(className);
+         System.setErr(ERR_PRINT_STREAM);
+      } catch (Throwable t) {
+         /* There are other problems during class loading that could lead to Errors -> ignore them */
```
mcimbora: This is very vague explanation, can you be more specific? In general, catching Throwable (which includes all Error subclasses!) is a bad practice.

core/src/main/java/org/radargun/config/ClasspathScanner.java
```diff
@@ -0,0 +1,103 @@
...
+         log.trace("Cannot load class " + className);
```
mcimbora: Logging the exception might be useful.

core/src/main/java/org/radargun/config/ClasspathScanner.java
```diff
@@ -0,0 +1,103 @@
...
+   }
+
+   public static <TClass, TAnnotation extends Annotation> void scanJar(String path,
+         Class<TClass> superClass, Class<TAnnotation> annotationClass, String requirePackage, Consumer<Class<? extends TClass>> consumer) {
+      log.tracef("Looking for @%s %s, loading classes from %s", annotationClass.getSimpleName(), superClass.getSimpleName(), path);
+      try (ZipInputStream inputStream = new ZipInputStream(new FileInputStream(path))) {
+         for(;;) {
+            ZipEntry entry = inputStream.getNextEntry();
+            if (entry == null) break;
+            if (!entry.getName().endsWith(".class")) continue;
```
mcimbora: I think introducing .class constant and using entry.getName().length() - CLASS.length would be more obvious at first sight.

core/src/main/java/org/radargun/config/ClasspathScanner.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.radargun.logging.Log;
+import org.radargun.logging.LogFactory;
+
+/**
+ * Helper for listing classes on classpath.
+ *
+ * @author Radim Vansa &lt;rvansa@redhat.com&gt;
+ */
+public class ClasspathScanner {
```
mcimbora: The helper class could be either final/have private ctor.

core/src/main/java/org/radargun/config/ClasspathScanner.java
rvansa: For example you try to load some class that tries to link native methods - you'd get LinkageError or something like that. Generally, I don't want to bother with *any* errors here - ignoring the class when you can't load it is just fine.
Btw., this code was already in, I've just moved it.

core/src/main/java/org/radargun/config/ClasspathScanner.java
rvansa: And too verbose.

mcimbora: Which is acceptable for trace log level. Alternatively there can be 2 versions of the message (debug/trace).

core/src/main/java/org/radargun/config/ClasspathScanner.java
mcimbora: Let's catch CNF & Linkage Error then. Cool, it's a good opportunity to make it right.

rvansa:
Okay, I'll fix those above. Have you taken a look on the example configs? WDYT, is that fine, or do you have any suggestions?

rvansa:
@mcimbora done

mcimbora:
@rvansa Minor one - looks like query module still contains some references to caching - e.g. ContinuousQuery:16 (could be replaced by 'container').

rvansa:
@alanfx Hi Alan, we would like to see your opinion as well before merging this (yes, I know it needs one more rebase).

alanfx:
@rvansa ok, I'll try to take a look at it today.

alanfx:
Overall, I think the changes are ok, but I think this PR should include a document that describes the changes being made and how to migrate RadarGun 2.x benchmark files to RadarGun 3.x. There is already a need to document the various features of the benchmark file.

rvansa:
@mcimbora Fixed commented stuff, + couple of dependency/classloading problems

mcimbora:
@rvansa Thanks. Have you met @alanfx 's requirements?

rvansa:
@mcimbora @alanfx https://github.com/radargun/radargun/wiki/Namespaces Happy?

mcimbora:
@rvansa Cool. Glad to see at least some documentation at last. I'd be nice to have a migration example as @alanfx suggested, but I'm OK with integrating now if you can provide it afterwards. @alanfx ?

alanfx:
@rvansa I'm not happy I had to beg! :-) This is a good start though for documenting the namespaces, but I think the benchmark file format and migration still needs to be documented. That can happen after this is integrated though. I think we need to come up with a more robust strategy for multi-version documentation, though. Is it time to enable GitPages on the repo? @mcimbora ?

rvansa:
@alanfx Right, benchmark file format described in https://github.com/radargun/radargun/wiki/Distributed-Benchmarks deserves an update. Though, I think that separate 'migration guide' is rather over-documenting since between 2.x and 3.0 there are only minimal changes in stage names (there was some renaming in 3.0 branch but this was effectively reverted). And let's not pretend that we have many outside users ATM.
Let's move the discussion about pages to https://github.com/radargun/radargun/issues/325

alanfx:
:+1: Yes, that benchmark file needs to be updated to include variables, repeats, external scenarios, etc. It's true that there aren't many outside users, but the differences between RG 2.x and 3.x are pretty large, so any documentation needs to clearly state which version it is describing.

mcimbora:
@alanfx Agreed, current wiki has its limitations. I'll investigate once I get rid of the PR queue (in progress, looks like I've got some time to work on that now).

mcimbora:
@rvansa I'm not able to build the project, rings a bell? https://gist.github.com/mcimbora/dd28ea0d6978dbff38db

rvansa:
@mcimbora what options? `mvn rebase -i master && mvn clean install` worked like a charm.

mcimbora:
@rvansa Have you committed all your changes? I can reproduce it with the following steps:

1. master branch -> git checkout -b rvansa
2. git pull --rebase https://github.com/rvansa/radargun.git t_namespace
3. mvn clean install

rvansa:
@mcimbora Fixed.

mcimbora:
@rvansa Integrated, thanks

Issue #295, opened 2015-11-05
Create JGroups 3.6.6 plugin


belaban:
New plugin to test JGroups 3.6.6

rvansa:
I was actually planning to drop JGroups plugins completely, since these are rather mocks than really implementing the cache API. They should eventually get back in as we're interested in JGroups performance but in very different shape; a new Messaging [trait](https://github.com/radargun/radargun/wiki/Design-Documentation#plugins-service-and-traits) should be defined that will focus on RPC/messages, and could be implemented by JGroups, JMS and others.

belaban:
I would only drop this *after* Messaging traits have been implemented; meanwhile this is a very useful feature to get a baseline against which to measure Infinispan's perf...

rvansa:
OK, fair enough.

Would you be interested in development of this trait? I guess that you are interested in comparisons with other messaging frameworks.

belaban:
Sorry, no time

belaban:
This will be a new plugin, not extending another `JGroupsService`. It will also contain a few changes, such as
* GET only sent to a single dest (saving us a round-trip)
* PUT sent to primary etc

belaban:
Submitted PR: https://github.com/radargun/radargun/pull/306

mcimbora:
Closing. Created new issue for messaging trait implementation https://github.com/radargun/radargun/issues/323.

PR #306, opened 2015-11-13
New JGroups plugin for 3.6.6.Final, features:

belaban:
- Various optimizations (RequestOptions are created up-front, not on every request)
- Mimics the way Infinispan does reads and writes more closely
  - Details in the javadoc
  - New attribute primary_replicates_puts (default: true) which mimicks the
    way Infinispan does writes
- GH issue: https://github.com/radargun/radargun/issues/295
- Changed defaults for oob (true), flowControl (true) and bundle (true)
  - Performance optimization
- Added (manual) test

### DIFF
--- a/plugins/jgroups36/src/main/java/org/radargun/service/JGroups36Service.java
+++ b/plugins/jgroups36/src/main/java/org/radargun/service/JGroups36Service.java
@@ -1,0 +1,494 @@
+package org.radargun.service;
+
+import org.jgroups.*;
+import org.jgroups.blocks.*;
+import org.jgroups.util.RspList;
+import org.jgroups.util.Util;
+import org.radargun.Service;
+import org.radargun.config.Property;
+import org.radargun.logging.Log;
+import org.radargun.logging.LogFactory;
+import org.radargun.traits.*;
+import org.radargun.utils.Utils;
+
+import javax.management.*;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.management.ManagementFactory;
+import java.lang.reflect.Method;
+import java.util.*;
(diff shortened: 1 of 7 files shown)


plugins/jgroups36/src/main/java/org/radargun/service/JGroups36Service.java
```diff
@@ -0,0 +1,494 @@
...
+        return ch != null && ch.isConnected();
+    }
+
+    public Object getFromRemote(Object key) {
+        assert key !=null;
+        return lastValue;
+    }
+
+    public boolean containsKeyFromRemote(Object key) {
+        assert this != null && key != null;
```
rvansa: `assert this != null`? I think this is Java, not C++

rvansa:
Do you think it would be possible to default configurations provided in JGroups JAR, instead of adding those two files? Many plugins do the same, but the less files the better.

rvansa:
Functionally it seems okay to me.

I would ignore some coding style oddities such as missing space after `if`, `for` or around operators, but please don't use underscores in variable names. And please remove those commented logs to System.out

We should add some basic coding style guidelines on wiki, and hopefully @mcimbora eventually writes a github hook to unify the formatting when issuing PR.

belaban:
OK, try now.

rvansa:
Please Ctrl + F for `exclude_rank`, `backup_owners`, `return_null_on_self_inclusion`, `skip_self`, `primary_rank`, `received_non_null_rsp` and `// System.out`

rvansa:
I've amended the commit with coding-style fixes with few more and integrated this into master branch. Thanks! And looking forward to rewriting this to message-oriented traits.

belaban:
ok, tkx

PR #308, opened 2015-11-19
Fix OOM exception when there is an infinite loop in an exception cause

rmacor:


rvansa:
Could you report the suspended exceptions there, too?

rmacor:
PR updated.

core/src/main/java/org/radargun/DistStageAck.java
```diff
@@ -54,9 +58,16 @@ public void setRemoteException(Throwable remoteException) {
             for (StackTraceElement ste : stackTraceElements) {
                sb.append("\n\t at ").append(ste.toString());
             }
+            sb.append("\nsuppressed: ");
+            sb.append(remoteException.getSuppressed().toString());
```
mcimbora: I think you'll need to use either Arrays.toString or iterate over the elements and print individual StackTraceElements if you want more detailed message (I'd opt for option no. 2).

rmacor:
PR updated.

mcimbora:
@rmacor Thanks Roman, integrated to 2.x branch as well.

PR #311, opened 2015-11-20
Update CQ listener class name, which was renamed

vjuranek:


vjuranek:
This is just a quick fix, which should work with ISPN 8.1 and JDG 6.6. If we want CQ working also for ISPN 8.0, we need separate plugin for ISPN 8.0 and ISPN 8.1 (but not sure if we need it for ISPN 8.0)

vjuranek:
Closing, covered in PR #318

PR #313, opened 2015-11-30
Remove default JVM/JGroups options

mcimbora:


alanfx:
@mcimbora Can you describe why you want to make this change and how it will affect people who use RadarGun from the shell scripts? Thanks

mcimbora:
@alanfx The problem is that these options remain specified even if you define custom opts (e.g. different heap size). I've seen case where my custom option was ignored and default heap size was used instead. The order in which the options are processed seems to be vendor dependent.

mcimbora:
As to the JGroups options, I've run some example configs without problems. If it makes sense to keep those opts for a good reason, I think it would make more sense to define these options somehow on per-plugin basis, rather than specifying them globally.

PR #315, opened 2015-12-01
Support for multiple continuous queries

vjuranek:
Covers also PR #290


plugins/infinispan80/src/main/resources/jgroups-udp.xml
```diff
@@ -0,0 +1,60 @@
+<config xmlns="urn:org:jgroups"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="urn:org:jgroups file:schema/JGroups-2.8.xsd
```
mcimbora: Looks like a quite an old config, e.g. MERGE2 has been deprecated. Why not using default-jgroups-udp.xml that ISPN ships with?

vjuranek:
Good idea @mcimbora , I updated PR accordingly

plugins/infinispan80/src/main/resources/query.xml
```diff
@@ -0,0 +1,42 @@
...
+  ~ 02110-1301 USA
+  -->
+
+<infinispan
+   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+   xsi:schemaLocation="urn:infinispan:config:8.0 http://www.infinispan.org/schemas/infinispan-config-8.0.xsd"
+   xmlns="urn:infinispan:config:8.0">
+
+   <jgroups>
+      <stack-file name="jgroupsStack" path="${infinispan.jgroups.config:default-jgroups-udp.xml}"/>
```
mcimbora: default-configs/default-jgroups-udp.xml is the correct path to default config, unfortunately this one doesn't work.

vjuranek:
err, sorry, fixed now (I did the change in RG target dir only when testing it :-( )

mcimbora:
Integrated, thanks.

PR #316, opened 2015-12-08
Add RG support for Spark

mcimbora:
Folks, this is the version of RG/Spark integration I'm currently using (based on 2.x branch). There may be some changes required (e.g. remove code duplication in RandomDataStage). Just wanted to make sure we'll agree on MR implementation from Spark's perspective. 

- Implemented master/worker/driver services
- Removed "slaveIndex==0" restriction from MapReduceStage as agreed in https://github.com/radargun/radargun/pull/282
- spark plugin uses lambdas as compilation of the plugin is conditioned by using Java 8 
- Updated mvn plugin versions
- I can possibly add more examples (e.g. WordCount based on text file input)

### DIFF
--- a/core/src/main/java/org/radargun/stages/mapreduce/MapReduceStage.java
+++ b/core/src/main/java/org/radargun/stages/mapreduce/MapReduceStage.java
@@ -9,6 +9,7 @@ import java.util.concurrent.TimeUnit;
 
 import org.radargun.DistStageAck;
 import org.radargun.StageResult;
+import org.radargun.config.Init;
 import org.radargun.config.Property;
 import org.radargun.config.Stage;
 import org.radargun.reporting.Report;
@@ -112,6 +113,10 @@ public class MapReduceStage<KOut, VOut, R> extends AbstractDistStage {
    @Property(doc = "The number of times to execute the Map/Reduce task. The default is 10.")
    private int numExecutions = 10;
 
+   @Property(doc = "Compare results of previous executions on entry-by-entry basis. WARNING: This can be lengthy operation " +
+         "on data sets with many distinct keys. On false, only result sizes are checked. The default is true.")
+   private boolean deepComparePreviousExecutions = true;
+
    @Property(doc = "The name of the key in the MasterState object that returns the total number of "
          + "bytes processed by the Map/Reduce task. The default is RandomDataStage.RANDOMDATA_TOTALBYTES_KEY.")
    private String totalBytesKey = RandomDataStage.RANDOMDATA_TOTALBYTES_KEY;
@@ -122,19 +127,30 @@ public class MapReduceStage<KOut, VOut, R> extends AbstractDistStage {
    @InjectTrait(dependency = InjectTrait.Dependency.MANDATORY)
    private MapReducer<KOut, VOut, R> mapReducer;
 
-   @InjectTrait(dependency = InjectTrait.Dependency.MANDATORY)
+   @InjectTrait(dependency = InjectTrait.Dependency.OPTIONAL)
    private Clustered clustered;
 
-   @InjectTrait(dependency = InjectTrait.Dependency.MANDATORY)
+   @InjectTrait
    private CacheInformation cacheInformation;
 
-   @InjectTrait(dependency = InjectTrait.Dependency.MANDATORY)
+   @InjectTrait
    private BasicOperations basicOperations;
 
    @InjectTrait(dependency = InjectTrait.Dependency.OPTIONAL)
    private Iterable iterable;
 
    private boolean supportsResultCacheName = false;
+
+   @Init
+   public void validate() {
+      // check on slaves only
+      if (slaveState != null) {
+         if (storeResultInCache && (basicOperations == null || cacheInformation == null)) {
+            throw new IllegalStateException("'storeResultInCache' can only be used with service which provides " +
+                                                  BasicOperations.class + " and " + CacheInformation.class + " implementation");
+         }
+      }
+   }
 
    @Override
    public StageResult processAckOnMaster(List<DistStageAck> acks) {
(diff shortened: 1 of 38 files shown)


core/src/main/java/org/radargun/stages/mapreduce/MapReduceStage.java
```diff
@@ -122,13 +123,13 @@
    @InjectTrait(dependency = InjectTrait.Dependency.MANDATORY)
    private MapReducer<KOut, VOut, R> mapReducer;
 
-   @InjectTrait(dependency = InjectTrait.Dependency.MANDATORY)
+   @InjectTrait(dependency = InjectTrait.Dependency.OPTIONAL)
```
rvansa: Optional dependency is default, so you don't have to declare that explicitly.

core/src/main/resources/benchmark-spark-mr.xml
```diff
@@ -0,0 +1,100 @@
...
+         </setup>
+         <!-- Configuration for the Spark driver group -->
+         <setup group="driver" plugin="spark">
+            <driver xmlns="urn:radargun:plugins:spark:2.2">
+               <host>localhost.localdomain</host>
+               <port>7077</port>
+               <source-class>org.radargun.service.demo.ispn.WordCountSource</source-class>
+               <source-params>setHotrodServer:${hotrodServer:127.0.0.1};setHotrodPort:${hotrodPort:11322}</source-params>
+               <properties>
+                  <key-value key="spark.serializer" value="org.apache.spark.serializer.KryoSerializer"/>
```
rvansa: If you you have outer tag &lt;properties&gt;, I would suggest rather inner element `<property name="..." value="...";/>`

core/src/main/resources/benchmark-spark-mr.xml
```diff
@@ -0,0 +1,100 @@
...
+      <service-start groups="worker"/>
+      <service-start groups="driver"/>
+      <service-start groups="ispn-server"/>
+      <service-start groups="ispn-client"/>
+
+      <!-- Fill the cache with 100 randomly generated words consuming 1% of the JVM heap -->
+      <random-data groups="ispn-client"
+                   value-size="1024"
+                   ram-percentage="0.01"
+                   random-seed = "13"
```
rvansa: whitespaces

plugins/infinispan52/src/main/java/org/radargun/service/HotRodOperations.java
```diff
@@ -37,7 +43,12 @@ public HotRodOperations(InfinispanHotrodService service) {
       }
    }
 
-   protected class HotRodCache<K, V> implements BasicOperations.Cache<K, V>, ConditionalOperations.Cache<K, V>, TemporalOperations.Cache<K, V> {
+   @Override
+   public <K, V> BulkOperations.Cache<K, V> getCache(String cacheName, boolean preferAsync) {
```
rvansa: You should honour the preferAsync flag (for putAll). Also, recent versions of RemoteCache have the getAll operation implemented natively.

plugins/spark/src/main/java/org/radargun/service/AbstractSparkLifecycle.java
```diff
@@ -0,0 +1,90 @@
...
+public abstract class AbstractSparkLifecycle extends ProcessLifecycle<AbstractSparkService> {
+
+   public AbstractSparkLifecycle(AbstractSparkService service) {
+      super(service);
+   }
+
+   protected void startInternal() {
+      super.startInternal();
+      service.registerAction(getStartPattern(), m -> {
+         for (int i = 0; i < 10; i++) {
```
rvansa: Configurable timeout?

plugins/spark/src/main/java/org/radargun/service/AbstractSparkLifecycle.java
```diff
@@ -0,0 +1,90 @@
...
+
+   @Override
+   protected synchronized StreamReader getErrorReader() {
+      if (errorReader == null) {
+         errorReader = new FileOutputReader(line -> service.reportError(line));
+      }
+      return errorReader;
+   }
+
+   private static class FileOutputReader extends ProcessOutputReader {
```
rvansa: It would be even better if you could configure Spark to log to stdout rather than looking up file...

plugins/spark/src/main/java/org/radargun/service/demo/ispn/WordCountSource.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.Properties;
+
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * @author Matej Cimbora
+ */
+public class WordCountSource implements SparkMapReduce.SparkJavaRDDSource<String> {
```
rvansa: You could break this into abtract class (setting the context, creating infinispan RDD) and the business logic `javaPairRDD.values().flatMap(s -> stream(s.split("[\\p{Punct}\\s&&[^'-]]+")).collect(toList()))`

core/src/main/resources/benchmark-spark-mr.xml
```diff
@@ -0,0 +1,100 @@
...
+               <port>7077</port>
+            </worker>
+         </setup>
+         <!-- Configuration for the Spark driver group -->
+         <setup group="driver" plugin="spark">
+            <driver xmlns="urn:radargun:plugins:spark:2.2">
+               <host>localhost.localdomain</host>
+               <port>7077</port>
+               <source-class>org.radargun.service.demo.ispn.WordCountSource</source-class>
+               <source-params>setHotrodServer:${hotrodServer:127.0.0.1};setHotrodPort:${hotrodPort:11322}</source-params>
```
rvansa: For non-trivial attributes, structured format is preferred:
```
<source-params>
   <source-param ... />
```

rvansa:
I would prefer if a new feature was first added to 3.0 branch, and only after that backported to 2.2.

mgencur:
Radim, although it's a common practice to send PR to upstream first and then backport to other branches, I think it is perfectly fine to do it in the opposite manner for RadarGun. We are the main users of RadarGun and we mostly use 2.x branch because we know its limitations and know what we can expect from it. It will take a while before RG3's logic is proven to be reliable and work well for all our test cases. I'd say branch 2.x is still the main branch. The master branch is more like an experimental one (even though promising).

plugins/spark/src/main/java/org/radargun/service/demo/ispn/WordCountSource.java
mcimbora: @rvansa Sounds overly complicated, this solution is simple & sufficient ATM.

plugins/spark/src/main/java/org/radargun/service/AbstractSparkLifecycle.java
mcimbora: Mission impossible without modifying Spark scripts.

mcimbora:
@rvansa Updated

plugins/spark/src/main/java/org/radargun/service/demo/ispn/WordCountSource.java
rvansa: Splitting class into one abstract class with one abstract method and concrete with just the business method is overly complicated?

mcimbora: It is, compared to current proposal. I'm not going to change it.

rvansa:
Integrated.

rvansa:
Please migrate to master.

PR #321, opened 2016-01-14
Add smoke tests for RG stages

mcimbora:
- Add basic tests for majority of stages
- Trait implementations provided by TraitRepository to avoid plugin-specific dependencies
- ClusteredStageRunner responsible for initialziation of test environment

Still requires some polishing, just making it public for you to see if you agree with those tests conceptually. Ideas are welcome. There is/are usually only one/few tests in a single Test class, this can be easily extended in the future by tweaking properties of Stage classes to make sure more possible code paths are covered.


core/src/test/java/org/radargun/smoke/stages/DefineStageTest.java
```diff
@@ -0,0 +1,37 @@
+package org.radargun.smoke.stages;
```
rvansa: Let's put the test to the same package as the stage.

core/src/test/java/org/radargun/smoke/stages/DefineStageTest.java
```diff
@@ -0,0 +1,37 @@
...
+public class DefineStageTest {
+
+   public void smokeTest() throws Exception {
+      ClusteredStageRunner stageRunner = new ClusteredStageRunner(1);
+      SlaveState slaveState = stageRunner.getSlaveState();
+
+      Assert.assertNull(slaveState.get("test"));
+
+      DefineStage defineStage = new DefineStage();
+      Utils.setField(DefineStage.class, "var", defineStage, "test");
```
rvansa: This is not very nice... Rather make that field package-protected and set it directly.

mcimbora: I know. If this is not acceptable we need to decide whether all property fields of stages should be public or we'll add getters.

core/src/test/java/org/radargun/util/TraitRepository.java
```diff
@@ -0,0 +1,1267 @@
...
+
+      @Override
+      public <K, V> Cache<K, V> getCache(String cacheName) {
+         return cache;
+      }
+   }
+
+   public static class BasicOperationsCache<K, V> implements BasicOperations.Cache<K, V>, Wrappable {
+
+      protected ConcurrentHashMap<K, V> cache;
```
rvansa: I am thinking rather of a log of executed operations

rvansa:
Overall, although you've written a lot of code, these are really just smoke tests, therefore, I am rather skeptical about the ability to really find any bugs (though, it will give you faster response than a special benchmark). I quite miss any exception handling tests - I think that there you'll find the actual bugs in current codebase. Every test needs to be written.
So, I think that this can conceptually go in (after rebase on top of the namespace PR) and it will help developer to write some tests. If he thinks that it is needed, I would rather not go the every-stage-needs-dummy-test path.

mcimbora:
Of course this won't discover plugin-specific issues caused by wrong trait implementation, but it's a good starting point to see whether executeOnSlave & processAckOnMaster methods of a stage are correctly implemented. So not sure what you mean by inability to find any bugs. Writing this already helped me to discover some stage issues. This has a potential to discover more once we test more configuration combinations.

core/src/test/java/org/radargun/smoke/stages/DefineStageTest.java
rvansa: I would be fine with all property fields being package-protected - the less code the better :)

rvansa:
I was not talking about plugin-specific issues; but we can expect that the methods on traits throw exceptions from time to time (e.g. lock timeouts) and the stage should handle that properly.

mcimbora:
@rvansa Painfully rebased :-)

extensions/cache/src/test/java/org/radargun/stages/topology/WaitForTopologySettleStageTest.java
```diff
@@ -50,7 +50,7 @@ public void testExecuteOnSlaveCacheStatusChanges() throws Exception {
    @Test(timeOut = 10_000)
    public void testExecuteOnSlaveMembershipChanges() throws NoSuchFieldException, IllegalAccessException {
       WaitForTopologySettleStage stage = initStage(EnumSet.noneOf(HistoryType.class), true);
-      setClassProperty(WaitForTopologySettleStage.class, stage, "period", 10l);
+      stage.period = 10l;
```
rvansa: In future please use 10L (the lower-case l may be mistaken for 1)

rvansa:
OK, but next time please check the PR on GH, as you've accidentally changed order of imports in many files.

mcimbora:
@rvansa Thanks for the review. Looks like some conflicts have been introduced by latest changes, I'll have a look (and fix the imports at the same time).

mcimbora:
@rvansa Rebased

rvansa:
Rebased & integrated.

PR #331, opened 2016-02-03
Add RESTEasy client and fix unix-func.sh portability

alanfx:
This PR is for the  branch_2.x branch. I'll create a PR for master once this is integrated.

### DIFF
--- a/plugins/process/src/main/resources/unix-func.sh
+++ b/plugins/process/src/main/resources/unix-func.sh
@@ -1,32 +1,36 @@
 #!/bin/sh
 
-findpids() {
-   PIDS=""
-   for pid in `ls /proc`; do
-      grep /proc/$pid/environ -e "RADARGUN_TAG=$1" > /dev/null 2> /dev/null;
-      if [ $? -eq 0 ]; then
-         PIDS="$PIDS $pid";
-      fi;
-   done;
-}
+#
+# PARAM 1 - The PID of the service
+#
+
+# This script requires pgrep and jps to be installed.
+command -v pgrep >/dev/null 2>&1 || { echo >&2 "This script requires pgrep, but it's not installed.  Aborting."; exit 1; }
+command -v jps >/dev/null 2>&1 || { echo >&2 "This script requires jps, but it's not installed.  Aborting."; exit 1; }
 
 findjvms() {
    JVMS=""
-   for pid in `jps | cut -f 1 -d " "`; do
-      grep /proc/$pid/environ -e "RADARGUN_TAG=$1" > /dev/null 2> /dev/null;
-      if [ $? -eq 0 ]; then
+   for pid in `jps -q`; do
+      # If the service process is a Java process
+      if [ "$pid" == "$1" ]; then
          JVMS="$JVMS $pid";
+      else
+         # If the service process is the parent of a Java process
+         jvm=`pgrep -P $1` > /dev/null 2> /dev/null;
+         if [ "$jvm" == "$pid" ]; then
+            JVMS="$JVMS $pid";
+         fi;
       fi;
    done;
 }
 
 killtree() {
-    local _pid=$1
-    local _sig=$2
-    kill -STOP ${_pid}
-    for _child in $(ps -o pid --no-headers --ppid ${_pid}); do
-        killtree ${_child} ${_sig}
-    done
-    kill -${_sig} ${_pid}
-    kill -CONT ${_pid}
+   local _pid=$1
+   local _sig=$2
+   kill -STOP ${_pid} > /dev/null 2> /dev/null;
+   for _child in $(pgrep -P ${_pid}); do
+      killtree ${_child} ${_sig};
+   done
+   kill -${_sig} ${_pid} > /dev/null 2> /dev/null;
+   kill -CONT ${_pid} > /dev/null 2> /dev/null;
 }
--- a/plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
+++ b/plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
@@ -1,0 +1,275 @@
+package org.radargun.service;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.UnsupportedEncodingException;
+import java.net.InetSocketAddress;
+import java.net.URLEncoder;
+
+import javax.ws.rs.HttpMethod;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.Invocation;
+import javax.ws.rs.client.Invocation.Builder;
+import javax.ws.rs.core.EntityTag;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+
+import org.radargun.logging.Log;
+import org.radargun.logging.LogFactory;
+import org.radargun.traits.BasicOperations;
+
+/**
+ * Implementation of {@link BasicOperations} through the HTTP protocol, using the JAX-RS 2.0 client
+ * api with RESTEasy implementation.
+ *
+ * @author Alan Field &lt;afield@redhat.com&gt;
+ */
+public class RESTEasyHTTPOperations implements BasicOperations {
+   private static final Log log = LogFactory.getLog(RESTEasyHTTPOperations.class);
+   private final RESTEasyHTTPService service;
+
+   public RESTEasyHTTPOperations(RESTEasyHTTPService service) {
+      this.service = service;
+   }
+
+   @Override
+   public <K, V> HTTPCache<K, V> getCache(String cacheName) {
+      if (service.isRunning()) {
+         if (cacheName != null && (service.cacheName == null || !service.cacheName.equals(cacheName))) {
+            throw new UnsupportedOperationException();
+         }
+         return new HTTPCache<K, V>();
+      }
+      return null;
+   }
+
+   protected class HTTPCache<K, V> implements BasicOperations.Cache<K, V> {
+
+      @Override
+      public V get(K key) {
+         return getInternal(key).value;
+      }
+
+      private WrappedValue<V> getInternal(K key) {
+         V value = null;
+         EntityTag eTag = null;
+         if (service.isRunning()) {
+            String target = buildUrl(service.cacheName, key);
+            Response response = null;
+            try {
+               Invocation get = service.getHttpClient().target(target).request().accept(service.getContentType())
+                     .buildGet();
+               response = get.invoke();
+               if (response.getStatus() == Status.NOT_FOUND.getStatusCode()) {
+                  log.warn("RESTEasyHTTPOperations.getInternal::Key: " + key + " does not exist in cache: "
+                        + service.cacheName);
+               } else {
+                  eTag = response.getEntityTag();
+                  value = decodeByteArray(response.readEntity(byte[].class));
+               }
+            } catch (Exception e) {
+               throw new RuntimeException("RESTEasyHTTPOperations::get request threw exception: " + target, e);
+            } finally {
+               if (response != null) {
+                  response.close();
+               }
+            }
+         }
+         return new WrappedValue<V>(eTag, value);
+      }
+
+      private class WrappedValue<W> {
+         EntityTag eTag;
+         W value;
+
+         public WrappedValue(EntityTag eTag, W value) {
+            super();
+            this.eTag = eTag;
+            this.value = value;
+         }
+      }
+
+      @Override
+      public boolean containsKey(K key) {
+         if (service.isRunning()) {
+            String target = buildUrl(service.cacheName, key);
+            Response response = null;
+            response = service.getHttpClient().target(target).request().build(HttpMethod.HEAD).invoke();
+            if (response.getStatus() == Status.OK.getStatusCode()) {
+               return true;
+            }
+            if (response.getStatus() == Status.NOT_FOUND.getStatusCode()) {
+               return false;
+            }
+            throw new RuntimeException("RESTEasyHTTPOperations.containsKey::Unexpected HttpStatus: "
+                  + response.getStatus() + " " + response.getStatusInfo().getReasonPhrase() + " for request " + target);
+         }
+         return false;
+      }
+
+      @Override
+      public void put(K key, V value) {
+         putInternal(key, value, null);
+      }
+
+      private void putInternal(K key, V value, EntityTag eTag) {
+         if (service.isRunning()) {
+            Response response = null;
+            try {
+               String target = buildUrl(service.cacheName, key);
+               Builder putBuilder = service.getHttpClient().target(target).request().accept(service.getContentType());
+               if (eTag != null) {
+                  // If the eTag doesn't match the current value for the key, then the put will fail
+                  putBuilder = putBuilder.header(HttpHeaders.IF_MATCH, eTag.getValue());
+               }
+               response = putBuilder.buildPut(Entity.entity(encodeObject(value), service.getContentType())).invoke();
+               int status = response.getStatus();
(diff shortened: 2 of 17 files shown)


plugins/process/src/main/resources/unix-func.sh
```diff
@@ -2,8 +2,8 @@
 
 findpids() {
    PIDS=""
-   for pid in `ls /proc`; do
-      grep /proc/$pid/environ -e "RADARGUN_TAG=$1" > /dev/null 2> /dev/null;
+   for pid in `ps -e -o pid | tail -n +2`; do
+      ps eww $pid | grep "RADARGUN_TAG=$1" > /dev/null 2> /dev/null;
```
rvansa: Just out of curiosity, why have you changed this?

rvansa: Eh, I should have looked on the commit message :) On what OS have you met the /proc limitation?

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,226 @@
...
+               }
+            }
+            return true;
+         }
+         return false;
+      }
+
+      @Override
+      public V getAndRemove(K key) {
+         // Not Atomic
```
rvansa: http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7/
200 -> return value
202 -> existing behaviour
204 -> return null

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,226 @@
...
+            } catch (NullPointerException e) {
+               throw new RuntimeException(e);
+            } catch (IllegalArgumentException e) {
+               throw new RuntimeException(e);
+            }
+         }
+      }
+
+      private String buildUrl(String cache, K key) {
+         String s = new StringBuffer(buildCacheUrl(cache)).append("/").append(key).toString();
```
rvansa: StringBuilder please

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,226 @@
...
+            } catch (Exception e) {
+               throw new RuntimeException("RESTEasyHTTPOperations::get request threw exception: " + target, e);
+            }
+         }
+         return result;
+      }
+
+      @Override
+      public boolean containsKey(K key) {
+         if (service.isRunning()) {
```
rvansa: OPTIONS could be more suitable here, since we don't want to transport the whole value

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,226 @@
...
+      public void put(K key, V value) {
+         if (service.isRunning()) {
+            try {
+               String target = buildUrl(service.cacheName, key);
+               Invocation put = service.getHttpClient().target(target).request().accept(service.getContentType())
+                     .buildPut(Entity.entity(encodeObject(value), service.getContentType()));
+               Response response = put.invoke();
+               int status = response.getStatus();
+               response.close();
+               if (status != Status.OK.getStatusCode()) {
```
rvansa: 201 and 204 are fine, too

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
rvansa: Or rather 202 -> throw exception as the contract (atomicity) was not realized.
We should log an ISPN JIRA for header that specifies whether we want to get 200(getAndRemove) or 202 (remove) value.

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
rvansa: Sorry, not OPTIONS, but HEAD

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,226 @@
...
+               throw new RuntimeException(e);
+            } catch (IllegalArgumentException e) {
+               throw new RuntimeException(e);
+            }
+         }
+      }
+
+      @Override
+      public V getAndPut(K key, V value) {
+         // Not Atomic
```
rvansa: Let's rather hope that the request will return the value in entity body, and if it does not, throw an exception.

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,226 @@
...
+      @Override
+      public boolean remove(K key) {
+         if (service.isRunning()) {
+            String target = buildUrl(service.cacheName, key);
+            Invocation delete = service.getHttpClient().target(target).request().accept(service.getContentType())
+                  .buildDelete();
+            Response response = delete.invoke();
+            int status = response.getStatus();
+            response.close();
+            if (status != Status.OK.getStatusCode()) {
```
rvansa: Why do you log a warning? And the method should not return true on 204. Just do a switch (status) { 200 => true, 204 => false, other => exception }

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPService.java
```diff
@@ -0,0 +1,198 @@
...
+   protected List<InetSocketAddress> servers;
+
+   @Property(doc = "Timeout for socket. Default is 30 seconds.", converter = TimeConverter.class)
+   protected long socketTimeout = 30000;
+
+   @Property(doc = "Timeout for connection. Default is 30 seconds.", converter = TimeConverter.class)
+   protected long connectionTimeout = 30000;
+
+   @Property(doc = ". Default is Integer.MAX_VALUE.")
+   protected int maxConnections = Integer.MAX_VALUE;
```
rvansa: minor thing: I would use 0 == unlimited and not call the sizing methods at all (I assume that unlimited is default). Implementation might differ for limited(max integer)/unlimited pools.

rvansa: And 'unlimited' in documentation string sounds a bit better.

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPService.java
```diff
@@ -0,0 +1,198 @@
...
+   /**
+    * Distributes HTTP requests among servers.
+    * 
+    * @return InetSocketAddress of the next server to use.
+    */
+   public InetSocketAddress nextServer() {
+      return servers.get((nextIndex.getAndIncrement() & Integer.MAX_VALUE) % servers.size());
+   }
+
+   private static class AddressListConverter implements Converter<List<InetSocketAddress>> {
```
rvansa: Looks like copy-paste from HotRod - could you move that to utils and reuse it?

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPService.java
```diff
@@ -0,0 +1,198 @@
...
+@Service(doc = "ApacheHTTP REST client")
+public class RESTEasyHTTPService implements Lifecycle {
+   private static final Log log = LogFactory.getLog(RESTEasyHTTPService.class);
+   private static final String PATTERN_STRING = "(\\[([0-9A-Fa-f:]+)\\]|([^:/?#]*))(?::(\\d*))?";
+   private static final Pattern ADDRESS_PATTERN = Pattern.compile(PATTERN_STRING);
+   private static final int DEFAULT_PORT = 8080;
+   private ResteasyClient httpClient = null;
+
+   @Property(doc = ". Defaults to 'rest'.")
+   private String serverLocation = "rest";
```
rvansa: Could we use serverPath, or rootPath or something like that? And please add more descriptive doc.

plugins/process/src/main/resources/unix-func.sh
alanfx: Mac OS X exhibits this behavior

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
alanfx: Thanks for posting this doc, @rvansa ! I'll read through to make sure I understand your suggestion here.

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
alanfx: Yep, that's a good idea

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPService.java
alanfx: Yes, as you found I missed a couple fo property descriptions in this version of the PR

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPService.java
alanfx: Yes, I'll do that

alanfx: Actually, this is more difficult. The code is shared with the SpyMemcachedService, and the issue is that the value of DEFAULT_PORT is different for memcached and REST clients. Converters are specified by class, so it isn't possible to parametrize this, unless you want to have 3 utility classes? (BaseAddressConverter, RESTAddressConverter, and MemcachedAddressConverter)

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
alanfx: From the document you sent above about DELETE: http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7/

"A successful response SHOULD be 200 (OK) if the response includes an entity describing the status, 202 (Accepted) if the action has not yet been enacted, or 204 (No Content) if the action has been enacted but the response does not include an entity."

It seems to me that the behavior should be:

200=>true
202=>false (Since the delete hasn't happened yet)
204=>true (Since the delete has happened, but no entity is returned)

Does that make sense, or am I reading that wrong?

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPService.java
alanfx: OK, this makes sense too. The defaults are 0 in RESTEasy as well.

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPService.java
alanfx: Done

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
alanfx: So, the return codes you are referring to are for the DELETE method?

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
alanfx: I'm not sure I am following you here. You think the put should return the previous value as an entity?

rvansa:
Please remove the label once ready.

alanfx:
Hey @rvansa, I think I have resolved all of your requests! Ready for review again

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,263 @@
...
+         }
+      }
+
+      @Override
+      public boolean containsKey(K key) {
+         if (service.isRunning()) {
+            String target = buildUrl(service.cacheName, key);
+            try {
+               Invocation head = service.getHttpClient().target(target).request().build(HttpMethod.HEAD);
+               head.invoke();
```
rvansa: Are you sure this throws NotFoundException? According to my javadoc, this is thrown by invoke(Class) and invoke(GenericType) but not this variant that returns response...

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,263 @@
...
+      public void put(K key, V value) {
+         putInternal(key, value, null);
+      }
+
+      private void putInternal(K key, V value, EntityTag eTag) {
+         if (service.isRunning()) {
+            try {
+               String target = buildUrl(service.cacheName, key);
+               Invocation put = null;
+               if (eTag != null) {
```
rvansa: Minor: duplicit code, I would use
```
Invocation.Builder builder = service.getHttpClient().target(target).request().accept(service.getContentType());
if (eTag != null) builder.header("If-Match", eTag.getValue())
Invocation put = .buildPut(Entity.entity(encodeObject(value), service.getContentType()));
```

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,263 @@
...
+         if (service.isRunning()) {
+            if (containsKey(key)) {
+               HTTPCache<K, V>.WrappedValue<V> wrap = getInternal(key);
+               eTag = wrap.eTag;
+               prevValue = wrap.value;
+            }
+            if (eTag == null) {
+               put(key, value);
+            } else {
+               putInternal(key, value, eTag);
```
rvansa: This throws exception upon concurrent modification between get and put. That's one way to do it; in other plugins we often loop until the optimistic step succeeds.
Is there any way how you could distinguish request failing due to If-Match and failure due to other reasons?

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
```diff
@@ -0,0 +1,263 @@
...
+         V value = null;
+         EntityTag eTag = null;
+         if (service.isRunning()) {
+            String target = buildUrl(service.cacheName, key);
+            try {
+               Invocation get = service.getHttpClient().target(target).request().accept(service.getContentType())
+                     .buildGet();
+               Response response = get.invoke();
+               eTag = response.getEntityTag();
+               value = decodeByteArray(response.readEntity(byte[].class));
```
rvansa: Minor: you could decode this lazily, only if the subsequent operation (in getAndPut/getAndRemove) succeeds.

rvansa:
Great, Alan. Please check the containsKey, it can go in.

plugins/resteasy-http/src/main/java/org/radargun/service/RESTEasyHTTPOperations.java
alanfx: It is possible to detect that the failure is due to the If-Match header, because the put or remove will return a 412 (Precondition Failed) response code. The question is whether retrying the operation is correct? None of the other operations are retried, so in my opinion, the current code is the correct behavior. What do you think?

alanfx:
@rvansa updated. The containsKey code was wrong, so thanks for spotting that!

alanfx:
@rvansa PING!

rvansa:
Eh, sorry... integrated into 2.x, please create 3.0 PR

Issue #327, opened 2016-01-28
Remove partial throughput charts for different operation types from report 


mcimbora:
- Based on https://github.com/radargun/radargun/wiki/Understanding-stress-test-results-in-RadarGun-2. Let's provide one throughput chart instead. Partial RTM charts should stay.

mgencur:
This has been integrated in 2.x branch.

PR #337, opened 2016-02-17
issue #327 Remove partial throughput charts for different operation types from report, 2.x

mcimbora:
- Backported templating PR for HTML reporter 
- Introduced logical operation groups
-- Defined on stage level
-- Specifies which operations should be treated as a single group (especially in reporting) - ATM only throughput chart will be generated for the group, while avoiding generation of Response time mean charts for partial operations, to avoid problems specified in https://github.com/radargun/radargun/wiki/Understanding-stress-test-results-in-RadarGun-2 - this applies to operations which define ratio of executed operations
- Added simple debugging test for HTMLReporter

### DIFF
--- a/core/src/main/java/org/radargun/stats/Statistics.java
+++ b/core/src/main/java/org/radargun/stats/Statistics.java
@@ -2,6 +2,7 @@ package org.radargun.stats;
 
 import java.io.Serializable;
 import java.util.Map;
+import java.util.Set;
 
 import org.radargun.Operation;
 import org.radargun.utils.ReflexiveConverters;
@@ -54,6 +55,19 @@ public interface Statistics extends Serializable {
    Statistics copy();
 
    /**
+    * Creates a logical group of operations identified by given group name. Should be used on per-stage basis.
+    *
+    * @param name Name of the group
+    * @param operations Operations to include in logical group
+    */
+   void registerOperationsGroup(String name, Set<Operation> operations);
+
+   /**
+    * @return Group name for supplied operation. Returns null if group with given operation is not registered.
+    */
+   String getOperationsGroup(Operation operation);
+
+   /**
     * Add the measurements collected into another instance to this instance.
     * @param otherStats Must be of the same class as this instance.
     */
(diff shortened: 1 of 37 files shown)


rvansa:
Haven't checked the FreeMarker stuff, but overally it's fine. Nevertheless, please let's try to keep the statistics object as lightweight as possible, therefore, share what's likely to be identical. I have already ran many times into OOMEs when copying results to master.

mcimbora:
@rvansa Updated. I removed merging of registered groups from the stats (as they are the same within a single test). The registered groups are sent only once as well. Some complexity was added by this change, let's decide whether it's worth it.

mcimbora:
https://github.com/radargun/radargun/pull/337/commits/4f06353b1e868b08429aec5ec1c5c18e9e141df7

core/src/main/java/org/radargun/stats/Statistics.java
```diff
@@ -54,6 +55,24 @@
    Statistics copy();
 
    /**
+    * Creates a logical group of operations identified by given group name. Should be used on per-stage basis.
```
rvansa: When this is on a per-stage basis, why do you add it to Statistics which have multiple instances in each stage (typically one for each thread)? Why can't you put this just to the stage itself? It's set to report from acks anyway.

rvansa:
Sorry about taking so long to handle this. Do you still have time to work on this? And will you port it to 3.0 branch once it is integrated?

mcimbora:
@rvansa Sorry, no time at the moment. I'll try to find some in the upcoming weeks (or give guidance to someone who would like to take this over).

mcimbora:
@rvansa @mgencur I recommend integrating the PR as-is right now as the current implementation is sufficient. Optimizations can be done later if required.

mgencur:
@rvansa @alanfx  I suggest integrating this PR into 2.x branch as is. We can do optimizations later. When this is integrated we'll create a similar fix for the master branch and we can possibly address the comments there (if really required). We should really go forward with master branch and abandon 2.x as soon as possible. Maintaining and using two branches is not feasible anymore.

alanfx:

PR #343, opened 2016-03-09
CacheManager needs to be started after the service has a reference to it

jmarkos:
With the newest infinispan, the listener for the view changes (InfinispanClustered) is called during the cachemanager.start() in Infinispan51EmbeddedService, but fails, because the reference to cacheManager in InfinispanEmbeddedService at line 113 is still null at the time the listener is called.


jmarkos:
Updated (amend + push -f, so Matej's comment is gone).

mcimbora:
@jmarkos Integrated to master as well. Thanks!

PR #344, opened 2016-03-31
Process plugin overhaul and RESTEasy client plugin

alanfx:


alanfx:
I am getting the following failure now that I am trying to figure out:

{noformat}
14:53:32,153 ERROR [org.radargun.Slave] (sc-main) Unexpected error in scenario
java.lang.IllegalArgumentException: Cannot load class org.radargun.service.RESTEasyHTTPService from plugin resteasy-http
	at org.radargun.ServiceHelper.createService(ServiceHelper.java:76) ~[radargun-core-3.0.0-SNAPSHOT.jar:?]
	at org.radargun.SlaveBase.scenarioLoop(SlaveBase.java:41) ~[radargun-core-3.0.0-SNAPSHOT.jar:?]
	at org.radargun.SlaveBase$ScenarioRunner.run(SlaveBase.java:155) [radargun-core-3.0.0-SNAPSHOT.jar:?]
Caused by: java.lang.ClassNotFoundException: org.radargun.service.RESTEasyHTTPService
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381) ~[?:1.8.0_72]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424) ~[?:1.8.0_72]
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331) ~[?:1.8.0_72]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357) ~[?:1.8.0_72]
	at java.lang.Class.forName0(Native Method) ~[?:1.8.0_72]
	at java.lang.Class.forName(Class.java:264) ~[?:1.8.0_72]
	at org.radargun.ServiceHelper.createService(ServiceHelper.java:74) ~[radargun-core-3.0.0-SNAPSHOT.jar:?]
	... 2 more
java.lang.IllegalArgumentException: Cannot load class org.radargun.service.RESTEasyHTTPService from plugin resteasy-http
	at org.radargun.ServiceHelper.createService(ServiceHelper.java:76)
	at org.radargun.SlaveBase.scenarioLoop(SlaveBase.java:41)
	at org.radargun.SlaveBase$ScenarioRunner.run(SlaveBase.java:155)
Caused by: java.lang.ClassNotFoundException: org.radargun.service.RESTEasyHTTPService
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at org.radargun.ServiceHelper.createService(ServiceHelper.java:74)
	... 2 more
{noformat}

alanfx:
Ok, this is really strange. If I run the `benchmark-cs.xml` as is with all three configurations, it runs to completion. If I remove the HotRod and Memcached configurations from the file, then the benchmark fails with the CNFE because the classpath for the second node only includes the JARs from the infinispan80 plugin instead of the resteasy-http plugin. @rvansa???

alanfx:
Figured it out. In org.radargun.Slave in the ```} else if (object instanceof RemoteSlaveConnection.Restart) {``` clause, slaveIndex is `-1`, so it should be retrieved from the state instead. Added a commit to the PR with this change

rvansa:
My fault, then. I always use dist.sh script for starting the benchmark, so the slave indexes are set correctly. Thanks for the fix.

rvansa:
Integrated, thanks!

alanfx:
@rvansa Hey, so is there any reason that an index of -1 is valid? I was wondering why `Cluster.getGroup()` handles -1 instead of treating it as an error? Thanks

rvansa:
No, -1 is the value slave gets from command line when --slaveIndex is not passed. Then it tells master 'give me any slave index'. If --slaveIndex is set, slave signalizes to master a preference for certain index.

alanfx:
I guess it seems to me that `Cluster.getGroup()` should thrown an error if the index is `-1` and callers should make sure to pass a valid index. The code at this point, basically converts `-1` to `0`, which I am not sure is correct?

rvansa:
Right, the check would make sense there.

alanfx:
OK, I'll add it to an upcoming PR! Thanks

PR #345, opened 2016-04-04
Refactored statistics recording API

rvansa:
There are several ideas that have driven this change:
* encapsulate time measurement into separate class
* differentiate the concept of RPC, aggregated operations (e.g. transactions) and one-way messages
* prepare PeriodicStatistics to be more light-weight (in the future I don't want to call `currentTimeMillis` for each request)
* prepare code for async RPCs


rvansa:
* Remove IterationData hack, introduce representations *.Series
* Make PeriodicStatistics merge-able
* Rename DefaultStatistics and DefaultOperationStats to BasicStatistics and BasicOperationStats
* Add CommonStatistics that works as a handy template for periodic statistics + histogram
* Refactor Statistics.getOperationStats + OperationStats.getRepresentation -> Statistics.getOperations + Statistics.getRepresentation
* Add the series charts to HTML report

core/src/main/java/org/radargun/stats/RequestSet.java
```diff
@@ -0,0 +1,57 @@
...
+   public void add(Request request) {
+      if (!request.isFinished()) {
+         throw new IllegalArgumentException();
+      }
+      begin = Math.min(begin, request.getRequestStartTime());
+      end = Math.max(end, request.getResponseCompleteTime());
+      sumDurations += request.duration();
+   }
+
+   public void finished(boolean successful, Operation operation) {
```
mcimbora: Looks redundant.

mcimbora: Ok, I see the point now. Definitely worth documenting.

rvansa: What? It's just a convenience method when you get the result as boolean, so you pass that instead of doing an if.
However, now I've noticed that the assignment must happen before record call.

rvansa:
@mcimbora I've fixed the order in RequestSet, do you have any other concerns?

rvansa:
@jmarkos Could you do the review and integrate, please? Matej briefly went through the changes, but you should start taking care.

core/src/main/java/org/radargun/stats/Message.java
```diff
@@ -0,0 +1,49 @@
...
+ * Tracks time for operations with different origin and destination nodes.
+ * The time can be tracked only using wall-clock time.
+ *
+ * @author Radim Vansa &lt;rvansa@redhat.com&gt;
+ */
+public final class Message {
+   private final Statistics statistics;
+   private long sendStartTime = Long.MAX_VALUE;
+   // private long sendCompleteTime; // hard to track when the message itself carries the timestamps
+   // private long receiveStartTime;
```
jmarkos: are these comments intentional?

core/src/main/java/org/radargun/stats/MultiStatistics.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.radargun.Operation;
+
+/**
+ * @author Radim Vansa &lt;rvansa@redhat.com&gt;
+ */
+public abstract class MultiStatistics implements Statistics {
```
jmarkos: Do you expect more subclasses in future? Because right now this has only 1 subclass.

core/src/main/java/org/radargun/stats/RequestSet.java
```diff
@@ -0,0 +1,61 @@
+package org.radargun.stats;
+
+import org.radargun.Operation;
+
+/**
+ * @author Radim Vansa &lt;rvansa@redhat.com&gt;
+ */
```
jmarkos: Some short documentation could be useful - it appears it's only used for transactions, is that right?

extensions/cache/src/main/java/org/radargun/stages/mapreduce/MapReduceStage.java
```diff
@@ -290,31 +288,28 @@ public DistStageAck executeOnSlave() {
...
-
       MapReduceAck ack = new MapReduceAck(slaveState);
       try {
          if (collatorFqn != null) {
-            start = TimeService.nanoTime();
+            Request request = stats.startRequest();
             payloadObject = task.executeWithCollator();
-            durationNanos = TimeService.nanoTime() - start;
-            stats.registerRequest(durationNanos, MapReducer.MAPREDUCE_COLLATOR);
+            // TODO handle exception
```
jmarkos: todo (and a few lines below is another one)

reporters/reporter-default/src/main/java/org/radargun/reporting/html/LineChart.java
```diff
@@ -42,12 +43,13 @@ public LineChart(String domainLabel, String rangeLabel) {
...
+      boolean drawShapes = IntStream.range(0, dataset.getSeriesCount()).map(dataset::getItemCount).max().orElse(0) < 10;
+      DeviationRenderer renderer = new DeviationRenderer(true, drawShapes);
       renderer.setAlpha(0.2f);
       chart.getXYPlot().setRenderer(renderer);
-      CategorizedAxis domainAxis = new CategorizedAxis();
-      domainAxis.setLabel(domainLabel);
-      chart.getXYPlot().setDomainAxis(domainAxis);
+//      CategorizedAxis domainAxis = new CategorizedAxis();
+//      domainAxis.setLabel(domainLabel);
+//      chart.getXYPlot().setDomainAxis(domainAxis);
```
jmarkos: comments?

extensions/cache/src/main/resources/benchmark-analysis.xml
```diff
@@ -56,12 +56,12 @@
       <l:basic-operations-test test-name="on-max-throughput"
                              duration="${test.long.duration:1m}" total-threads="#{ 10 * ${top.iteration} + 10 }">
          <statistics>
-            <default>
+            <basic>
```
jmarkos: I don't feel the rename was worth it to break the existing rg3 benchmarks, but I guess it's not a big deal.

jmarkos:
An example benchmark using periodic statistics with the series-representations could be useful, I don't see for what this is supposed to be used. It could also verify it's doing what you intended it to do.

core/src/main/java/org/radargun/stats/Message.java
rvansa: Yes, those are expected to be present when we'll have traits with asynchronous API.

core/src/main/java/org/radargun/stats/MultiStatistics.java
rvansa: It's more about a separation of concerns. MultiStatistics does multiplex all the calls, while CommonStatistics is kind of template for MultiStatistics - it defines concrete mix of stats.

core/src/main/java/org/radargun/stats/RequestSet.java
rvansa: Ack, docs can't hurt. It is used for transactions, but it should be used for any grouped requests.

extensions/cache/src/main/java/org/radargun/stages/mapreduce/MapReduceStage.java
rvansa: Matej is about to update MR stages - there was no exception handling and I haven't added one, but marked this place as missing it... I've focused just on the API change.

extensions/cache/src/main/resources/benchmark-analysis.xml
rvansa: I wasn't comfortable with something called 'default' not being the default anymore.

alanfx:
I have used this branch and it the new statistics are working well. If you address @jmarkos' comments, then I think this can be integrated, @rvansa.

rvansa:
Rebased, added few fixes in testsuite, and the desired example benchmark.

rvansa:
Rebased again on last master.

jmarkos:
Looks OK (after another rebase).
Do I have rights to push to master?

mcimbora:
@jmarkos Yes.

jmarkos:
@mcimbora great!

Integrated, thanks!

PR #346, opened 2016-04-05
A set of small fixes

alanfx:
I'll create a similar PR for branch_2.x once this is integrated


core/src/main/java/org/radargun/config/Cluster.java
```diff
@@ -53,23 +53,27 @@ public int getSize() {
    }
 
    public Group getGroup(int slaveIndex) {
-      int index = slaveIndex;
-      for (Group g : groups) {
-         if (index < g.size) {
-            return g;
+      if (slaveIndex != -1) {
```
rvansa: slaveIndex < 0 would be better

plugins/jdg70/src/main/resources/server.xml
```diff
@@ -0,0 +1,399 @@
+<?xml version="1.0" ?>
+
+<server xmlns="urn:jboss:domain:4.0">
```
rvansa: Is there any difference to the previous config? I am thinking about symlinks rather than copy-paste.

rvansa: I would also consider XPath modifications on the default config - storing only the list of modifications in RG sources, rather than the whole file.

core/src/main/java/org/radargun/config/Cluster.java
alanfx: I'm not sure I agree. `-1` is the value used when no index has been assigned, so I think it is correct to check for that value, unless it is possible that some other value can be set?

plugins/jdg70/src/main/resources/server.xml
alanfx: I just copied the file from the ER2 build, so it could change. I'm not sure what you mean by symlinking and keeping the modifications in RG sources. You mean doing XML transformations in RG?

rvansa: To be clear, I don't require the above in this PR, just my eyes hurt from hundreds of lines just copied, and I've hijacked your PR for discussion.

So there are two ways to reduce the configuration:
a) use only one configuration file (if that's possible) and symlink other versions of plugins to that (so if you just copied this, you would use symlink instead).
b) use default server configs; as we need few modifications, keep a list of them in the conf/ dir and apply them (do the XML transformation) in RG.

core/src/main/java/org/radargun/config/Cluster.java
rvansa: -1 is just one invalid value (or value with meaning of 'unset' value), but as you already added a check, you should check all 'unset' values. But this is just a nitpicking, I'll integrate this soonish.

plugins/jdg70/src/main/resources/server.xml
alanfx: From my point of view, using the default server configs for each plugin would be really useful. That's all I am doing at this point is copying in the default and avoiding this step when creating a new plugin would be helpful. What we do for changes to the default is a separate discussion. I'm not opposed to adding XML transformation but using a separate config file in the `etc` project is also a possibility.

core/src/main/java/org/radargun/config/Cluster.java
alanfx: OK, if there are other places in the code that use a default value besides `-1`, I am happy to change it. Can you point me to the other places where there are 'unset' values?

rvansa:
Thanks, integrated.

PR #358, opened 2016-05-04
Remove default JVM/JGroups options

mcimbora:


mcimbora:
This is necessary to avoid clashes of properties passed externally.

PR #360, opened 2016-05-09
Add RG support for Apache Spark

mcimbora:


mcimbora:
Moved MapReduce stage to a separate extension.

extensions/mapreduce/src/main/java/org/radargun/stages/mapreduce/MapReduceStage.java
```diff
@@ -121,20 +126,31 @@
...
+   @InjectTrait
    private BasicOperations basicOperations;
 
    @InjectTrait(dependency = InjectTrait.Dependency.OPTIONAL)
    private Iterable iterable;
 
    private boolean supportsResultCacheName = false;
 
+   @Init
+   public void validate() {
```
rvansa: Why do you use init method instead of the trait-dependency mechanism?

rvansa: Ah, I see, because of the 'store result in cache' functionality. Hey, could you store the result temporarily in slave state and do the store operation in another stage? Ideally, I would remove the dependency between mapreduce and cache extensions alltogether.

extensions/mapreduce/src/main/resources/benchmark-spark-mr.xml
```diff
@@ -0,0 +1,107 @@
+<!-- RadarGun 3.0 benchmark -->
+<benchmark xmlns="urn:radargun:benchmark:3.0">
+
+   <!-- Specifies where should the master open socket -->
+   <master bindAddress="${master.address:127.0.0.1}" port="${master.port:2103}"/>
+
+   <!-- List of cluster configurations where the benchmark should run -->
+   <clusters>
+      <cluster size="5">
+         <group name="master" size="1" />
```
rvansa: Would it be possible to write a benchmark that would do apple-apple comparison of Spark vs. Infinispan (or generally anything that has MR)?

extensions/mapreduce/src/test/java/org/radargun/util/MapReduceTraitRepository.java
```diff
@@ -0,0 +1,133 @@
...
+      private static class Builder implements org.radargun.traits.MapReducer.Builder {
+
+         private ConcurrentHashMap cache;
+
+         public Builder(ConcurrentHashMap cache) {
+            this.cache = cache;
+         }
+
+         @Override
+         public org.radargun.traits.MapReducer.Builder distributedReducePhase(boolean distributedReducePhase) {
```
rvansa: Java allows covariance in method return types, just use Builder.

plugins/infinispan52/src/main/java/org/radargun/service/HotRodOperations.java
```diff
@@ -188,5 +205,75 @@ public boolean putIfAbsent(K key, V value, long lifespan, long maxIdleTime) {
...
+               try {
+                  V value = entry.getValue().get();
+                  if (value != null) {
+                     result.put(entry.getKey(), value);
+                  }
+               } catch (Exception e) {
+                  throw new RuntimeException(e);
+               }
+            }
+         } else {
```
rvansa: `preferAsync` means that async ops should be _preferred_ to native getAll operation. If there's not native getAll operations, it should still do the async version, not retrieve the values sequentially.

plugins/infinispan52/src/main/java/org/radargun/service/HotRodOperations.java
```diff
@@ -188,5 +205,75 @@ public boolean putIfAbsent(K key, V value, long lifespan, long maxIdleTime) {
...
+            }
+         }
+         return result;
+      }
+
+      @Override
+      public void putAll(Map<K, V> entries) {
+         if (trace) log.tracef("PUT_ALL cache=%s keys=%s", noReturn.getName(), entries);
+         if (preferAsync) {
+            noReturn.putAllAsync(entries);
```
rvansa: Incorrect, this should call noReturn.putAsync in a loop (and then wait for all the futures), not just do the operation asynchronously.

plugins/infinispan52/src/main/java/org/radargun/service/HotRodOperations.java
```diff
@@ -188,5 +205,75 @@ public boolean putIfAbsent(K key, V value, long lifespan, long maxIdleTime) {
...
+            }
+            for (Map.Entry<K, Future<V>> entry : futureMap.entrySet()) {
+               try {
+                  entry.getValue().get();
+               } catch (Exception e) {
+                  throw new RuntimeException(e);
+               }
+            }
+         } else {
+            for (K key : keys) {
```
rvansa: Same as with get.

plugins/infinispan53/pom.xml
```diff
@@ -92,6 +92,14 @@
       </dependency>
 
       <dependency>
+         <groupId>org.radargun</groupId>
+         <artifactId>radargun-mapreduce</artifactId>
+         <version>${project.version}</version>
+         <scope>provided</scope>
+         <!--<optional>true</optional>-->
```
rvansa: remove that

rvansa:
My biggest concern is that if the implementation requires plugin-dependent classes related to the task itself (word-count), the design totally ignores the primary purpose of RG, and is conceptually wrong.

extensions/mapreduce/src/main/java/org/radargun/stages/mapreduce/MapReduceStage.java
alanfx: Hey @rvansa, I'm not sure that Map/Reduce can be divorced from the cache. Distributed Execution can be, but not Map/Reduce.

rvansa: Spark does not have the cache, so it should be separable. Why do you think you can't separate those? It seems to me that the code dealing with cache here should just do what Infinispan would do internally if you some option was supported; why can't that be moved to another stage?
Anyway, I would really love to see a single benchmark that would do the same MR job in both Infinispan and Spark. And I mean the same job, not different implementations of WordCount.

mcimbora: @alanfx I agree with @rvansa here. As long as we're moving MR to a separate module, we shouldn't impose storing the results in a cache as we want to make the extension as generic as possible.

alanfx: @rvansa and @mcimbora I have no problem removing the ability to store the results in a cache. I don't think it is ever used in any of our other benchmarks. However, I do see a distinction about where the input data comes from. The Spark plugin does't need to reply on the cache, because the Spark connector uses HotRod to get the input data from a RemoteCache running in an Infinispan/JDG server. Map/Reduce's input data is coming from a cache, so I think it still needs to rely on the cache, right? Or am I missing some other way that the input data can be retrieved? 
I'm also not sure how the same job can be used for Infinispan and Spark unless they support the same API, and I thought that's what the distributed streams API allowed?

rvansa: The implementation will naturally depend on the existence of a cache, be it remote or embedded one. But the MR stage shouldn't require cache-related traits. The source of the data will be just and identifier (e.g. cache name), and the implementation of the trait is free to interpret it.

The whole point of Traits is to cover vendor-specific API (in as lightweight wrapper as possible), and let benchmark code operate only on this wrapped one.

To put both into one benchmark, I would be more troubled about the setup, since Infinispan would have some groups just mocked, and when X is in Spark executed on slave A and Y on slave B, in Infinispan both should go to single node.

alanfx: OK, I see where you are going with this, and I agree that passing the cache name in is sufficient for the Map/Reduce task. However, I'll also say that I added the ability to store results in the cache, but then it became an Infinispan Map/Reduce feature. In other words, the trait is covering the vendor-specific API and allowing benchmark writers to use this functionality even if the vendor doesn't support it. OTOH, Map/Reduce is going away in favor of Streams, which doesn't have this feature.

mcimbora:
@rvansa Changes sent

extensions/mapreduce/src/main/java/org/radargun/traits/MapReducer.java
```diff
@@ -74,7 +48,7 @@
        *           parameters for the Mapper object
        * @return this builder instance
        */
-      Builder<KOut, VOut, R> mapper(String mapperFqn, Map<String, String> mapperParameters);
+      Builder<KOut, VOut, R> mapper(String mapperFqn, Collection<KeyValueProperty> mapperParameters);
```
rvansa: Any reason not to use java.util.Properties?

mcimbora: MapReduceStage uses `KeyValueProperty` type to specify the mapper/reducer/... params. No need to convert this twice.

https://github.com/radargun/radargun/pull/360/files#diff-faaddb54bc834e3a91fddc93c42acba1R61

rvansa:
Integrated with a warning notice. Thanks.

PR #362, opened 2016-05-12
Plugin ISPN80 should depend on ISPN80

vjuranek:


mcimbora:
@vjuranek Integrated, thanks

PR #366, opened 2016-06-02
infinispan90 plugin + fixes

jmarkos:


rvansa:
@jmarkos I've merged the tools.jar commit, ad updated version in jdg66. Not sure if we should push ER versions here.
The CQ integration needs rebase, though.

alanfx:
PR https://github.com/radargun/radargun/pull/377 also includes the jdg66 version change, so I'll remove that. The jdg70 version should not be changed to an ER version. It should use the line that is deleted in the commit.

jmarkos:
I wasn't sure about the ER version, but was annoyed that master couldn't be built since GA wasn't released yet. Now that we have GA it doesn't matter.

jmarkos:
Rebased here https://github.com/radargun/radargun/pull/382

Issue #314, opened 2015-12-01
Log tailing is broken on slaves


vjuranek:
When log tail is turned on (with `-t` option) on slaves, it doesn't work any more and the output is stopped almost immediately, just after acquiring lock:
```
15:08:00,937 INFO  [org.radargun.RemoteMasterConnection] (main) Response successfully sent to the master
15:08:00,937 INFO  [org.radargun.ShutDownHook] (Thread-0) Slave process is being shutdown
15:08:00,999: Waiting for lock on file /tmp/restart-02360494704594478985.tmp
15:08:01,253: Lock acquired
```

alanfx:
The issus is that the PID of the process changes when the slave processes are restarted, so the tailing stops. Removing the `--pid $SLAVE_PID` allows tailing the log file to work, but it no longer exits when the benchmark completes.

vjuranek:
I tested the fix with Jenkins RadarGun plugin and everything seems to work fine - nice work @alanfx ! I guess we can close this issue (and eventually create new one if some issues appear in the future).

alanfx:
Thanks, @vjuranek. I think there still might be some corner cases where the tail is not ended, but the logs should be complete at this point. I'm still considering changes to RadarGun to fix this more completely.

alanfx:
There is definitely something wrong with the bash log tailer around the timeouts, so I am reopening this issue.

alanfx:
https://github.com/radargun/radargun/pull/378

alanfx:
Resolving this issue based on the PR above...unless I hear otherwise!

PR #369, opened 2016-06-06
Add properties to VmArgs and partial fix for issue #314

alanfx:


alanfx:
@jmarkos The `tail` command in the slave.sh script never exits, and then the jobs keep running in Jenkins. I think @vjuranek's suggestion in https://github.com/vjuranek/radargun-plugin/issues/35 will help the situation.

core/src/main/java/org/radargun/config/VmArgs.java
```diff
@@ -28,6 +33,9 @@
    @Property(doc = "Ignore all VM arguments passed to slave and use only those specified here. Default is false.")
    private boolean ignoreDefault = false;
 
+   @Property(doc = "Print all VM arguments. Default is false.")
+   private boolean printFlagsFinal = false;
```
rvansa: Please use Boolean instead of boolean; we need a tri-state logic to set the parameter to enabled/disabled or not set it at all (null).

core/src/main/java/org/radargun/config/VmArgs.java
```diff
@@ -139,10 +158,32 @@ private static void ensureArg(Collection<String> args, String arg) {
       @Property(doc = "Min memory", converter = SizeConverter.class)
       private Long min;
 
+      @Property(doc = "Eden memory", converter = SizeConverter.class)
+      private Long eden;
```
rvansa: This is misleading; -Xmn sets the size of whole new generation, which consists of eden AND survivor spaces. The space is split between Eden and Survivor spaces according to -XX:SurvivorRatio
Source: https://docs.oracle.com/cd/E19900-01/819-4742/abeik/index.html

core/src/main/java/org/radargun/config/VmArgs.java
```diff
@@ -139,10 +158,32 @@ private static void ensureArg(Collection<String> args, String arg) {
...
-         if (min != null) replace(args, "-Xms", String.valueOf(min));
-         if (max != null) replace(args, "-Xmx", String.valueOf(max));
+         if (min != null)
+            replace(args, "-Xms", String.valueOf(min));
+         if (max != null)
+            replace(args, "-Xmx", String.valueOf(max));
+         if (eden != null)
+            replace(args, "-Xmn", String.valueOf(eden));
+         if (threadStackSize != null)
+            replace(args, "ThreadStackSize", String.valueOf(threadStackSize));
```
rvansa: -Xss is probably more 'portable' flag: http://mail.openjdk.java.net/pipermail/hotspot-dev/2011-June/004272.html

rvansa: And you probably wanted to use -XX:ThreadStackSize

alanfx: Yeah, changed to `-Xss`

core/src/main/java/org/radargun/config/VmArgs.java
alanfx: Done

core/src/main/java/org/radargun/config/VmArgs.java
alanfx: So should I rename the property from `eden` to `young`?

alanfx:
@rvansa I have addressed your comments on VmArgs. @vjuranek and @jmarkos, I think I have a more complete solution for issue #314 now.

alanfx:
Updated the solution for #314 to use a 10 minute timeout when reading from the tail process. If the timeout occurs and no java process spawned by the shell script is found, then the tail process is killed.

rvansa:
Integrated, thanks!

PR #374, opened 2016-07-05
Fix KeyExpirationTestStage with transactions

alanfx:
This fix makes transactionEnded() the same as in
BasicOperationsTestStage. Previous to this change, the stage would not
run on a transactional cache.

### DIFF
--- a/core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
+++ b/core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
@@ -221,8 +221,8 @@ public class LegacyStressor extends Thread {
    }
 
    private void clearTransaction() {
-      ongoingTx = null;
       logic.transactionEnded();
+      ongoingTx = null;
    }
 
    public int getThreadIndex() {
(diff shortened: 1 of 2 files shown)


alanfx:
@rvansa, @mcimbora, @jmarkos, can you guys verify that this change makes sense? Thanks

mgencur:
I'm not exactly sure what the problem was but I suspect it was a NPE? Your fix should fix this particular case but I think the right fix is to change method in LegacyStressor. Currently it looks like this
private void clearTransaction() {
      ongoingTx = null;
      logic.transactionEnded();
   }
First it sets ongoingTx to null and later in transactionEnded it uses the ongoingTx through a wrap method. 
Setting ongoingTx to null should be placed after calling transactionEnded, IMO.

alanfx:
Hey @mgencur. Yes the problem was an NPE, and I think you are correct about the code in `clearTransaction`. I'll check the other test stages and update this PR. Thanks!

alanfx:
@mgencur I added your suggested change to this PR. This change in `LegacyStressor` by itself was not enough to fix the NPE issue in the `KeyExpirationTestStage` though.

rvansa:
Looks good, thanks!

PR #376, opened 2016-07-14
RG-375 Use BigDecimal for Fuzzy entry size computation

mgencur:


mgencur:
Hmm. Wait. I will send a fixed version of this PR.

mgencur:
Fixed.

core/src/main/java/org/radargun/utils/Fuzzy.java
```diff
@@ -51,57 +55,69 @@ public String toString() {
...
+            || (cumulatedProbability.compareTo(BigDecimal.ONE) == 0 && weights.size() > 0) //equal to one
+            || (cumulatedProbability.compareTo(BigDecimal.ONE) == -1 && weights.size() == 0)) { //less than one
             throw new IllegalStateException("Probability: " + cumulatedProbability);
          }
-         double cumulatedWeight = 0;
-         for (Double w : this.weights) {
-            cumulatedWeight += w;
-            cumulativeProbability[i++] = (1 - cumulatedProbability) * cumulatedWeight / sumWeight;
+         BigDecimal cumulatedWeight = BigDecimal.ZERO;
+         MathContext mc = new MathContext(2, RoundingMode.HALF_UP); //required for division
```
rvansa: If rounding is involved, how can you be sure that the cumulated probabilities won't overflow 1? (which is the issue this PR fixes I think).

core/src/main/java/org/radargun/utils/Fuzzy.java
```diff
@@ -12,24 +15,25 @@
...
-   private Fuzzy(Serializable[] values, double[] probability) {
+   private Fuzzy(Serializable[] values, BigDecimal[] probabilities) {
       this.values = values;
-      this.probability = probability;
+      this.probabilities = probabilities;
    }
 
-   public static <T extends Serializable> Fuzzy<T> always(T singleValue) {
-      return new Fuzzy<T>(new Serializable[] {singleValue}, new double[] {1.0});
+   public static <T extends Serializable> Fuzzy<T> uniformEntrySize(T entrySize) {
```
rvansa: Fuzzy is generic util class, I would stick to just `uniform(T value)`.

core/src/main/java/org/radargun/utils/Fuzzy.java
mgencur: Rounding is required because some results of division can have indefinite number of decimal points. Anyway, I still set the last cumulativeProbability to 1.0 manually because I had this concern too.

core/src/main/java/org/radargun/utils/Fuzzy.java
mgencur: I can only see the usage in CacheTestSetupStage and LoadStage. In both places it is used to define uniform entry size. Can you explain where else this can be used? Right now this class is a bit too "fuzzy" to me:) It's too generic, IMO.

mgencur:
I've renamed the method to uniform as suggested. Anyway, I plan to modify this class a bit and make it less general, and send a separate pull request. This class has been there for almost three years and it is still used just for defining variable entry sizes in tests. The changes should make it more clear how it is supposed to be used and what we can expect from it. Without reading every line of the class.

mgencur:
I added the changes here instead of creating a new PR.

rvansa:
@mgencur I agree that 'fuzzy' may not be the best name, looking it up it seems that the proper term is [Probability mass function](https://en.wikipedia.org/wiki/Probability_mass_function). Though, that would be quite a long name. I am happy to replace the 'fuzzy' with anything more meaningful (do you have any suggestions?).

The fact that it's not used anywhere else is irrelevant. The PMF is a well-known concept, and using field called 'entry size' which uses this as type instead of scalar should be understandable.

I am fine with specializing this to ints, if you find generics inconvenient, but the type does not contain anything specific to entries - scalar `int` does not either.

mgencur:
The last commit removes the generics and uses Integer internally. I wanted to leave the Fuzzy word in the name because it's easy to remember. I extended this to FuzzyEntrySize. Another good name could be VariableEntrySize. I don't want to give names that require google or Wikipedia for explanation. Not everybody knows exact meaning of Probability Mass Function. And the class is not even exact implementation of this concept because one can add probabilities where the sum is less then 1 and add also weights. The overall goal is to make it easier to use. Pulling in terms from probability theory does not make it easier.

rvansa:
Pushed the first commit and added better javadoc for the class.

PR #377, opened 2016-07-21
Split JDG plugin builds

alanfx:
JDG versions 6.0.x through 6.2.x were not released to the external Maven
repository, so this PR splits those versions from the later versions. Also the `jdg` profile still builds all of the plugins Also fixed the version in the `jdg66` plugin, because it was wrong.

### DIFF
--- a/plugins/jdg63/src/main/java/org/radargun/service/JDG63EmbeddedService.java
+++ b/plugins/jdg63/src/main/java/org/radargun/service/JDG63EmbeddedService.java
@@ -18,12 +18,6 @@ public class JDG63EmbeddedService extends Infinispan60EmbeddedService {
    protected static final String SERVICE_DESCRIPTION = "Service hosting JDG in embedded (library) mode.";
 
    @Override
-   @ProvidesTrait
-   public Infinispan70MapReduce createMapReduce() {
-      return new Infinispan70MapReduce(this);
-   }
-
-   @Override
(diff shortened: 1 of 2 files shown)


alanfx:
Also the jdg profile still builds all of the plugins

mgencur:
I've sent an email to infinispan-internal which is related to this PR too.

rvansa:
Does typing `-Djdg63-70` feel convenient to you? I wouldn't mind, but having this as the plain `-Djdg` would seem more handy (the name of the old profile shouldn't matter too much).
Feel free to integrate after you give it second thought.

alanfx:
Yeah, I think you are correct `-Pjdg` will build `jdg63` to `jdg70`, and then I'll have to come up with a name to build `jdg60` to `jdg62`. I also considered adding a profile for each jdg version, and adding the dependencies in the profile. This will add a lot of profiles to the POM, though.

alanfx:
Updated the PR. Now the `jdg` profile builds `jdg63` to `jdg70` plugins, and profile `jdg-early` builds  `jdg60` to `jdg62` plugins.

Also fixes the code on master after https://github.com/radargun/radargun/pull/360 was integrated.

rvansa:
Thanks, pushed. Sorry about the #360, I got local build running fine, but haven't ran clean.

alanfx:
Radim , no worries and thanks!

PR #378, opened 2016-07-23
Fix issue 314

alanfx:
Bash read timeouts did not return an exit code greater than 127 on a
timeout. Also added some additional exit strings to master.sh.

https://github.com/radargun/radargun/issues/314


alanfx:
Still something wrong. When you start the benchmark, you get two bash processes which messes up the search for the parent PID. Will keep looking at it and reopen when it's ready

PR #379, opened 2016-07-24
Fix issue 314

alanfx:
The previous strategy of checking for processes based on the shell
process was not working, because the new java processes do not have the
expected parent PID. Instead tail_log takes a new argument that is a
search string for the 'ps -ef' command. If the read times out and the
process doesn't exist, then the shell process exits. Also added some
additional exit strings to master.sh.

https://github.com/radargun/radargun/issues/314

### DIFF
--- a/core/src/main/resources/includes.sh
+++ b/core/src/main/resources/includes.sh
@@ -102,12 +102,13 @@ tail_log() {
    # Bash log tailer
    # Param 1 - log file
    # Param 2 - String in the log file. When it is found, the tail process is killed
+   # Param 3 - The string in 'ps -ef' to search for. If it doesn't exist, then exit.
    #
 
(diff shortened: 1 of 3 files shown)


rvansa:
@vjuranek Could you check that this works well with Jenkins plugin? (and integrate afterwards)

jmarkos:
I'm running a job with these changes (but even if it finishes fine it won't prove that all the corner cases are handled).

jmarkos:
+1 from me

alanfx:
Thanks @jmarkos, and sorry for the last, less functional version...

rvansa:
Okay then, integrating. Thanks!

vjuranek:
Little bit late to the part, but anyway - work fine with Jenkins plugin.

PR #380, opened 2016-07-25
Remove duplicit properties

jmarkos:
If you define the num-threads property on the org.radargun.stages.cache.test.LoadStage, it is not set, because the PropertyHelper will find a property with the same name in the superclass org.radargun.stages.test.LoadStage (it's really unlucky that they are both called the same) and set the value in the superclass. So when I set num-threads=20, the superclass will have 20, and org.radargun.stages.cache.test.LoadStage will have 10, leading to weird behaviour.


rvansa:
Thanks, integrated.

PR #382, opened 2016-07-26
Add infinispan90 plugin

jmarkos:
Fixed the CQ support and also modified the example benchmark. I lowered the amount of requests, but it still sometimes fails with
ERROR [org.radargun.stages.test.RunningTest] (Stressor-5) Creating new thread during steady-state!
but that shouldn't be related to this PR.

### DIFF
--- a/extensions/query/src/main/resources/benchmark-continuous-query.xml
+++ b/extensions/query/src/main/resources/benchmark-continuous-query.xml
@@ -3,15 +3,20 @@
     <master bindAddress="${master.address:127.0.0.1}" port="${master.port:2103}"/>
 
     <clusters>
-        <scale from="1" to="2">
+        <scale from="2" to="2">
             <cluster/>
         </scale>
     </clusters>
 
     <configurations>
-        <config name="ContinuousQuery">
+        <config name="ContinuousQuery80">
             <setup plugin="infinispan80">
                 <default xmlns="urn:radargun:plugins:infinispan80:3.0" file="query.xml"/>
+            </setup>
+        </config>
+        <config name="ContinuousQuery90">
+            <setup plugin="infinispan90">
+                <default xmlns="urn:radargun:plugins:infinispan90:3.0" file="query.xml"/>
             </setup>
         </config>
     </configurations>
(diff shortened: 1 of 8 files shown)


alanfx:
Hey @jmarkos, I am getting the following compilation error with this code:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /Users/afield/Development/projects/radargun-master/plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90HotrodService.java:[14,27] createContinuousQuery() in org.radargun.service.Infinispan90HotrodService cannot override createContinuousQuery() in org.radargun.service.Infinispan80HotrodService
  return type org.radargun.traits.ContinuousQuery is not compatible with org.radargun.service.InfinispanHotrodContinuousQuery
```

I *think* that `Infinispan90HotrodContinuousQuery` needs to extend `InfinispanHotrodContinuousQuery`, but from looking at the classes, I can't tell what is different in Infinispan 9 versus 8. If you still need those new classes, then you should put your name as the author as well!

rvansa:
Wrt author: I think we should remove the author tags completely (in another PR), it's just a pre-git nuisance.

jmarkos:
@alanfx it compiles fine for me with 
mvn clean install -Dno-chm -Dno-ehcache -Dno-jbosscache -Dno-zip -DskipTests -Dlog4j2 -Pjdg
What command do you use for building?
The new class is needed because of some small changes in infinispan api.

alanfx:
@jmarkos I get the same error with your command. I am using `mvn clean install -DskipTests -Dmaven.test.skip.exec=true -U`. I'll look at it some more.

alanfx:
The problem is that `Infinispan90HotrodService` extends `Infinispan81HotrodService`, but `Infinispan90HotrodContinuousQuery` does not extend `InfinispanHotrodContinuousQuery` and the return type from `Infinispan90HotrodService.createContinuousQuery` is incompatible with `Infinispan81HotrodService.createContinuousQuery`. I can see the differences now in the `Infinispan90HotrodContinuousQuery.Listener` and `Infinispan90HotrodContinuousQuery.createContinuousQuery` code now. You can see my changes to make this compile here: https://github.com/alanfx/radargun/commit/15d448810ba87fb66277ed22129f8479f0a4c6e5

alanfx:
Integrated. Thanks, @jmarkos!

PR #384, opened 2016-07-27
Issue 373: Fix class path scanner subclass check

alanfx:
Use correct method to find subclasses and interface implementors

### DIFF
--- a/core/src/main/java/org/radargun/config/ClasspathScanner.java
+++ b/core/src/main/java/org/radargun/config/ClasspathScanner.java
@@ -1,18 +1,20 @@
 package org.radargun.config;
 
 import java.lang.annotation.Annotation;
+import java.util.Collections;
 import java.util.List;
 import java.util.function.Consumer;
 
 import io.github.lukehutch.fastclasspathscanner.FastClasspathScanner;
+import io.github.lukehutch.fastclasspathscanner.scanner.ScanResult;
 
 import org.radargun.logging.Log;
 import org.radargun.logging.LogFactory;
 
 /**
+ * 
  * Helper for listing classes on classpath.
  *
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/radargun/config/ClasspathScanner.java
```diff
@@ -25,29 +26,40 @@ private ClasspathScanner() {
...
-            if (superClass != null && !superClass.isAssignableFrom(clazz)) {
-               continue;
-            }
             consumer.accept((Class<? extends TClass>) clazz);
          } catch (Throwable e) {
             // static ctor can throw non-wrapped error
-            log.error("Cannot load class " + className, e);
-            continue;
+            log.trace("Cannot load class " + className, e);
+            return;
```
rvansa: Why do you want to quit the lookup once single class throws?

rvansa:
Hmm, I didn't realize that I could just compute the intersection of those two sets. Good, go for it.

core/src/main/java/org/radargun/config/ClasspathScanner.java
alanfx: I don't! Just keep me away from the code...

alanfx:
Hang on, I realized why the name of the superclass needs to be in the constructor

alanfx:
Well, I guess I didn't. I added the check for `assignableFrom.size() > 0`, because `getNamesOfSubclassesOf` returns 0 classes when the superclass is `java.lang.Object`. I thought that whitelisting that class in the constructor would change that, but it didn't help, so I didn't make any other changes.

alanfx:
Ok, this latest version skips checking for the superclass if `null` is passed to `ClasspathScanner.scanClasspath`

PR #385, opened 2016-07-28
Removed @author tags

rvansa:
- remove author tags

``` bash
    grep -R -n . -e '@author' | awk 'BEGIN { FS=":" }; { system("echo -e \""$2"d\\nw\" | ed " $1); }'
```
- remove empty javadocs

``` bash
    perl -0777 -pi -e 's/\/\*\*[\n\t *]*\*\/\n?//g' `find . -name "*.java"`
```
- remove trailing lines in comments

``` bash
    perl -0777 -pi -e 's/[\n\t *]*(\n *\*\/)/\1/g' `find . -name "*.java"`
```

### DIFF
--- a/core/src/test/java/org/radargun/config/EvaluatorTest.java
+++ b/core/src/test/java/org/radargun/config/EvaluatorTest.java
@@ -6,7 +6,6 @@ import static org.testng.Assert.assertEquals;
 
 /**
  * @author Mircea Markus &lt;Mircea.Markus@jboss.com&gt;
- * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  */
 @Test
 public class EvaluatorTest {
--- a/core/src/main/java/org/radargun/Directories.java
+++ b/core/src/main/java/org/radargun/Directories.java
@@ -4,9 +4,6 @@ import java.io.File;
 
 import org.radargun.utils.Utils;
 
-/**
- * @author Radim Vansa &lt;rvansa@redhat.com&gt;
- */
 public final class Directories {
    public static final File ROOT_DIR;
    public static final File LIB_DIR;
--- a/core/src/main/java/org/radargun/Operation.java
+++ b/core/src/main/java/org/radargun/Operation.java
@@ -5,8 +5,6 @@ import java.util.concurrent.ConcurrentHashMap;
 /**
  * Abstraction of any operation executed through traits. Each operation
  * has ID (this may differ on different nodes) and name (that should be unique).
- *
- * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  */
 public final class Operation {
    private static int nextId = 1;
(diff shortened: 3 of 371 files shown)


rvansa:
Btw., @alanfx we seem to have a battle in our IDE settings regarding newline after last `}` in file; have you set it somehow manually?

core/src/test/java/org/radargun/config/EvaluatorTest.java
```diff
@@ -5,9 +5,7 @@
 import static org.testng.Assert.assertEquals;
 
 /**
- * @author Mircea Markus &lt;Mircea.Markus@jboss.com&gt;
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
- */
```
jmarkos: This doesn't seem correct.

mgencur:
Why do you want to remove all the @author tags? Do you have any good reason? It is not necessary to remove them for existing code. 
We should have a consensus on this change before it is silently merged in! It should not be a decision of a single developer. Sorry.

rvansa:
@mgencur I don't want to merge this 'silently', this is a good place to discuss it. PRs can be rejected, too.

I think that these tags are often outdated, and the true authorship can be found out from git history. That makes them redundant.

core/src/test/java/org/radargun/config/EvaluatorTest.java
rvansa: Oops, it seems the PR really did not remove all of them...

mgencur:
See some good points for not removing the author tags:
https://webweavertech.com/ovidiu/weblog/archives/000319.html#000319
https://webweavertech.com/costin/archives/000320.html
The tags are not functionally required but they provide some hints about the authorship (without manually asking git to tell you) and evolution of the software as you go through the source code. They're form of communication.
@rvansa you're free to remove your own tags if you like. But I don't see how it will help anybody.

rvansa:
@mgencur Note that these comments come from 2004 (though they mention CSV, I believe the tools now in common use make finding out author a two-click task). And I understand the arguments as 'author does not lose his authorship due to community participation.

My intent is not to remove anyone's authorship but make sure the authorship is not mistaken by plainly lookin on the author tag.

Nevertheless, since you prefer to keep your tags, I won't remove all tags but only those who give the permission. @alanfx @mcimbora @jmarkos @vjuranek WDYT?

mgencur:
Thanks for not removing all the tags. I suppose there are still classes written by @mmarkus  and @maniksurtani .

jmarkos:
Personally I rarely find the tags useful, so I don't mind them being gone.

alanfx:
I agree. `git blame` seems much more useful than `@author`!

alanfx:
@rvansa I am using the Infinispan Eclipse IDE settings which do not add a newline at the end of the file. Are you using these same settings or different ones?

vjuranek:
feel free to remove my tags, it would make my life more easy - I'm always confused which name I should write there as 99% of my code is copy-paste from other files or internet :innocent:

rvansa:
@vjuranek [computer-programming-to-be-officially-renamed-googling-stackoverflow](http://www.theallium.com/engineering/computer-programming-to-be-officially-renamed-googling-stackoverflow/)

rvansa:
Closing, stale, outdated and rotten.

PR #393, opened 2016-08-18
Include git information in the distribution

jmarkos:
I sometimes run into a situation where I don't know how I built a particular RG distro (what version? what branch?). This adds a git.properties file in the distribution/bin folder, containing such information.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -736,6 +736,26 @@
                </dependency>
             </dependencies>
          </plugin>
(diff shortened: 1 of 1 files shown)


rmacor:
+1 This would indeed help a lot.

rvansa:
I would put it into root directory rather than `bin`, but it's good enough.

PR #396, opened 2016-09-02
Add jdg661 plugin

jmarkos:


alanfx:
I don't think we want to add a plugin for a micro release. I thought new plugins were only added for minor releases? (infinispan50, infinispan51, infinispan52, and infinispan53, but no infinispan501 for example.)

mgencur:
@alanfx , how would you do a comparison between 6.6.1 and 6.6 if there was no 6.6.1 plugin? Would you need to separate builds of RadarGun and two separate runs? Would you be able to put it in a single charts? Thanks

mcimbora:
@mgencur You can always make a copy of existing plugin and set different versions at build time. However this is indeed a bit user-unfriendly process.

alanfx:
@mgencur yes, you would have to do something similar to what @mcimbora suggests which is what @jmarkos has been doing when comparing different branches of the same product. I am just trying to avoid an explosion of plugins that slow down the build. The current set of plugins are synched to the latest micro release. (infinispan80->8.0.2, infinispan81->8.1.3, infinispan82->8.2.4, jdg63->6.3.2, jdg64->6.4.1, etc) I don't expect that we will need to compare 6.6.0 to 6.6.1 too many times as well. Just my opinion, though.

mgencur:
OK. That sounds good.

rvansa:
+1 to avoid micros, it's really no different from comparing branches/commits. If we want to ease the process, we could allow separate builds of plugins, and renaming the destination - something like
```
mvn install -pl :plugin-jdg66 -Djdg66.infinispan.version=6.6.1.Final-redhat-1 -Djdg66.plugin.name=jdg661
```
That could create a jdg661 directory in `target/distribution/Radargun-XXX/plugins`. However as currently the packing process is in the aggregator pom file, it would need non-trivial refactoring of the ant scripts.

alanfx:
+1 to @rvansa's idea for modifying the build process to build a plugin based on different versions, and -1 to the adding a plugin for a micro version. Closing for now

PR #401, opened 2016-09-05
Fail the stress test if cluster size changed during the test

jmarkos:


extensions/cache/src/main/java/org/radargun/stages/cache/test/legacy/CacheTestStage.java
```diff
@@ -34,11 +35,38 @@
...
+   protected int clusterSizeAtTestStart = 0;
+   protected int membershipHistorySizeAtStart = 0;
+
+   @Override
+   protected void destroy() {
+      if (failIfClusterSizeChanges && clustered != null) {
+         for (int i = membershipHistorySizeAtStart; i < clustered.getMembershipHistory().size(); i++) {
+            int clusterSizeDuringTest = clustered.getMembershipHistory().get(i).members.size();
+            if (clusterSizeDuringTest != clusterSizeAtTestStart) {
+               throw new RuntimeException("Cluster size at the beginning of the test was " + clusterSizeAtTestStart + " but changed to " + clusterSizeDuringTest +
```
jmarkos: Now it occured ot me that maybe it would be enough to check if the size of getMembershipHistory() changed.

jmarkos:
This won't work for client/server tests because the test stage runs only on clients...

jmarkos:
@alanfx actually I'm not sure if changes are needed - I don't know if there can be membership events where there are no actual changes in the membership, so I'll leave it as is just to be safe.
I don't have a solution for client/server tests, but if I find something nice I can implement it later.

rvansa:
This is pretty ortogonal to purpose of stage itself. Could you rather make this a separate stage (in benchmark configuration, you'd add one stage before test stage, that would record whatever it needs into slaveState and then after the test stage add a check?

jmarkos:
Thanks for the tip!
Updated (works also for client/server tests now).

alanfx:
@jmarkos I do like this implementation better, but I am not sure it achieves the desired goal. If a split happens during the test and then heals, this stage will not detect that. Is it possible to do something similar to the JMX monitoring classes that run in the background during the test stages, instead of just running before and after the test stages? Thanks, Alan

alanfx:
@jmarkos explained that the membership history will show any changes in cluster size that happened during the test. Integrating and thanks!

PR #402, opened 2016-09-06
NetworkBytesMonitor now shows traffic per second instead of total bytes

mgencur:
Example chart can be seen here: https://jdg-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/PERF-CS/job/mgencur-test-job/7/artifact/results/html/timeline_JDG%206.6.1%20-%20HotRod_0.html

There's a warm up at the beginning (2 mins) with writes and reads and then there are 3x1min iterations with just put operations on client-lon.


mgencur:
Fixed this PR. RadarGun didn't want to generate csv report because the name of the statistics contained parentheses. Now the name is changed and it works.

jmarkos:
Merged, thanks!

PR #398, opened 2016-09-02
Html report improvements

zhostasa:
Verticalized HTML Report tables
Added tooltip to clarify highlights
"Gross throughput" changed to "Troughput w/ errors" and hidden in reports without errors (NG in Chrome)

PR #404, opened 2016-09-08
Issue 394

jmarkos:
Depends on https://github.com/radargun/radargun/pull/398 .
Fixes the 2nd exception mentioned in the issue.
The problem was that this commit https://github.com/radargun/radargun/commit/e7d330a95e5e6a92eca16f79b5c080d2c257e17b removed the null check on statistics in ReportDocument - it is needed for tests where some nodes don't produce any statistics, like client/server.


alanfx:
Integrated. Thanks, @jmarkos!

Issue #405, opened 2016-09-08
RG build fails when git repo is not present


vjuranek:
bebd3b08689f70ead59773777c76ab152e8afb22 adds information from git, but it fails the build when git repo is not present:
```
[ERROR] Failed to execute goal pl.project13.maven:git-commit-id-plugin:2.2.1:revision (get-the-git-infos) on project radargun-aggregator: .git directory is not found! Please specify a valid [dotGitDirectory] in your pom.xml -> [Help 1]
```
The build should deal with such situation, ideally should print a warning and continue with build.

For completeness: it fails [my RG docker images](https://hub.docker.com/r/vjuranek/radargun-snapshot/builds/) - docker images should as small as possible, so I'm trying to avoid downloading git repo.

jmarkos:
Hmm there is <failOnUnableToExtractRepoInfo>false</failOnUnableToExtractRepoInfo>, I'd expect it to deal with this situation. I'll look into it.

PR #406, opened 2016-09-08
Issue #405 Do not fail the build if .git is not present

jmarkos:
Fixes https://github.com/radargun/radargun/issues/405


vjuranek:
Merged, thanks!

PR #383, opened 2016-07-26
Aggregations

jmarkos:
Includes also the commit to add infinispan90 plugin from https://github.com/radargun/radargun/pull/382.

Rebased (was https://github.com/radargun/radargun/pull/283).

PR #409, opened 2016-09-20
Hazelcast client-server plugin

rmacor:


parent/main/resources/checkstyle.xml
```diff
@@ -36,13 +36,6 @@
       <!-- Checks for imports                              -->
       <!-- See http://checkstyle.sf.net/config_import.html -->
       <module name="IllegalImport"/> <!-- defaults to sun.* packages -->
-      <module name="ImportOrder">
-         <property name="groups" value="java,javax,*"/>
-         <property name="ordered" value="true"/>
-         <property name="separated" value="false"/>
-         <property name="option" value="bottom"/>
-         <property name="sortStaticImportsAlphabetically" value="true"/>
-      </module>
```
jmarkos: you probably didn't want to commit this

plugins/hazelcast2/src/main/java/org/radargun/service/HazelcastService.java
```diff
@@ -40,7 +40,7 @@
    protected List<Membership> membershipHistory = new ArrayList<>();
 
    @Property(name = Service.FILE, doc = "Configuration file.")
-   private String config;
+   protected String config;
```
jmarkos: is this necessary?

plugins/hazelcast2/src/main/java/org/radargun/service/HazelcastService.java
```diff
@@ -132,7 +132,7 @@ protected synchronized void updateMembers(Set<com.hazelcast.core.Member> members
       membershipHistory.add(Membership.create(mbrs));
    }
 
-   private InputStream getAsInputStreamFromClassLoader(String filename) {
+   protected InputStream getAsInputStreamFromClassLoader(String filename) {
```
jmarkos: same as above, is it necessary?

plugins/hazelcast37/src/main/java/org/radargun/service/Hazelcast37ClientService.java
```diff
@@ -0,0 +1,57 @@
...
+import org.radargun.traits.ProvidesTrait;
+
+/**
+ * Hazelcast client service
+ *
+ * @author Roman Macor &lt;rmacor@redhat.com&gt;
+ */
+
+@Service(doc = "Hazelcast client")
+public class Hazelcast37ClientService extends Hazelcast36Service {
```
jmarkos: the only problem I see with this approach is if the client doesn't support some of the 'library' features, but from what I could find it seems it can do everything

plugins/hazelcast37/src/main/java/org/radargun/service/Hazelcast37ClientService.java
```diff
@@ -0,0 +1,57 @@
...
+
+   @Property(doc = "List of server addresses the clients should connect to, separated by semicolons (;).")
+   protected String servers;
+
+   @Override
+   public void start() {
+      ClientConfig clientConfig = new ClientConfig();
+      clientConfig.getGroupConfig().setName("dev").setPassword("dev-pass");
```
rvansa: The credentials should be configurable.

plugins/hazelcast37/src/main/java/org/radargun/service/Hazelcast37ClientService.java
```diff
@@ -0,0 +1,57 @@
...
+ * Hazelcast client service
+ *
+ * @author Roman Macor &lt;rmacor@redhat.com&gt;
+ */
+
+@Service(doc = "Hazelcast client")
+public class Hazelcast37ClientService extends Hazelcast36Service {
+
+   @Property(doc = "List of server addresses the clients should connect to, separated by semicolons (;).")
+   protected String servers;
```
rvansa: Make the property `List<String>` (or `String[]`) and provide appropriate property converter. And what's single element, is that IP:port? In that case it should not be `List<String>` but `List<Address>`.

plugins/hazelcast37/src/main/java/org/radargun/service/Hazelcast37ClientService.java
rvansa: Do the clients use the same API as library mode, and are the capabilities same? If it's the same and the only difference is configuration, you don't need separate service, just add the config options to Hazelcast36Service itself.

parent/main/resources/checkstyle.xml
rmacor: It was driving me crazy :) do we need to check for import order? the rest of the checkstyle plugin is fine (The problem is that intelliJ Idea is ordering imports differently than the plugin, and I don't see the option in Idea settings to change it)

plugins/hazelcast2/src/main/java/org/radargun/service/HazelcastService.java
rmacor: I'll remove it, I needed it when I was using xml configuration for the client.

plugins/hazelcast2/src/main/java/org/radargun/service/HazelcastService.java
rmacor: Same as the above.

plugins/hazelcast37/src/main/java/org/radargun/service/Hazelcast37ClientService.java
rmacor: ok

plugins/hazelcast37/src/main/java/org/radargun/service/Hazelcast37ClientService.java
rmacor: It can be IP or IP:port, I will make it List<String> and add the property converter.

rmacor:
@rvansa @jmarkos Thanks, for the review, I will make the changes and update the PR.

plugins/hazelcast37/src/main/java/org/radargun/service/Hazelcast37ClientService.java
rmacor: Yes, the capabilities are the same, but the client and the server require different properties. Isn't it going to be confusing if we put them to the same service?

rvansa: If Hazelcast uses the same API to configure either embedded mode or client, we should follow that approach.

rmacor: Well, they don't use the same API to configure embedded mode and client. Configurations API and XML configuration files are different. You can't use the same XML file to configure both client and embedded instance. But once you have Hazelcast Instance or Hazelcast client instance you can use the same API to perform operations on it.

Moving it to Hazelcast36Service would mean there would be properties which would only apply to embedded mode and properties which would only apply to client in the same service. (They would be differentiated by documentation) If that's not an issue I can move on with this.

alanfx: I guess I don't see why Hazelcast shouldn't have multiple services, similar to the Infinispan plugin. Especially since I suspect different groups will use the different services in the benchmarks as well. Am I missing something?

mgencur: Yes, if the configuration is different for both modes let's have two versions, one for library and one for client-server.

rvansa: Okay, I probably misunderstood - looking into code, you start with `HazelcastClient.getInstance(...)` and use different configuration class.

But make sure that the client and embedded may independently evolve - Client service shouldn't extend Embedded service, then.

alanfx: 👍 to what @rvansa said! The two services should be separated.

parent/main/resources/checkstyle.xml
rmacor: @rvansa @jmarkos so what do you think? do we want to keep check style for input order?

mcimbora: @rmacor @jmarkos Right, the project classes already adhere to these import rules and I see no good reason to change it.

rmacor: alright

rvansa: @rmacor The ordering can be changed in Editor/Code style/Java tab Imports. To not mess with imports, go to Editor/General/Auto Import and uncheck "Optimize imports on the fly".

rmacor:
PR updated.

mgencur:
Roman, maybe I have missed that but can you confirm that in client-server mode the Hazelcast client and server run in separate JVMs? I can see in your PR that you re-use the old Hazelcast plugin that was running in library mode (same JVM) and only implemented the new client. Is there a separate process running for the Hazelcast server and clients? Thanks

rmacor:
Yes, I can confirm that each service runs in separate JVM.

jmarkos:
This needs to wait for https://github.com/radargun/radargun/pull/383, because I have a feeling there will be conflicts and I don't want to rebase mine anymore :neckbeard:

jmarkos:
The example benchmark needs https://github.com/radargun/radargun/pull/398 and https://github.com/radargun/radargun/pull/404 to run correctly, so let's wait for those.

jmarkos:
Merged, thanks!

PR #412, opened 2016-09-22
Fix LegacyStressor statistics recording at the end of the test

jmarkos:
RequestSet can throw this exception
```07:32:25,296 ERROR [org.radargun.stages.test.legacy.LegacyStressor] (Stressor-19) Unexpected error in stressor!
    java.lang.IllegalArgumentException
    at org.radargun.stats.RequestSet.add(RequestSet.java:21) ~[radargun-core-3.0.0-SNAPSHOT.jar:?]```
during the end of the test, when a Request is discarded (because the test already ended) and thus not finished, but still passed to RequestSet which excepts only finished Requests.
I don't really like the fix, but a nicer one would probably require some refactoring in LegacyStressor, which I'm not eager to do.

### DIFF
--- a/core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
+++ b/core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
@@ -213,7 +213,7 @@ public class LegacyStressor extends Thread {
             log.error("Failed to end transaction", e);
          }
       } finally {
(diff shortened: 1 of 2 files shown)


jmarkos:
Seems like this was already done in https://github.com/radargun/radargun/pull/389 
There goes a wasted afternoon :(

PR #413, opened 2016-09-25
JGroups 4 changes

rmacor:
Changes needed to run with JGroups4 - https://github.com/infinispan/infinispan/pull/4529

### DIFF
--- a/plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90Lifecycle.java
+++ b/plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90Lifecycle.java
@@ -1,0 +1,24 @@
+package org.radargun.service;
+
+import org.infinispan.remoting.transport.jgroups.JGroupsTransport;
+import org.jgroups.JChannel;
+
+/**
+ * @author Roman Macor (rmacor@redhat.com)
+ */
+public class Infinispan90Lifecycle extends Infinispan52Lifecycle {
+
+   public Infinispan90Lifecycle(Infinispan90EmbeddedService service) {
+      super(service);
+   }
+
+   @Override
+   protected JGroupsTransport getTransport() {
+      return (JGroupsTransport) (service.cacheManager).getTransport();
+   }
+
+   @Override
+   protected JChannel getTransportChannels() {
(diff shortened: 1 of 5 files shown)


plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90Lifecycle.java
```diff
@@ -0,0 +1,24 @@
...
+ * @author Roman Macor (rmacor@redhat.com)
+ */
+public class Infinispan90Lifecycle extends Infinispan52Lifecycle {
+
+   public Infinispan90Lifecycle(Infinispan90EmbeddedService service) {
+      super(service);
+   }
+
+   @Override
+   protected JGroupsTransport getTransport() {
```
rvansa: The code is identical to 5.2; I recall needing to do similar things, though - what was your error? Adding a comment explaining why this is necessary could clarify... And no need for parentheses here, btw.

rmacor: It was NoSuchMethodError, you can see the full stack trace here: https://jdg-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/PERF-LIB/job/perf-lib-perf-ack/12/console-edg-perf01/

rmacor: @rvansa any other issues, or can this be integrated?

alanfx:
@rmacor I just have a question on this. Does it affect any of the other ISPN/JDG plugins, or will they continue to work with JGroups 3.x?

plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90Lifecycle.java
rvansa: Nope. @alanfx please integrate when @rmacor confirms that it will work with JGroups 3.6 which are still in master.

rmacor:
@alanfx this doesn't affect other plugins, I can confirm, that all plugins (including infinispan90) will continue to work with JGroups 3.x

Issue #387, opened 2016-07-29
CPU and Memory usage charts do not show numbers for RG master process


mgencur:
Even though it can be selected on the web page, the charts for RG master don't show anything.

jmarkos:
Fixed by https://github.com/radargun/radargun/pull/416.

PR #416, opened 2016-09-30
gh-387 CPU and Memory usage charts for Master node

mgencur:


jmarkos:
Isn't there some way to deduplicate the code? MasterMonitors and SlaveMonitors looks almost the same.

mgencur:
There may be a way but that would include abstract classes and more complicated hierarchy. The current solution looked reasonable to me. Can you suggest something better?

mgencur:
Actually, it was not too complicated. Here it is:)

jmarkos:
Great, thanks! Integrated.

PR #420, opened 2016-10-11
JGroups dumper for JGroups 4

rmacor:
JGroupsDumper modified for JGroups 4. Please note, that this won't work with JGroups 3.x so don't integrate this before https://github.com/infinispan/infinispan/pull/4529 is integrated.

It would be great if someone could review this before that, though.
Thanks!

### DIFF
--- a/plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
+++ b/plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
@@ -1,0 +1,58 @@
+package org.radargun.service;
+
+import java.lang.reflect.Field;
+import java.util.Map;
+import java.util.concurrent.ThreadPoolExecutor;
+import org.jgroups.protocols.TP;
+import org.jgroups.stack.Protocol;
+import org.jgroups.stack.ProtocolStack;
+import org.jgroups.util.TimeScheduler;
+import org.radargun.utils.Utils;
+
+/**
+ * JGroupsDumper modified for JGroups 4
+ *
+ * @author Roman Macor (rmacor@redhat.com)
+ */
+public class JGroups4Dumper extends JGroupsDumper {
+
+   public JGroups4Dumper(ProtocolStack protocols, long interval) {
+      super(protocols, interval);
+   }
+
+   @Override
+   protected void addDumpers(){
+      dumpers.add(new UNICAST3Dumper());
+      dumpers.add(new JGroups4Dumper.TPDumper());
+   }
+
+   @Override
+   protected Map<String, Object> getStats(Protocol protocol){
+      return stack.dumpStats(protocol.getName(), null);
+   }
+
+   protected static class TPDumper extends JGroupsDumper.TPDumper {
+      @Override
+      public void dump(Protocol protocol) {
+         TP tp = (TP) protocol;
+         boolean dump = false;
+         dump = checkThreadPool((ThreadPoolExecutor) tp.getThreadPool(), "JGroups") || dump;
+         try {
+            Field f = TP.class.getDeclaredField("internal_pool");
+            f.setAccessible(true);
+            dump = checkThreadPool((ThreadPoolExecutor) f.get(tp), "Internal") || dump;
+         } catch (Exception e) {
+            log.error("Failed to get internal pool");
+         }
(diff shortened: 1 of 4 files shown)


plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * JGroupsDumper modified for JGroups 4
+ *
+ * @author Roman Macor (rmacor@redhat.com)
+ */
+public class JGroups4Dumper extends JGroupsDumper {
+
+   static {
+      // first we need to clear dumpers from JGroupsDumper static block
```
rvansa: There's little need to have `dumpers` as static variable.

jmarkos:
@rmacor FYI https://github.com/infinispan/infinispan/pull/4529 was merged

rmacor:
@jmarkos @rvansa thanks for the review, I've updated the PR and removed the static blocks.

plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
```diff
@@ -0,0 +1,58 @@
...
+   }
+
+   @Override
+   protected void addDumpers(){
+      dumpers.add(new UNICAST3Dumper());
+      dumpers.add(new JGroups4Dumper.TPDumper());
+   }
+
+   @Override
+   protected Map<String, Object> getStats(Protocol protocol){
```
rvansa: why does this need an override?

plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
```diff
@@ -0,0 +1,58 @@
...
+      return stack.dumpStats();
+   }
+
+   protected static class TPDumper extends JGroupsDumper.TPDumper {
+      @Override
+      public void dump(Protocol protocol) {
+         TP tp = (TP) protocol;
+         boolean dump = false;
+         dump = checkThreadPool((ThreadPoolExecutor) tp.getThreadPool(), "Regular") || dump;
```
rvansa: I would call that just "JGroups" or "Common", since it handles both regular and OOB messages

plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
rmacor: that's because Protocol.dumpStats() is not available in JGroups 4x

rvansa: Ah, I haven't noticed that here it works on `stack`. However, this means that the stats will be dumped for each protocol - too many times, doesn't it?

plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
rmacor: I'm not sure what are you referring to, what do you want me to rename? Anyway names JGroups and Common seem to general to me.

plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
rmacor: you are right, good point, how about this:
return stack.dumpStats(protocol.getName(), null);

rvansa: @rmacor IIUC that returns single-entry map in form of Map<String, Map<String, Object>>. Just test that before updating PR.

plugins/infinispan90/src/main/java/org/radargun/service/JGroups4Dumper.java
rvansa: There are not "Regular" and "OOB" pools in JGroups 4, just one unified.

rmacor:
@rvansa I've updated the PR according to your comments (and tested that it works).

rmacor:
@rvansa any other comments?

rvansa:
I am good, merging... Thanks

PR #426, opened 2016-10-21
Issue 425 - Check that group is configured when stage doesn't execute

alanfx:
Check that the group(s) configured on the stage have been configured in
the benchmark. If not, then fail the stage with an error.

### DIFF
--- a/core/src/main/java/org/radargun/config/Cluster.java
+++ b/core/src/main/java/org/radargun/config/Cluster.java
@@ -77,6 +77,17 @@ public class Cluster implements Serializable, Comparable<Cluster> {
       }
       throw new IllegalStateException("Slave index is " + slaveIndex + ", cluster is " + toString());
    }
+   
+   public boolean groupExists(String groupName) {
+      boolean result = false;
+      for (Group group : groups) {
+         if (group.name.equals(groupName)) {
+            result = true;
+            break;
+         }
+      }
+      return result;
+   }
 
    public int getClusterIndex() {
       return index;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/radargun/SlaveBase.java
```diff
@@ -73,6 +74,17 @@ protected void scenarioLoop() throws IOException {
             } else if (!stage.shouldExecute()) {
                log.info("Stage '" + stage.getName() + "' should not be executed");
                response = new DistStageAck(state);
+               if (stage instanceof AbstractDistStage) {
```
rvansa: I think this check should rather be a part of `AbstractDistStage.getExecutingSlaves`, WDYT?

alanfx: The issue is that `AbstractDistStage.shouldExecute()` is called before `AbstractDistStage.getExecutingSlaves`. In the case that the stage specifies a non-existing group, `AbstractDistStage.shouldExecute()` returns `false`, so `AbstractDistStage.getExecutingSlaves` isn't executed and neither the benchmark or stage fails.

alanfx: I guess I could call Cluster.groupExists() after the AbstractDistStage.shouldExecute() call. I'll look at that tomorrow!

rvansa: You're right of course; I thought that `getExecutingSlaves` is called from `shouldExecute`. Yes, checking that in shouldExecute is the solution I'd prefer.

alanfx:
Move the group existence check to `AbstractDistStage.shouldExecute()`. If the group doesn't exist, the benchmark fails.

rvansa:
Thanks, Alan!

PR #427, opened 2016-10-25
Chart naming change

zhostasa:
* Renamed a net troughput charts to "just" troughput charts
* Removed gross troughput chart
* Fixed a minor bug (array index)

### DIFF
--- a/reporters/reporter-default/src/main/java/org/radargun/reporting/html/TimelineChart.java
+++ b/reporters/reporter-default/src/main/java/org/radargun/reporting/html/TimelineChart.java
@@ -97,7 +97,7 @@ public class TimelineChart {
       for (Timeline.Event event : events) {
          if (event instanceof Timeline.Value) {
             Timeline.Value value = (Timeline.Value) event;
-            int bucket = (int) ((event.timestamp - startTimestamp) * MAX_EVENT_VALUES / (endTimestamp - startTimestamp));
+            int bucket = (int) ((event.timestamp - startTimestamp) * (MAX_EVENT_VALUES-1) / (endTimestamp - startTimestamp));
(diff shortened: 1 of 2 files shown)


reporters/reporter-default/src/main/java/org/radargun/reporting/html/TimelineChart.java
```diff
@@ -97,7 +97,7 @@ public void setEvents(List<? extends Timeline.Event> events, int slaveIndex, lon
       for (Timeline.Event event : events) {
          if (event instanceof Timeline.Value) {
             Timeline.Value value = (Timeline.Value) event;
-            int bucket = (int) ((event.timestamp - startTimestamp) * MAX_EVENT_VALUES / (endTimestamp - startTimestamp));
+            int bucket = (int) ((event.timestamp - startTimestamp) * (MAX_EVENT_VALUES-1) / (endTimestamp - startTimestamp));
```
zhostasa: Not exactly, I was not looking into this too deeply, but from what I did look at, when you have a very specific timestamp combination the result, which is used as index to store something to, will equal  MAX_EVENT_VALUES, which is also used for a size of array the index is used on, resulting in ArrayIndexOutOfBounds exception. It happens only occasionally, I believe it was an oversight.

rvansa: Yep, the fix is fine - `endTimestamp` is a max of all event timestamps so it's possible that `event.timestamp == endTimestamp` and it overflows due to `bucket == MAX_EVENT_VALUES`

mgencur:
Pushed. Thanks, @zhostasa !

PR #428, opened 2016-10-25
Add operation count based functionality to benchmarks

alanfx:
This PR adds the ability to execute a specific number of operations
during a benchmark.

### DIFF
--- a/core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
+++ b/core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
@@ -1,6 +1,7 @@
 package org.radargun.stages.test.legacy;
 
 import java.util.Random;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ThreadLocalRandom;
 
 import org.radargun.Operation;
@@ -39,8 +40,10 @@ public class LegacyStressor extends Thread {
    private Transactional.Transaction ongoingTx;
    private Statistics stats;
    private boolean started = false;
+   private CountDownLatch threadCountDown;
 
-   public LegacyStressor(LegacyTestStage stage, OperationLogic logic, int globalThreadIndex, int threadIndex, boolean logTransactionExceptions) {
+   public LegacyStressor(LegacyTestStage stage, OperationLogic logic, int globalThreadIndex, int threadIndex,
+         boolean logTransactionExceptions, CountDownLatch threadCountDown) {
       super("Stressor-" + threadIndex);
       this.stage = stage;
       this.threadIndex = threadIndex;
@@ -50,10 +53,11 @@ public class LegacyStressor extends Thread {
       this.completion = stage.getCompletion();
       this.operationSelector = stage.getOperationSelector();
       this.logTransactionExceptions = logTransactionExceptions;
+      this.threadCountDown = threadCountDown;
    }
 
    private boolean recording() {
-      return started && !stage.isFinished();
+      return this.started && !stage.isFinished();
    }
 
    @Override
@@ -76,6 +80,7 @@ public class LegacyStressor extends Thread {
    }
 
    private void runInternal() {
+      boolean counted = false;
       try {
          // the operation selector needs to be started before any #next() call
          operationSelector.start();
@@ -87,18 +92,23 @@ public class LegacyStressor extends Thread {
                if (ongoingTx != null) {
                   endTransactionAndRegisterStats(null);
                }
-               stats.begin();
-               this.started = started;
                break;
-            }
-            Operation operation = operationSelector.next(random);
-            try {
-               logic.run(operation);
-            } catch (OperationLogic.RequestException e) {
-               // the exception was already logged in makeRequest
+            } else {
+               if (!counted) {
+                  threadCountDown.countDown();
+                  counted = true;
+               }
+               Operation operation = operationSelector.next(random);
+               try {
+                  logic.run(operation);
+               } catch (OperationLogic.RequestException e) {
+                  // the exception was already logged in makeRequest
+               }
             }
          }
 
+         stats.begin();
+         this.started = true;
          completion.start();
          int i = 0;
          while (!stage.isTerminated()) {
--- a/core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
+++ b/core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
@@ -151,7 +151,10 @@ public abstract class LegacyTestStage extends BaseTestStage {
       completion.setCompletionHandler(new Runnable() {
          @Override
          public void run() {
-            finished = true;
+            //Stop collecting statistics for duration-based tests
+            if (duration > 0) {
+               finished = true;
+            }
             finishCountDown.countDown();
          }
       });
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
```diff
@@ -169,7 +169,18 @@ public void run() {
             throw new IllegalStateException("Interrupted during ramp-up.", e);
          }
       }
-      started = true;
+      
+      //Wait until all threads are started to start stage
+      while (!isStarted()) {
+         int stressorStartedCount = 0;
+         for (LegacyStressor stressor : stressors) {
+            if (stressor.isStarted()) {
```
rvansa: Ouch, busy wait? `LegacyStressor.started` isn't volatile. Why don't you just use `CountDownLatch`?

Btw. the field is not volatile for a purpose, because it's repeatedly read in `recoding()` and given that it was set only once, it's better to keep the result in local field; otherwise we could use stage.isStarted() all the time.

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
```diff
@@ -87,15 +91,18 @@ private void runInternal() {
...
-               stats.begin();
-               this.started = started;
                break;
-            }
-            Operation operation = operationSelector.next(random);
-            try {
-               logic.run(operation);
-            } catch (OperationLogic.RequestException e) {
-               // the exception was already logged in makeRequest
+            } else {
```
rvansa: This seems like you do one operation, call `stats.begin()` (before the other threads have started) and then just busy loops polling stage.isStarted() without running any ops. I don't think this is what you want.

core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
alanfx: I can look into using `CountDownLatch` instead here.

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
alanfx: That is exactly what the code does to guarantee that every thread spawned from the stage is ready to run before starting to collect statistics. I thought this was what we discussed?

rvansa: I'll reformulate the problems:
1) Until all other threads are started (`stage.isStarted()` is true), the thread should execute operations in a loop (though these are not recorded). Your change executes only single operation before that.
2) `stats.begin()` should be called as late as possible, before starting to execute recorded operations. Your change calls that right after first operation, before the other threads start (`stage.isStarted()` returns true).

alanfx:
Updated `LegacyTestStage.startStressors` to use a `CountDownLatch` and moved the call to `stats.begin()` right before the call to `completion.start()` in `LegacyStressor.runInternal`, and I think it is a lot simpler and closer to the ideal. I also set the stage `started` to true after the `rampUp` block. Let me know if you think that's correct.

core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
```diff
@@ -201,6 +206,14 @@ public void run() {
 
    protected Completion createCompletion() {
       Completion completion = new TimeStressorCompletion(duration);
+      if (numOperations > 0) {
```
rvansa: Please put the creation of `TimeStressorCompletion` to else branch (you can omit the variable and return new instance right away).

core/src/main/java/org/radargun/stages/test/legacy/TimeStressorCompletion.java
```diff
@@ -12,6 +12,7 @@
  */
 public class TimeStressorCompletion extends AbstractCompletion {
 
+   protected static final String PROGRESS_STRING = "Number of operations executed by this thread: %d. Elapsed time: %s. Remaining: %s. Total: %s.";
```
rvansa: Why have you moved this out of `AbstractCompletion`?

core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
```diff
@@ -166,7 +169,9 @@ public void run() {
             throw new IllegalStateException("Interrupted during ramp-up.", e);
          }
       }
+
```
rvansa: It's strange that checkstyle does not complain about whitespaces at end of line...

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
```diff
@@ -50,10 +53,15 @@ public LegacyStressor(LegacyTestStage stage, OperationLogic logic, int globalThr
...
       this.logTransactionExceptions = logTransactionExceptions;
+      this.threadCountDown = threadCountDown;
    }
 
    private boolean recording() {
-      return started && !stage.isFinished();
+      return this.started && stage.started && !stage.isFinished();
+   }
+
+   public boolean isStarted() {
```
rvansa: This method is not needed anymore.

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
```diff
@@ -87,18 +95,22 @@ private void runInternal() {
...
-            }
-            Operation operation = operationSelector.next(random);
-            try {
-               logic.run(operation);
-            } catch (OperationLogic.RequestException e) {
-               // the exception was already logged in makeRequest
+            } else {
+               if (!isStarted()) {
+                  threadCountDown.countDown();
+                  this.started = true;
```
rvansa: Okay, this looks much better. The only thing is that I would set `this.started = true` only when we see that `stage.isStarted()` - put that next to `stats.begin()`. For the countdown check, you could use just a local var.
Then, in `recording()` you won't have to check the `stage.started` which is a volatile read.

core/src/main/java/org/radargun/stages/test/legacy/TimeStressorCompletion.java
alanfx: Because this `PROGRESS_STRING` reports the time elapsed and duration while the one in `CountStressorCompletion` reports executed and total operation count.

core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
alanfx: No clue here

alanfx:
Updated

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
alanfx: Hey @rvansa, I have made the changes, but I am still curious why it matters that the threads execute operations that won't be recorded while the stage is starting up. What is the purpose of that behavior?

core/src/main/java/org/radargun/stages/test/legacy/TimeStressorCompletion.java
rvansa: Ok, my bad.

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
rvansa: Because when one thread starts a request, there should be already some other overlapping request being executed - otherwise the request would be uncontended (and likely faster).

rvansa:
Thanks Alan, integrating.

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
alanfx: OK, that makes sense to me. Thanks for explaining this.

PR #429, opened 2016-10-31
Allow to specify file with parameters for PerfRepo reporter

jmarkos:


reporters/reporter-perfrepo/src/main/java/org/radargun/reporting/perfrepo/PerfrepoReporter.java
```diff
@@ -231,6 +235,17 @@ private void addBasicParameters(TestExecutionBuilder testExecutionBuilder, Repor
             testExecutionBuilder.parameter(buildParam.getKey(), buildParam.getValue());
          }
       }
+      if (buildParamsFile != null) {
+         Properties paramsFromFile = new Properties();
+         try {
+            paramsFromFile.load(new FileInputStream(buildParamsFile));
```
rvansa: Please use try-with-resources, `Properties.load` does not close the stream.

jmarkos:
Done. Yes I use it for passing git info (and possibly other build information) to perfrepo.

PR #432, opened 2016-11-01
Testsuite fix

vjuranek:
* runs tests sequentially as when run in parallel they clash
* PowerMock doesn't work with abstract classes - convert `AbstractConfigurationTest` to normal class, however without `@Test` annotation so that it's not run by TestNG. Nicer solution can be done once switched to jUnit, which doesn't require test class to inherit from `PowerMockTestCase`

### DIFF
--- a/core/src/test/java/org/radargun/config/AbstractConfigurationTest.java
+++ b/core/src/test/java/org/radargun/config/AbstractConfigurationTest.java
@@ -31,16 +31,17 @@ import static org.testng.Assert.*;
  *
  * @author Roman Macor &lt;rmacor@redhat.com&gt;
  */
-@Test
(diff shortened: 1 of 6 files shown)


rvansa:
@vjuranek what and why hangs when running them in parallel?

vjuranek:
@rvansa 
what: random tests in `extensions/cache`, I tired now and it was `BackgroundStressorsGeneralTest`, waiting for checking the operations (in `BackgroundOpsManager.waitUntilChecked()`)

why: unfortunately no idea:-(

jmarkos:
This makes the testsuite pass, so +1 from me. @rvansa unless you have further comments I'll integrate it.

jmarkos:
Merged, thanks.

PR #433, opened 2016-11-01
Change required for testing Tomcat with mod_cluster

mgencur:
The main changes include:
* Tomcat 8 plugin
* Docker plugin for running docker containers.
* Example config files for both Tomcat and Docker plugins and for testing stateful web applications


plugins/resteasy-http/src/main/java/org/radargun/http/service/RESTEasyService.java
```diff
@@ -41,6 +44,14 @@
    @Property(doc = "Semicolon-separated list of server addresses.", converter = RESTAddressListConverter.class)
    protected List<InetSocketAddress> servers;
 
+   @Property(doc = "Ratio between the number of connections to individual servers. " +
+                   "Servers from the 'servers' list are indexed from 0. When the client " +
+                   "is first created it will choose a server to communicate with according " +
+                   "to this load balancing setting. The client will keep communicating with " +
+                   "this single server until redirected.", converter = Fuzzy.IntegerConverter.class)
+   //By default clients only connect to server 0. Example for two servers with equal load: "50%:0,50%:1"
+   protected Fuzzy<Integer> serversLoadBalance = Fuzzy.uniform(0);
```
mgencur: @rvansa , I've found a good use case for your Fuzzy class:)

rvansa: I would make the default to uniformly talk to all servers, otherwise the user could be surprised that the other servers are not used at all.

plugins/resteasy-http/src/main/java/org/radargun/http/service/RESTEasyService.java
```diff
@@ -68,6 +78,8 @@ public Lifecycle getLifecycle() {
 
    @Override
    public synchronized void start() {
+      checkValidLoadBalancingSettings();
+      random = ThreadLocalRandom.current();
```
rvansa: No kidding, sharing thread local random across threads?

plugins/docker/src/main/java/org/radargun/service/DockerService.java
```diff
@@ -0,0 +1,251 @@
...
+         .withStdErr(true)
+         .withStdOut(true)
+         .withFollowStream(true)
+         .withTailAll()
+         .exec(loggingCallback);
+
+      try {
+         loggingCallback.awaitCompletion(3, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+         e.printStackTrace();
```
rvansa: use logging and interrupt the thread

plugins/docker/src/main/java/org/radargun/service/DockerService.java
```diff
@@ -0,0 +1,251 @@
...
+      }
+      log.infof("Container log:\n%s", loggingCallback.toString());
+   }
+
+   private static class LogCallback extends LogContainerResultCallback {
+      protected final StringBuilder log = new StringBuilder();
+
+      @Override
+      public void onNext(Frame frame) {
+         log.append(new String(frame.getPayload()));
```
rvansa: Shouldn't you rather log directly?

plugins/docker/src/main/java/org/radargun/service/DockerService.java
```diff
@@ -0,0 +1,251 @@
...
+      started = false;
+      log.infof("Stopped container with id %s", dockerContainerId);
+   }
+
+   @Override
+   public boolean isRunning() {
+      return started;
+   }
+
+   private static class EnvsConverter implements Converter<List<String>> {
```
rvansa: Just move the ProcessService.EnvsConverter somewhere to utils and reuse that...

plugins/docker/src/main/java/org/radargun/service/DockerService.java
```diff
@@ -0,0 +1,251 @@
...
+         StringBuilder sb = new StringBuilder();
+         for (ExposedPort port : ports) {
+            sb.append(port.getPort()).append("/").append(port.getProtocol().toString()).append('\n');
+         }
+         return sb.toString();
+      }
+
+      @Override
+      public String allowedPattern(Type type) {
+         return Converter.ANY_MULTI_LINE;
```
rvansa: use proper regexp to match the `port/[udp|tcp]` (btw. I would use round parentheses as `[]` usually means optional part)

plugins/tomcat8/src/main/java/org/radargun/service/TomcatServerService.java
```diff
@@ -0,0 +1,284 @@
...
+ * with "manager-script" role and corresponding credentials.
+ *
+ * @author Martin Gencur
+ */
+@Service(doc = TomcatServerService.SERVICE_DESCRIPTION)
+public class TomcatServerService extends JavaProcessService {
+
+   protected final Log log = LogFactory.getLog(getClass());
+   protected static final String SERVICE_DESCRIPTION = "Tomcat Server";
+   private static final String JAVA_HOME = "JAVA_HOME";
```
mgencur: OK. The existing InfinispanServerService has these constants and fields related to Java too. I'll move them up to JavaProcessService as well.

plugins/resteasy-http/src/main/java/org/radargun/http/service/RESTEasyService.java
mgencur: OK. Makes sense. I was too lazy to think how to implement it.

plugins/tomcat8/src/main/java/org/radargun/service/TomcatServerService.java
```diff
@@ -0,0 +1,284 @@
...
+   private void processResponse(final String command, final HttpURLConnection hconn) throws IOException {
+      final int httpResponseCode = hconn.getResponseCode();
+      if (httpResponseCode >= 400 && httpResponseCode < 500) {
+         throw new IllegalStateException(
+            "Unable to connect to Tomcat manager. "
+               + "The server command (" + command + ") failed with responseCode ("
+               + httpResponseCode + ") and responseMessage (" + hconn.getResponseMessage()+ ").\n\n"
+               + "Please make sure that you provided correct credentials for a user which is allowed to access Tomcat manager application.\n"
+               + "The user must have 'manager-script' role specified in tomcat-users.xml file.\n");
+      } else if (httpResponseCode >= 300) {
```
mgencur: We don't want to deal with redirections because there are many reasons why redirection happens and this code is not supposed to deal with that. Hence I consider this an error. And I believe Arqullian engineers had a similar opinion. This code is taken from there and slightly modified.

plugins/tomcat8/src/main/java/org/radargun/service/TomcatServerService.java
```diff
@@ -0,0 +1,284 @@
...
+         throw new IllegalStateException("The server command (" + command + ") failed with responseCode ("
+            + httpResponseCode + ") and responseMessage (" + hconn.getResponseMessage() + ").");
+      }
+      BufferedReader reader = null;
+      try {
```
mgencur: Good point.

plugins/resteasy-http/src/main/java/org/radargun/http/service/RESTEasyService.java
rvansa: Use @Init annotation on a method, that will be called after properties are filled in, so then you can initialize this property (if it is still null)

extensions/rest/src/main/java/org/radargun/traits/RESTOperations.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+@Trait(doc = "Http operations.")
+public interface RESTOperations {
+   String TRAIT = RESTOperations.class.getSimpleName();
+
+   Operation GET = Operation.register(TRAIT + ".Get");
+
+   RESTOperationInvoker getRESTInvoker();
+
+   interface RESTOperationInvoker {
```
mgencur: I will move the get operation directly to RESTOperations, the invoker is probably not needed. Regarding the trait and URL, currently there's only one implementation of RESTOperations and that is tightly couple with RESTEasyService which defines the URL to be used. I'd have to immediately ignore this new String parameter in the single implementation because I want to define the URL directly in the configuration file. We can add this later if there's a need to request different URLs during a single stage.

plugins/docker/src/main/java/org/radargun/service/DockerService.java
mgencur: The problem is that dockerClient.withEnv expects a list of strings where each value is "key=value". I'd need to further convert the map to this list of strings hence I wrote my own converter.

mgencur: Let me check if streams can help with this.

extensions/rest/src/main/java/org/radargun/traits/RESTOperations.java
rvansa: We do not want to change interfaces all the time, though there is (and probably will be) only one implementation, there will be multiple usages.
You wouldn't ignore that parameter - you would simply append that to the base URL. You should add the parameter to the test stage and keep it empty if you don't want to use it.

plugins/docker/src/main/java/org/radargun/service/DockerService.java
rvansa: Yep. `map.entrySet().stream().map(entry -> entry.getKey() + "=" + entry.getValue()).collect(Collectors.toList())` should do the trick. But I don't insist on that.

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
```diff
@@ -94,8 +96,13 @@ private void runInternal() {
             Operation operation = operationSelector.next(random);
             try {
                logic.run(operation);
+               if (delayBetweenRequests > 0)
```
mgencur: b) is correct:) This is supposed to simulate the think time of the user/client that is sending requests to the server. In this case only latency makes sense, not throughput because the throughput is mostly given by the delay between requests. The name is "delayBetweenRequests" because there's already the same parameter in background.Stressor. So they're called same to prevent confusion.

mgencur:
@rvansa , pushed one commit that should address most of your concerns. The invoker is indeed needed. It holds the URL for each thread which remains constant afterwards.

mgencur:
OK. Pushed my latest changes. This should be ready.

core/src/main/java/org/radargun/stages/test/legacy/LegacyStressor.java
rvansa: I know the implementation does b); my concern is that you can't quantify the difference int latency between two impls/versions/configs when these two have different throughput.
In my view think time makes sense when client executes a sequence of operations, but in this test the requests don't present any sequence, therefore, I would expect that you are interested in latency when executing 1000 req/s, 2000 req/s etc. That's why we've used fixed schedule in the past; the throughput was close to number of threads * duration/period. It will suffer from coordinated omission, but you could at least compensate by setting the `requestStartTime` to the scheduled time (that wasn't done in the past since I wanted to fix everything right away, not realizing that it's impossible with sync API).

rvansa: Well, if you think that the results will be of any use to you, go for it. I just won't explain anyone why the results look the way they do :)

mgencur:
Fixed the conflicts.

extensions/rest/src/main/java/org/radargun/traits/RESTOperations.java
```diff
@@ -0,0 +1,23 @@
...
+@Trait(doc = "Http operations.")
+public interface RESTOperations {
+   String TRAIT = RESTOperations.class.getSimpleName();
+
+   Operation GET = Operation.register(TRAIT + ".Get");
+
+   RESTOperationInvoker getRESTInvoker(String contextPath);
+
+   interface RESTOperationInvoker {
+      WrappedHttpResponse get(List<Cookie> cookies, MultivaluedMap<String, Object> headers);
```
rvansa: We could push this principle even to the level of returning `javax.ws.rs.client.Client`, but that's rather too much for our purpose, since we want to set the servers in configuration, not in the stage. So about the proper level could be:
```
interface RESTOperationInvoker {
     Invocation.Builder invoke(String contextPath);
}
```

core/src/main/java/org/radargun/utils/EnvsConverter.java
```diff
@@ -0,0 +1,43 @@
...
+      StringBuilder sb = new StringBuilder();
+      for (Map.Entry<String, String> envVar : value.entrySet()) {
+         sb.append(envVar.getKey()).append('=').append(envVar.getValue()).append('\n');
+      }
+      return sb.toString();
+   }
+
+   @Override
+   public String allowedPattern(Type type) {
+      return Converter.ANY_MULTI_LINE;
```
mgencur: I was just moving the existing code but OK.This one should be close to what we want: 
"([A-Z_]+[A-Z0-9_]*=.*(\n|\r)*)+"

It could be even slightly better but it would be too complicated and unreadable

mgencur: There are two missing asterix in the example above. One after second closing square bracket and one after dot.

extensions/rest/src/main/java/org/radargun/RESTOperationInvocations.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Provides {@link Invocation} implementations for operations from traits
+ * {@link RESTOperationInvocations}
+ *
+ * @author Martin Gencur
+ */
+public class RESTOperationInvocations {
+   public static final class Get implements Invocation<WrappedHttpResponse> {
+      public static final Operation GET_NULL = RESTOperations.GET.derive("Null");
```
mgencur: OK. Null is probably not useful here.

extensions/rest/src/main/java/org/radargun/stages/RESTOperationsTestStage.java
```diff
@@ -0,0 +1,104 @@
...
+         validateSession(response);
+         isFirstRequest = false;
+      }
+
+      private void validateSession(WrappedHttpResponse<String> response) {
+         NewCookie newSessionId = response.getCookies().get(JSESSIONID);
+         if (newSessionId != null) {
+            jsessionid = newSessionId.toCookie();
+            if (!isFirstRequest) {
+               log.error("Session lost!");
```
mgencur: OK. Thanks for the tip. I was expecting to check the log but having this in the report is better.

plugins/docker/src/main/java/org/radargun/service/DockerService.java
```diff
@@ -0,0 +1,234 @@
...
+   @Property(doc = "The list of ports exposed by the container. This list should mimic ports exposed from a Docker image through EXPOSE. Empty by default.", converter = PortsConverter.class)
+   protected List<ExposedPort> exposedPorts = Collections.emptyList();
+
+   @ProvidesTrait
+   public Lifecycle getLifecycle() {
+      return this;
+   }
+
+   @Init
+   public void init() {
```
mgencur: OK. There's no indication that Init method should be used for this purpose. Why can't the logic be in the Init method? Is there any functional problem with this?

extensions/rest/src/main/java/org/radargun/traits/RESTOperations.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * @author Martin Gencur
+ */
+@Trait(doc = "Http operations.")
+public interface RESTOperations {
+   String TRAIT = RESTOperations.class.getSimpleName();
+
+   Operation GET = Operation.register(TRAIT + ".Get");
+
+   RESTOperationInvoker getRESTInvoker(String contextPath);
```
mgencur: I've already implemented the Fuzzy logic to evenly spread the load but somehow the changes got lost. Will add it in this PR.

mgencur: I think it is good enough to have the contextPath set at the beginning and then just re-used in the test. This is going to be used as a load test. It's not going to mimic Selenium which is able to go through several web pages and clicking on multiple buttons and requesting different URLs. Also, RG is not prepared for passing multiple different URLs and I'm not going to implement it now. We can add this option later when it's really needed. It will need a change in the configuration file anyway.

mgencur:
Pushed new fixes. This should fix everything except passing the contextPath differently.

PR #436, opened 2016-11-11
Fix versions of transaction manager in infinispan90 and jdg70 plugins

jmarkos:


mgencur:
Looks good. Pushed. Thanks, Jakub.

PR #437, opened 2016-11-16
Added CacheInformation Trait to resteasy-client.

andyuk1986:


mgencur:
@andyuk1986 , there are conflicts. Can you please rebase?

andyuk1986:
@mgencur thanks for the comment. Rebase is done!

plugins/resteasy-http/src/main/java/org/radargun/http/service/RESTEasyCacheInfo.java
```diff
@@ -0,0 +1,107 @@
...
+            try {
+               Invocation get = service.getHttpClient().target(target).request().accept(CONTENT_TYPE).buildGet();
+               response = get.invoke();
+
+               responseStr = response.readEntity(String.class);
+               size = Long.parseLong(responseStr);    //if the call was done to nodejs app, then the long size is returned
+
+            } catch (NumberFormatException ex) {
+               //If the call was done to Infinispan Server, then the response is the list of all entries
+               String[] entries = responseStr.split(System.getProperty("line.separator"));
```
mgencur: Hmm. It sounds like it could download the whole content of the cache. What if we have a few GB of data?

mgencur: Ah. Actually it returns just keys. Not whole entries. This might be fine if used rarely. What's the purpose of this method? When is it used? How often?

andyuk1986: Actually, this is how current ISPN Rest API work. It doesn't have a specific method for retrieving the size of the cache. The only available method which can be used for getting the number of elements is http://host:xxxx/rest/cacheName which returns all available elements in the cache.
At the moment there is no use for this block of code (I mean the block in the catch statement) as before these changes resteasy-client plugin didn't have this trait. 
I can add some comment here so that if later it is necessary to create a separate benchmark with resteasy-client plugin which would need to rely on RESTEasyCacheInfo trait, QE know about this risk. WDYT @mgencur ?

andyuk1986:
Ok, so as agreed - I have added a comment with warning about performing heavy operation.

mgencur:
Merged. Thanks, Anna.

rvansa:
@andyuk1986 And why do you need to know the number of entries when there's no API for that?
Could you file a JIRA to Infinispan to return the number of entries in upon HEAD request in X-Size header?

andyuk1986:
@rvansa  I needed that for verifying the xsite failover for nodejs client. I have written a separate rest client for nodejs, which returns number of entries in the cache. But in case if ISPN Rest API is used, then it gets a bit complicated, even though at the moment taking the size for REST API is not used in any radargun benchmark. Sure, I will open a jira with the request as you have suggested. Thanks.

PR #440, opened 2016-12-06
Registry related changes to docker plugin

andyuk1986:
Changed the docker service so that the name of pulled image also contains the registry name.
Implemented PullImageResultCallback for avoiding program failure in case of successfull image pulling.


mgencur:
Thanks, Anna!. Pushed with a minor fix for InterrupedException.

PR #441, opened 2016-12-09
add option to specify Kryo and Protostuff marshallers

rmacor:


mgencur:
Looking.

plugins/infinispan90/pom.xml
```diff
@@ -124,6 +124,17 @@
          <optional>true</optional>
       </dependency>
 
+      <dependency>
+         <groupId>org.infinispan</groupId>
+         <artifactId>infinispan-marshaller-kryo</artifactId>
+         <version>${version.infinispan}</version>
```
mgencur: My Maven build can't find the artifacts (Kryo and Protostuff) because I have 9.0.0-SNAPSHOT in version.infinispan. We should probably use an independent version for the Kryo and Protostuff libs. I can see that Nexus contains only 9.0.0.Beta1

plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90HotrodService.java
```diff
@@ -10,9 +12,23 @@
 @Service(doc = Infinispan60HotrodService.SERVICE_DESCRIPTION)
 public class Infinispan90HotrodService extends Infinispan81HotrodService {
 
+   @Property(doc = "Fully specified Marshaller class, if no class is defined, the default Marshaller - JBossMarshaller is used" +
```
mgencur: The two lines of the doc will actually be on a single line. There's a missing space or line separator.

plugins/infinispan90/pom.xml
rvansa: @mgencur until 9.0.0.Final is out, you should use `-Dinfinispan90.infinispan.version=9.0.0.Beta1` for build.

mgencur: Alright. The Kryo and Protostuff marshaller is part of the main repository. So that explains it.

mgencur:
Thanks, Roman. Integrated with a minor change for the docs message.

PR #444, opened 2016-12-14
Add REST stages that run in background

mgencur:
* useful for resilience tests where we need to kill some servers from stages running in foreground

Example benchmark looks like this: https://gist.github.com/mgencur/45dcd4af4d0a9dc5046676fab092c000

### DIFF
--- a/core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
+++ b/core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
@@ -70,13 +70,18 @@ public abstract class LegacyTestStage extends BaseTestStage {
    @InjectTrait
    protected Transactional transactional;
 
-   private CountDownLatch finishCountDown;
    private Completion completion;
    private OperationSelector operationSelector;
 
    protected volatile boolean started = false;
    protected volatile boolean finished = false;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/radargun/stages/test/legacy/LegacyTestStage.java
```diff
@@ -96,10 +101,11 @@ public DistStageAck executeOnSlave() {
       try {
          long startNanos = TimeService.nanoTime();
          log.info("Starting test " + testName);
-         List<LegacyStressor> stressors = execute();
+         stressorsManager = setUpAndStartStressors();
+         waitForStressorsToFinish(stressorsManager);
```
jmarkos: Couldn't these use the stressorsManager field directly? So you'd call just
setUpAndStartStressors();
waitForStressorsToFinish();

mgencur: These two methods are used separately in BackgroundRESTOperationsStartStage and BackgroundRESTOperationsStopStage . I cannot do what you suggest because in the "stop" stage I need to obtain the stressor manager from the "start" stage. So I'm passing it as a parameter.

jmarkos:
Merged, thanks.

PR #447, opened 2016-12-16
Use empty docker registry by default

mgencur:
Before this fix it was not possible to build the image locally and use it. It was always expected the image in the remote repository.


rvansa:
@mgencur The default is `null` - are you sure line 95 does not throw NPE?

mgencur:
Oops. Fixed.

jmarkos:
Rebased and merged, thanks.

Issue #449, opened 2017-01-12
Enable getting slave addresses via properties


mgencur:
Example:
When I have the following definition of clusters:
`<clusters>
      <cluster>
         <group name="servers" size="1"/>
         <group name="clients" size="1"/>
      </cluster>
   </clusters>
`

It should be possible to get slave's address via a property:
`<setup group="clients" plugin="resteasy-http">
     <rest-client-stateful xmlns="urn:radargun:plugins:resteasy-http:3.0">
           <servers>${group.servers.0}:8080</servers>
     </rest-client-stateful>
</setup>
`

rvansa:
Hmm, regrettably the slaves addresses are unknown to both master and slave nodes. Configuration is evaluated on all slaves separately (and not at all on master), so you would have to check the addresses when all slaves connect, and post this to slaves in `Master` - or maybe rather add that into `o.r.config.Cluster`.

Another problem is that you can have multiple ifaces, and not always you want to use the same iface as for Slave - Master connection. As you can't discover the other ifaces from master automatically, that requires sending that info from Slave (probably using `RemoteMasterConnection.connectToMaster`).

As a last remark, RG already has convention for accessing element of a list and that is `#{ ${foo}.get(0) }`. Syntax would then be `${ #{ ${groups.servers}.get(0) }.eth2 }` - though I admit that this is rather cumbersome.

PR #450, opened 2017-01-12
gh-449 Enable getting slave addresses via properties

mgencur:
Fixes https://github.com/radargun/radargun/issues/449

### DIFF
--- a/core/src/main/java/org/radargun/RemoteSlaveConnection.java
+++ b/core/src/main/java/org/radargun/RemoteSlaveConnection.java
@@ -24,6 +24,7 @@ import org.radargun.config.Scenario;
 import org.radargun.logging.Log;
 import org.radargun.logging.LogFactory;
 import org.radargun.reporting.Timeline;
+import org.radargun.utils.SlaveConnectionInfo;
 import org.radargun.utils.TimeService;
 
 /**
@@ -45,6 +46,7 @@ public class RemoteSlaveConnection {
 
    private ServerSocketChannel serverSocketChannel;
    private SlaveRecord[] slaves;
+   private SlaveAddresses slaveAddresses;
 
    private ByteBuffer mcastBuffer;
    private Map<SocketChannel, ByteBuffer> writeBufferMap = new HashMap<SocketChannel, ByteBuffer>();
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/radargun/Slave.java
```diff
@@ -46,6 +49,8 @@ private void run(int slaveIndex) throws Exception {
          if (object == null) {
             log.info("Master shutdown!");
             break;
+         } else if (object instanceof ArrayList) {
```
rvansa: Eh, this is rather generic object... what about adding the addresses into `Cluster` instead? Or at least add a new class that distinguishes the data.

core/src/main/java/org/radargun/RemoteSlaveConnection.java
```diff
@@ -132,6 +132,7 @@ private int connectSlaves(long timeout) throws IOException {
                throw new IllegalArgumentException("We are expecting 0th generation slave " + slaveIndex + " but it already has UUID set!");
             }
             slaves[slaveIndex] = new RemoteSlaveConnection.SlaveRecord(slaveIndex, uuid, socketChannel);
+            slaveAddressList.add((InetSocketAddress) socketChannel.getRemoteAddress());
```
rvansa: You can be putting the addresses in wrong order, you need to use `slaveIndex`.

core/src/main/java/org/radargun/Slave.java
```diff
@@ -121,6 +126,11 @@ protected void sendResponse(DistStageAck response) throws IOException {
       extras.put(Properties.PROPERTY_GROUP_NAME, group.name);
       extras.put(Properties.PROPERTY_GROUP_SIZE, String.valueOf(group.size));
       for (Cluster.Group g : cluster.getGroups()) {
+         for(int i = 0; i != g.size; i++){
+            extras.put(Properties.PROPERTY_GROUP_PREFIX + g.name + "." + i, state.getSlaveAddress(cluster, g.name, i).getAddress().toString().substring(1));//remove initial slash
```
rvansa: use `getHostAddress()` instead of `toString()`

core/src/main/java/org/radargun/RemoteSlaveConnection.java
```diff
@@ -161,6 +163,11 @@ public void sendConfiguration(Configuration configuration) throws IOException {
       flushBuffers(0);
    }
 
+   public void sendSlaveAddressList() throws IOException {
+      mcastObject(slaveAddressList, slaves.length);
```
rvansa: Why do you send a `InetSocketAddress` and not directly the `InetAddress`?

core/src/main/java/org/radargun/Slave.java
mgencur: I thought Cluster is there only to hold configuration and not real data about the slaves. That's why I didn't add it there. Something more specific than ArrayList would be good probably.

mgencur:
OK. I think all comments are addressed now. The user is now able to specify any of slave's interfaces, e.g. `${group.xxx.0.lo}` or `${groups.xxx.0.eth2}` 
Right now there's no option to omit the interface name so it's necessary to specify it.

rvansa:
I am not particularly happy about adding the addresses to SlaveState, but I don't see any better place atm, so let's keep it that way.

Not sure about changing `127.0.0.1` in example configs to `lo` - that could break on machines that are missing interface `lo`.

Quite minor: if the interface has multiple addresses bound, the value should be a list of addresses (simply concatenate with ","), not just the first one.

mgencur:
OK. Let me fix the last two points later today.

mgencur:
@rvansa , fixed.

rvansa:
Integrated, thanks.

PR #451, opened 2017-01-12
Provide option -Dpackage-latest for packaging only latest version of …

mgencur:
…all plugins

The question is: Do we want to make it default to package only latest versions of plugins?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,13 +33,20 @@
       <plugins.docker/>
       <plugins.jcache/>
       <plugins.ehcache/>
+      <plugins.ehcache-latest/>
       <plugins.hazelcast/>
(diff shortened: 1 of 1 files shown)


jmarkos:
I vote for it being the default, I don't remember the last time I ran something else than the latest version of any plugin.

rvansa:
@jmarkos If that's a default, how will a new user get to know that there's this option and he can switch it off?

If this is a default, please make sure that all the example configs that compare two versions are working on default build. As a rule of thumb, I made all plugins that have publicly available dependencies be compiled by default, to prevent such issues.

jmarkos:
You're right, let's not make it a default then.

jmarkos:
Merged, thanks.

PR #453, opened 2017-01-18
Code style - add import order

rmacor:


ide-settings/intellij/RadarGunCodeStyle.xml
```diff
@@ -14,7 +14,19 @@
     </value>
   </option>
   <option name="LINE_SEPARATOR" value="&#xA;" />
+  <option name="LAYOUT_STATIC_IMPORTS_SEPARATELY" value="false" />
   <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="10" />
+  <option name="IMPORT_LAYOUT_TABLE">
+    <value>
+      <package name="java.io" withSubpackages="true" static="false" />
```
rvansa: Why do you lay out `java.io` and `java.util` separately? moreover, the checkstyle says that order is `java, javax, *`.

rmacor:
@rvansa I've updated the order, but please note that the groups property in check style plugin doesn't work as you would expect. You can change the order of the groups and the plugin will still pass, it only complains if the order inside group is incorrect.

PR #454, opened 2017-01-24
Streaming API testing

zhostasa:
* Added StreamingOperations trait and invocation thereof
 * Added RandomStreamGenerator for streamed data generation
 * Added StreamingOperationsTestStage
 * Implemented Streaming on ISPN 9
 * Added ISPN 9 config file for testing
 * Added new benchmark for streaming tests
 * Minor POM changes

This is my first major addition into RG, all feedback is welcome

### DIFF
--- a/extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
+++ b/extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
@@ -1,0 +1,99 @@
+package org.radargun.stages.cache.generators;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Random;
+
+import org.radargun.config.DefinitionElement;
+
+/**
+ * Generator for {@link RandomStream} objects
+ * 
+ * @author zhostasa
+ */
+@DefinitionElement(name = "randomStream", doc = "Generates stream of random data, specifically for stream operation testing")
+public class RandomStreamGenerator implements ValueGenerator {
+
+   public Object generateValue(Object key, int size, Random random) {
+      return new RandomStream(size);
+   }
+
+   /**
+    * Will attempt to retrieve length of {@link RandomStream}
+    * 
+    * @return Size of {@link RandomStream} or 0
+    */
+   public int sizeOf(Object value) {
+      if (value instanceof RandomStream)
+         return ((RandomStream) value).length;
+      return 0;
+   }
+
+   /**
+    * If value is instance of {@link InputStream} validates the length against
+    * expectedSize
+    * 
+    * @param value
+    *           Should be {@link InputStream}
+    * @param expectedSize
+    *           Expected size of data
+    * 
+    * @return true if value is {@link InputStream} and returns expectedSize of
+    *         values
+    */
+   public boolean checkValue(Object value, Object key, int expectedSize) {
+      int count = 0;
+      try (InputStream is = (InputStream) value) {
+         while (is.read() != -1)
+            count++;
+      } catch (IOException e) {
+         return false;
+      }
+      return count == expectedSize;
+   }
+
+   /**
+    * Stream returning a predetermined count of random values
+    * 
+    * @author zhostasa
+    *
+    */
+   public class RandomStream extends InputStream {
+
+      private Random random;
+
+      private int length, processed;
+
+      /**
+       * 
+       * @param length
+       *           The number of int values the stream will return
+       */
+      public RandomStream(int length) {
+         random = new Random();
+         this.length = length;
+         reset();
+      }
+
+      /**
+       * @return Positive int or -1 upon reaching stream length
+       */
+      @Override
+      public int read() throws IOException {
+         return length > processed++ ? Math.abs(random.nextInt() & 0xFF) : -1;
+      }
+
+      /**
+       * Resets the stream to starting state
+       */
+      @Override
+      public void reset() {
+         processed = 0;
+         this.random = new Random();
+      }
+
+      public int getLength() {
+         return length;
+      }
+   }
+}
--- a/extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
+++ b/extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
@@ -1,5 +1,8 @@
 package org.radargun.stages.cache.test;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.Map;
 import java.util.Set;
 
@@ -8,12 +11,14 @@ import org.radargun.stages.test.Invocation;
 import org.radargun.traits.BasicOperations;
 import org.radargun.traits.BulkOperations;
 import org.radargun.traits.ConditionalOperations;
+import org.radargun.traits.StreamingOperations;
 import org.radargun.traits.TemporalOperations;
 
 /**
- * Provides {@link org.radargun.stages.test.Invocation} implementations for operations from traits
- * {@link org.radargun.traits.BasicOperations}, {@link org.radargun.traits.ConditionalOperations},
- * and {@link org.radargun.traits.BulkOperations}.
+ * Provides {@link org.radargun.stages.test.Invocation} implementations for
+ * operations from traits {@link org.radargun.traits.BasicOperations},
+ * {@link org.radargun.traits.ConditionalOperations}, and
+ * {@link org.radargun.traits.BulkOperations}.
  *
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  */
@@ -599,4 +604,83 @@ public class CacheInvocations {
          return async ? ASYNC_TX : NATIVE_TX;
       }
    }
+
+   public static final class GetViaStream<K, V extends Number> implements Invocation<Integer> {
+      public static final Operation GET_NULL = StreamingOperations.GET.derive("Null");
+      public static final Operation TX = StreamingOperations.GET.derive("TX");
+      private final StreamingOperations.StreamingCache<K> cache;
+      private final K key;
+      private Integer value;
+      private byte[] buffer;
+
+      public GetViaStream(StreamingOperations.StreamingCache<K> cache, K key, byte[] buffer) {
+         this.cache = cache;
+         this.key = key;
+         this.buffer = buffer;
+      }
+
+      @Override
+      public Integer invoke() {
+         Integer result = 0;
+         try (InputStream in = cache.getViaStream(key)) {
+            if (in == null)
+               return null;
+            int read = 0;
+            while ((read = in.read(buffer)) != -1)
+               result = result + read;
+            return value = result;
+         } catch (IOException e) {
+            throw new RuntimeException(e);
+         }
+      }
+
+      @Override
+      public Operation operation() {
+         return value == null ? GET_NULL : BasicOperations.GET;
+      }
+
+      @Override
+      public Operation txOperation() {
+         return TX;
+      }
+   }
+
+   public static final class PutViaStream<K, V extends InputStream> implements Invocation<Void> {
+      public static final Operation TX = StreamingOperations.PUT.derive("TX");
+      private final StreamingOperations.StreamingCache<K> cache;
+      private final K key;
+      private final InputStream value;
+      private byte[] buffer;
+
+      public PutViaStream(StreamingOperations.StreamingCache<K> cache, K key, V value, byte[] buffer) {
+         this.cache = cache;
+         this.key = key;
+         this.value = value;
+         this.buffer = buffer;
+      }
+
+      @Override
+      public Void invoke() {
+         try (OutputStream out = cache.putViaStream(key)) {
+            int read = 0;
+            while ((read = value.read(buffer)) != -1) {
+               out.write(buffer, 0, read);
+            }
+         } catch (IOException e) {
+            throw new RuntimeException(e);
+         }
+         return null;
+      }
+
+      @Override
+      public Operation operation() {
+         return StreamingOperations.PUT;
+      }
+
+      @Override
+      public Operation txOperation() {
+         return TX;
+      }
+   }
+
 }
--- a/extensions/cache/src/main/java/org/radargun/stages/cache/test/legacy/StreamingOperationsTestStage.java
+++ b/extensions/cache/src/main/java/org/radargun/stages/cache/test/legacy/StreamingOperationsTestStage.java
@@ -1,0 +1,80 @@
+package org.radargun.stages.cache.test.legacy;
+
+import java.io.InputStream;
+import java.util.Random;
+
+import org.radargun.Operation;
+import org.radargun.config.Namespace;
+import org.radargun.config.Property;
+import org.radargun.config.Stage;
+import org.radargun.stages.cache.test.CacheInvocations;
+import org.radargun.stages.test.Invocation;
+import org.radargun.stages.test.legacy.LegacyStressor;
+import org.radargun.stages.test.legacy.LegacyTestStage;
+import org.radargun.stages.test.legacy.OperationLogic;
+import org.radargun.stages.test.legacy.OperationSelector;
+import org.radargun.stages.test.legacy.RatioOperationSelector;
+import org.radargun.traits.InjectTrait;
+import org.radargun.traits.StreamingOperations;
+
+@Namespace(LegacyTestStage.NAMESPACE)
+@Stage(doc = "Streaming operations test stage")
+public class StreamingOperationsTestStage extends CacheOperationsTestStage {
+
+   @Property(doc = "Ratio of GET requests. Default is 4.")
+   protected int getRatio = 4;
+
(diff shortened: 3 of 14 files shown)


extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
```diff
@@ -599,4 +606,98 @@ public Operation txOperation() {
...
+      }
+   }
+
+   public static final class PutViaStream<K, V> implements Invocation<Void> {
+      public static final Operation TX = StreamingOperations.PUT.derive("TX");
+      private final StreamingOperations.StreamingCache<K> cache;
+      private final K key;
+      private final V value;
+
+      public PutViaStream(StreamingOperations.StreamingCache<K> cache, K key, V value) {
```
rvansa: When should this method get an instance of byte[] as argument? Why is the V a generic argument at all?

I think that `V` should be always an `InputStream` and in `invoke()` you should copy that to the `OutputStream`. Java 9 has `transferTo` method, but we can't use that, regrettably.

I wonder about buffering here, too - again delegate the decision to stage.

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Generator for {@link RandomStream} objects
+ * 
+ * @author zhostasa
+ */
+@DefinitionElement(name = "randomStream", doc = "Generates stream of random data, specifically for stream operation testing")
+public class RandomStreamGenerator implements ValueGenerator {
+
+   public Object generateValue(Object key, int size, Random random) {
+      return new RandomStream(key.hashCode(), size);
```
rvansa: Generating identical streams for the same key is not a good property of a *random value* generator.

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
```diff
@@ -599,4 +606,98 @@ public Operation txOperation() {
...
+      @Override
+      public V invoke() {
+         InputStream in;
+         Integer result = 0;
+         if ((in = cache.getViaStream(key)) == null)
+            return null;
+
+         try {
+            int read = 0;
+            byte[] arr = new byte[100];
```
rvansa: This is an ad-hoc buffering. Delegate buffer creation (or at least size definition) to stage. You can reuse the buffer there, too.

extensions/cache/src/main/resources/benchmark-stream-operations.xml
```diff
@@ -0,0 +1,132 @@
...
+            <server xmlns="urn:radargun:plugins:infinispan90:3.0" file="streaming.xml"
+               jmx-domain="jboss.datagrid-infinispan" start-timeout="120000" cache-manager-name="clustered">
+               <args>
+                  -Djava.net.preferIPv4Stack=true
+                  -Djboss.node.name=node${slave.index}
+                  -Djboss.socket.binding.port-offset=${slave.index}00
+               </args>
+               <home>${RADARGUN_WORK}/slave${slave.index}</home>
+               <server-zip>${ISPN_ZIP_PATH}/infinispan-server-9.0.0-SNAPSHOT.zip</server-zip>
+               <env>JAVA_OPTS=-server -Xms2g -Xmx4g -XX:NewRatio=3
```
rvansa: Do not use these complex options in example benchmark.

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
zhostasa: Yeah, it is not "Random" random, but I wanted the value to map back to the key. The naming is misleading I guess, what about "DataStreamGenerator"?

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+    *           Hashcode will he used as seed for {@link RandomStream}
+    * @param expectedSize
+    *           Expected size of data
+    * 
+    * @return true if value is byte[] and validation succeeded
+    */
+   public boolean checkValue(Object value, Object key, int expectedSize) {
+      boolean ret;
+      if (value instanceof byte[]) {
+         RandomStream rand = new RandomStream(key.hashCode(), expectedSize);
```
zhostasa: I will need you to re-word this comment, I do not understand you.

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+      public RandomStream(long seed, int length) {
+         this.seed = seed;
+         this.length = length;
+         reset();
+      }
+
+      /**
+       * @return Positive int or -1 upon reaching stream length
+       */
+      public int read() throws IOException {
```
zhostasa: Sure

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+         return length > processed++ ? Math.abs(random.nextInt()) : -1;
```
zhostasa: Good point

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+       * Validates one value against current stream state
+       * 
+       * @param i
+       * @return true if value is valid
+       */
+      public boolean validate(int i) {
+         try {
+            return i == read();
+         } catch (Exception e) {
+            e.printStackTrace();
```
zhostasa: Done

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+      public boolean validate(byte[] arr) {
+         for (byte b : arr) {
+            if (!validate(b))
+               return false;
+         }
+         return true;
+      }
+
+      public void close() {
+         try {
```
zhostasa: Right

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
zhostasa: Wouldn't it make it shared between invocations (and require synchronization and hell and damnation etc.)? I admit to be somewhat lost in this part of RG workings.

Size definition - sure

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
```diff
@@ -599,4 +606,98 @@ public Operation txOperation() {
...
+            while ((read = in.read(arr)) != -1) {
+               result = result + read;
+            }
+            in.close();
+         } catch (IOException e) {
+            e.printStackTrace();
```
zhostasa: I know, I added return null so the problem manifests in the results. Should there be runtime exception?

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
rvansa: Why do you map it to the key? When doing repeated writes, we usually want to change the value.

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
zhostasa: byte[] was used before I decided to add DataStream (former RandomStream). I opted to leave it in as side option, but I can get rid of it.

extensions/cache/src/main/java/org/radargun/stages/cache/test/legacy/StreamingOperationsTestStage.java
```diff
@@ -0,0 +1,75 @@
...
+      @Override
+      public void run(Operation operation) throws RequestException {
+         Object key = keyGenerator.generateKey(keySelector.next());
+         Random random = stressor.getRandom();
+
+         Invocation<Integer> invocation;
+         if (operation == StreamingOperations.GET) {
+            invocation = new CacheInvocations.GetViaStream(cache, key);
+         } else if (operation == StreamingOperations.PUT) {
+            invocation = new CacheInvocations.PutViaStream(cache, key,
```
zhostasa: It does, surprisingly:-) Funny that

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
rvansa: 1) creating a stream and closing that should be done using try-with-resources construct
2) I suggest moving the contents of `validate` to this method, rather than creating a new instance of `RandomStream` in order to execute the instance-method
3) I am wondering what is the code path that leads to executing this method with `byte[]` argument, since if the RSG created `RandomStream`, it should check `RandomStream`s again (or at least `InputStream`s).

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
rvansa: It would be shared between invocations from the same thread (if you write it that way), so no synchronization required.

extensions/cache/src/main/resources/benchmark-stream-operations.xml
zhostasa: What options are considered complex?

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
rvansa: Yes, that should be rather recorded as error. And please check for other uses of `e.printStackTrace()`.

extensions/cache/src/main/resources/benchmark-stream-operations.xml
rvansa: All JAVA_OPTS here can be omitted. You're demonstrating use of a feature.

extensions/cache/src/main/resources/benchmark-stream-operations.xml
```diff
@@ -0,0 +1,132 @@
...
+         <statistics>
+            <basic>
+               <operation-stats>
+                  <basic />
+               </operation-stats>
+            </basic>
+         </statistics>
+      </l:streaming-operations-test>
+
+      <rg:repeat name="StreamOperationsTest" from="1" to="11" inc="1">
```
zhostasa: The loop is done on increasing value sizes (counter*100kb to remain reasonable on single machine), but cutting it by half or two thirds would not decrease its exemplary value, good enough?

extensions/cache/src/main/resources/benchmark-stream-operations.xml
```diff
@@ -0,0 +1,132 @@
...
+         <l:streaming-operations-test test-name="Streaming_Operations_test"
+            num-threads-per-node="10" duration="30s" groups="client"
+            entry-size="#{${repeat.StreamOperationsTest.counter} * 100000}" amend-test="true"
+            iteration-property="entry-size">
+            <l:value-generator>
```
zhostasa: Yeah, the original test was alternating between this and BasicOperationsTestStage, it is redundant now.

extensions/cache/src/main/resources/benchmark-stream-operations.xml
rvansa: https://en.wikipedia.org/wiki/Minimalism

extensions/cache/src/main/resources/benchmark-stream-operations.xml
```diff
@@ -0,0 +1,132 @@
...
+            </l:value-generator>
+            <l:cache-selector>
+               <l:use-cache cache="streamed" />
+            </l:cache-selector>
+            <l:key-selector>
+               <l:concurrent-keys>
+                  <l:total-entries>100</l:total-entries>
+               </l:concurrent-keys>
+            </l:key-selector>
+            <statistics>
```
zhostasa: I am not sure what default is honestly, but it took a but of experimentation to get value which will not be liable to run the server out of memory.

plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90HotRodOperations.java
```diff
@@ -0,0 +1,40 @@
...
+   public Infinispan90HotRodOperations(InfinispanHotrodService service) {
+      super(service);
+   }
+
+   public <K, V> StreamingCache<K> getStreamingCache(String cacheName) {
+      HotRodCache<K, V> cache = super.getCache(cacheName);
+      return new StreamingHotRodCache<K, V>(cache.noReturn, cache.forceReturn, cache.noReturn.streaming());
+   }
+
+   protected class StreamingHotRodCache<K, V> extends HotRodOperations.HotRodCache<K, V>
```
zhostasa: Seeing that StreamingRemoteCache is just extension (not in programming sense) of the normal RemoteCache and all configuration applies to it, it did seem reasonable at the time, but you are right.

reporters/reporter-default/pom.xml
```diff
@@ -22,5 +21,10 @@
          <artifactId>freemarker</artifactId>
          <version>2.3.23</version>
       </dependency>
+      <dependency>
```
zhostasa: Yeah, didn't mean that

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
zhostasa: So the value can be validated when retrieved without actually storing it. But I just realized it could throw of results if the cluster was doing some sort of hash-based optimization. I will make this actually-random.

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
zhostasa: The buffer is a throwaway anyway, removed.

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
zhostasa: Done

extensions/cache/src/main/resources/benchmark-stream-operations.xml
zhostasa: I am pretty sure I am demonstrating a use case. Anyway, I left this configuration there specifically so there is some example of how to do this. I was unable to find one when I was starting this and would be grateful for one.

I will be more than happy to remove it if you can point me to a example I was unable to find. Otherwise I will remove it, but not be happy.

extensions/cache/src/main/resources/benchmark-stream-operations.xml
zhostasa: ... is ugly and non-practical. The idea of this test is to show memory consumption and GC utilization over time, which kinda requires some time. What is "too long"?

extensions/cache/src/main/java/org/radargun/stages/cache/generators/RandomStreamGenerator.java
zhostasa: I actually reduced validation to only length-based (upon reflection on these points)

zhostasa:
Did some changes, have a looksie

zhostasa:
Did some more changes, feel free to check it out

extensions/cache/src/main/resources/benchmark-stream-operations.xml
zhostasa: Default of what specifically?

zhostasa:
Changed the benchmark, sorry about the misunderstanding

extensions/cache/src/main/java/org/radargun/stages/cache/test/CacheInvocations.java
```diff
@@ -599,4 +604,83 @@ public Operation txOperation() {
...
+      public Integer invoke() {
+         Integer result = 0;
+         try (InputStream in = cache.getViaStream(key)) {
+            int read = 0;
+            while ((read = in.read(buffer)) != -1)
+               result = result + read;
+            return value = result;
+         } catch (IOException e) {
+            throw new RuntimeException(e);
+         } catch (NullPointerException e) {
```
rvansa: I think that `if (in == null) return null` is a slightly better way to deal with this...

zhostasa:
Done, any more nits to pick? :-)

Issue #442, opened 2016-12-12
Creation of timeline charts sometimes fails


rmacor:
Please see https://jdg-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/perf-cs-marshaller/23/console-edg-perf09-e7ec65ad-3dc6-4873-8b65-7d127efe7e09/ for full log.

ERROR [org.radargun.reporting.html.TimelineDocument] (main) Failed to generate on of the charts: 
java.util.concurrent.ExecutionException: java.lang.IllegalStateException: Current timestamp 1481467264825 is bigger then end timestamp 1481467264135
	at java.util.concurrent.FutureTask.report(FutureTask.java:122) ~[?:1.8.0_111]
	at java.util.concurrent.FutureTask.get(FutureTask.java:192) ~[?:1.8.0_111]
	at org.radargun.reporting.html.TimelineDocument.createTestCharts(TimelineDocument.java:180) [reporter-default-3.0.0-SNAPSHOT.jar:?]
	at org.radargun.reporting.html.HtmlReporter.writeTimelineDocuments(HtmlReporter.java:194) [reporter-default-3.0.0-SNAPSHOT.jar:?]
	at org.radargun.reporting.html.HtmlReporter.run(HtmlReporter.java:73) [reporter-default-3.0.0-SNAPSHOT.jar:?]
	at org.radargun.Master.run(Master.java:135) [radargun-core-3.0.0-SNAPSHOT.jar:?]
	at org.radargun.LaunchMaster.main(LaunchMaster.java:34) [radargun-core-3.0.0-SNAPSHOT.jar:?]
Caused by: java.lang.IllegalStateException: Current timestamp 1481467264825 is bigger then end timestamp 1481467264135
	at org.radargun.reporting.html.TimelineChart.setEvents(TimelineChart.java:100) ~[?:?]
	at org.radargun.reporting.html.TimelineDocument$1.call(TimelineDocument.java:137) ~[?:?]
	at org.radargun.reporting.html.TimelineDocument$1.call(TimelineDocument.java:130) ~[?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[?:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) ~[?:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) ~[?:1.8.0_111]

[...]

Issue #434, opened 2016-11-03
HtmlReporter fails with ConcurrentModificationException


vjuranek:
```
08:52:08,541 ERROR [org.radargun.Master] (main) Error in reporter org.radargun.reporting.html.HtmlReporter@16f7c8c1
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at org.radargun.reporting.html.TimelineDocument.<init>(TimelineDocument.java:64)
	at org.radargun.reporting.html.HtmlReporter.writeTimelineDocuments(HtmlReporter.java:191)
	at org.radargun.reporting.html.HtmlReporter.run(HtmlReporter.java:73)
	at org.radargun.Master.run(Master.java:135)
	at org.radargun.LaunchMaster.main(LaunchMaster.java:34)
```

vjuranek:
Improved exception is logged under #442

vjuranek:
Exception update:
```
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at org.radargun.reporting.html.TimelineDocument.<init>(TimelineDocument.java:67)
	at org.radargun.reporting.html.HtmlReporter.writeTimelineDocuments(HtmlReporter.java:194)
	at org.radargun.reporting.html.HtmlReporter.run(HtmlReporter.java:75)
	at org.radargun.Master.run(Master.java:140)
	at org.radargun.LaunchMaster.main(LaunchMaster.java:35)
```

PR #457, opened 2017-02-17
Monitor timeline bug fix

zhostasa:
Issue 434 - Concurrent modification exception and timetine bug fix 
 * POM dependency fix (for sake of Eclipse)
 * MasterMonitors added as MasterListener and stopped on configuration end
 * MasterListener.Adapter removed in favor of Java 8 default interface implementation
 * Comment added

 Overenthusiastic refactoring
 * United listeners under StateListener interface
 * Moved Listener management under StageBase class
 * Moved shared functionality from Monitors to AbstractMonitors class
 * Replaced interface Adapters with default interface implementation
 * Added some javadocs
 * Refactoring on MonitorStartStage

The first commit gets rid of the #434 and #442, which are related
The second commit is minor refactoring attempt gone somewhat more enthusiastic than originally intended. Especially in case of AbstractMonitors class I am not absolutely sure I did a right thing and hope for some input, bless my heart...


extensions/jpa/pom.xml
```diff
@@ -18,6 +19,11 @@
          <artifactId>radargun-core</artifactId>
          <version>${project.version}</version>
       </dependency>
+      <dependency>
```
rvansa: again this eclipse thing?

rvansa:
Maybe purpose is not the right word... But I find merging them as overgeneralization. And btw., keep commit titles a bit more 'professional', please :)

extensions/jpa/pom.xml
zhostasa: This is the last time I do this. Until next time, of course.

zhostasa:
I agree that they are hardly doing the same thing, but their usage is so similar that putting them under one interface and pushing the functionality to parent class, while not exactly pretty, seemed like a slightly better approach. That said, I am hardly authority on the project as a whole, so if you decide these changes are not beneficial I will remove them.

I will do the best ape can do to remain professional

extensions/jpa/pom.xml
alanfx: Don't pick on us Eclipse users, @rvansa!

rvansa: I would never mock elderly people!

core/src/main/java/org/radargun/state/SlaveState.java
```diff
@@ -68,7 +65,7 @@ public int getGroupSize() {
    public void reset() {
       super.reset();
       traits = null;
-      serviceListeners.clear();
+      clearlisteners();
```
rvansa: this should go to super.reset(), shouldn't it?

extensions/jpa/pom.xml
alanfx: Some dinosaurs are deadly!

core/src/main/java/org/radargun/state/SlaveState.java
zhostasa: Right, missed that one

zhostasa:
Fixed a minor issue

PR #458, opened 2017-02-21
Minor build fixes

zhostasa:
* excluded .xcf and .png files from assembly into conf folder
 * renamed rest-cs-tomcat configuration to match naming practices


jmarkos:
Looks good, thanks!

PR #461, opened 2017-03-02
New site mechanism addition

zhostasa:
* Added docs folder with original site converted into new format
 * Added make-site profile to pom to generate site on build

This is just converted site radargun.github.io for change reference and general approach discussion, more changes are coming. @mgencur @alanfx @rvansa @jmarkos are especially welcome to share their thoughts.


docs/measuring_performance/measuring_performance_rg1.md
```diff
@@ -0,0 +1,96 @@
+---
+---
+
+        <h3>
+          <a id="user-content-note-this-document-is-related-to-radargun-1x" class="anchor" href="#note-this-document-is-related-to-radargun-1x" aria-hidden="true"></a>Note: this document is related to RadarGun 1.x</h3>
```
alanfx: This still looks like an HTML file and not a Markdown file, but I also don't think it is referenced anywhere

docs/architecture/stage_list.md
```diff
@@ -0,0 +1,1060 @@
+---
+---
+
+Stage list
+----------
```
alanfx: I think the code that generates this document should be updated to include the namespace that the stage belongs in.

pom.xml
```diff
@@ -476,6 +476,36 @@
             <plugins.docker>docker</plugins.docker>
          </properties>
       </profile>
+      
+      <profile>
+         <id>site</id>
+         <activation>
+            <property>
+               <name>make-site</name>
+            </property>
```
alanfx: Just so no one else misses this. `-Pmake-site` doesn't work, and `-Dmake-site` does work

docs/measuring_performance/measuring_performance_rg1.md
zhostasa: Right, I left it there in case we wanted it used eventually, but it is not referenced on the original site either

docs/architecture/stage_list.md
zhostasa: I am under impression that html page I used as original for this MD file was generated from wiki page, which was in turn generated (as MD) by some SomethingSomethingGenerator or other @rvansa created some time ago. The plan to rewrite this to generate MD directly into page source is in the pipe, intended end-game for this and traits-list too. Also, I have this notion of making this more searchable/interactive, probably via some JS plugin, but this is in the wishful thinking stage presently.

pom.xml
zhostasa: I just finished rewriting Building binaries page to include build parameters, this one included, coming soon.

docs/architecture/stage_list.md
mgencur: What about excluding this part of documentation for now?

zhostasa: It is not incorrect, just incomplete and unreadable. I can make it not visible, but it is not like this is not exactly what is on the site currently and absence thereof would be better state of affairs (IMO of course).

docs/measuring_performance/measuring_performance_rg1.md
alanfx: IMO, I don't think we should keep any docs about the 1.x release. I also had a **thought** about how to version the docs. I was thinking the menu could include links to previous versions of the docs that would be stored in a subdirectory of the `site` folder. For example, the 2.x docs could be accessed in the path `site/other_versions/2.x/index.html`. WDYT?

zhostasa: Ok, I was not sure whether to keep it or lose it.
The versions - bigger topic, will talk to you in IRC

zhostasa:
I have pushed most recent changes (updates to original content) as separate commit to maintain original baseline. Writing custom plugins section of getting started is excluded from site for now as it will require a lot more attention and expertise to bring up to speed.

zhostasa:
Pushed latest batch of changes, the site is now completely validated -> AFAIK (!) it does not contain any outdated or incorrect information. It also, incidentally, does not contain a whole lot of content, a issue that will be the focus of my future endeavours.

zhostasa:
Reworked benchmark configuration section

mgencur:
@zhostasa a few more comments:
* building_binaries.html
    - Compile and build distribution    - wrong indentation
    - A Java Development Kit version 1.7 or above -> A Java Development Kit version 1.8 or above
    - Distributions’s content (for RadarGun 3.0.x) - let's remove the bracket and leave only "Distributions’s content"

* General benchmark configuration
    - it would be good to provide a most recent graph, the current one is 7 years old and the reports look different now
    - RadarGun is using a client/server architecture ->  RadarGun is using a master/slaves architecture    
    
* Configurations page
    neccesary -> necessary

* Scenario page
    - the list of items in the "Scenario design" paragraph is incorrectly rendered (as one block of text instead of individual list items)

* Stage list is outdated

* Failover tests
    - the page talks about "Legacy" tests. They're not legacy anymore. 
    - this page need a thorough review (later)

mgencur:
OK. Regarding Failover tests. The document includes very low level information and a lot of technical details. It is generally correct and it makes sense to have it there for someone who is really keen on technical details. But we need do a little update.  I suggest the following change: replace all of the following occurrences with "BackgroundStressorLogic":  "Legacy" (int the title), "legacy logic", "legacy mode"
The legacy logic has been renamed to BackgroundStressorLogic and we need to reflect it.

zhostasa:
Added changes proposed by @mgencur

mgencur:
Looks good. Integrated.

Issue #462, opened 2017-03-08
Remove experimental tests and move legacy tests back to main package.


mgencur:
The goal is to revert changes that were done in the commit with title "Refactoring of test stages to get rid of the coordinated omission problem"
The tests proved to be unusable with system under tests that do not fully support asynchronous calls (i.e. API supporting handler/callbacks and chaining them)

PR #463, opened 2017-03-08
gh-462 Remove experimental tests

mgencur:
Fixes gh-462

The last commit adds an alias legacy -> cache so it is possible to use both of them and leave the current benchmarks as they are.


extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStressorLogic.java
```diff
@@ -19,7 +19,7 @@
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  * @author Michal Linhard &lt;mlinhard@redhat.com&gt;
  */
-class LegacyLogic extends AbstractLogic {
+class BackgroundStressorLogic extends AbstractLogic {
```
rvansa: I think you were too eager in renaming this - LegacyLogic in background ops tests is not related to legacy/experimental tests.

extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStressorLogicConfiguration.java
```diff
@@ -7,11 +7,11 @@
 import org.radargun.stages.cache.generators.ValueGenerator;
 
 /**
- * Configuration specific to {@link LegacyLogic}
+ * Configuration specific to {@link BackgroundStressorLogic}
  *
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  */
-public class LegacyLogicConfiguration {
+public class BackgroundStressorLogicConfiguration {
```
rvansa: The same as LegacyLogic.

extensions/rest/src/main/java/org/radargun/stages/RESTOperationsTestStage.java
```diff
@@ -25,10 +26,12 @@
  *
  * @author Martin Gencur
  */
-@Namespace(TestStage.NAMESPACE)
+@Namespace(RESTOperationsTestStage.NAMESPACE)
 @Stage(doc = "Test using RESTOperations with specific URL")
 public class RESTOperationsTestStage extends TestStage {
 
+   public static final String NAMESPACE = "urn:radargun:stages:rest:" + Version.SCHEMA_VERSION;
```
rvansa: Why are you introducing a new namespace?

core/src/main/java/org/radargun/config/StageHelper.java
```diff
@@ -54,6 +55,15 @@ private static void addStage(String stageName, Class<? extends org.radargun.Stag
       }
    }
 
+   private static String handleLegacyNamespace(String namespace) {
```
rvansa: Uh, that's a nasty hack. What about adding a `deprecatedName` to `@Namespace` and properly registering the stage in `StageHelper.addStage`?

extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStressorLogic.java
mgencur: OK. But I don't see a reason why anything should have "Legacy" in its name. So let's take is as a step to better readability of this class.

extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStressorLogicConfiguration.java
mgencur: Same reply as above.

extensions/rest/src/main/java/org/radargun/stages/RESTOperationsTestStage.java
mgencur: Because the REST stage doesn't have much in common with "cache" namespace which is the default. This stage is for sending HTTP requests to a web application.

core/src/main/java/org/radargun/config/StageHelper.java
mgencur: Let me look into this.

extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStressorLogic.java
rvansa: The other (LogLogic) also use 'stressors' to execute the requests; but be it if you wish.

extensions/rest/src/main/java/org/radargun/stages/RESTOperationsTestStage.java
rvansa: Okay, makes sense. `rest:rest-operations-test` in XML duplicates the `rest`, but I don't know about more specific name for the operations.

mgencur:
@rvansa , second try

rvansa:
Looks good, integrated.

PR #468, opened 2017-04-05
Add redis32 plugin

jmarkos:
Few notes (also see the ones in RedisService javadoc)
- the RedisClientOperations was copied from https://github.com/GhostInAMachine/radargun/commit/e0b190bda2ccc342a9bcab551d94611684473f16 with minimal changes - if I understand https://tldrlegal.com/license/apache-license-2.0-(apache-2.0) correctly we can do this (under Can - Distribute - Describes the ability to distribute original or modified (derivative) works.)
- no example benchmark because it requires the steps from "Using dist.sh" from http://radargun.github.io/radargun/getting_started/using_the_scripts.html to work

### DIFF
--- a/docs/getting_started/building_binaries.md
+++ b/docs/getting_started/building_binaries.md
@@ -42,6 +42,7 @@ This operation might take a few minutes, as it will download the dependencies ne
         jcache
         jgroups30/32/33/34/35/36
         process
+        redis32
         resteasy-http
         spymemcached
         tomcat8
@@ -100,6 +101,7 @@ All properties have to be provided separatelly preceeded by "-D" -> property "sk
     * **jgroups**		(parameter `-Dno-jgroups` to disable)
     * **infinispan**		(parameter `-Dno-infinispan` to disable)
     * **process**		(parameter `-Dno-process` to disable)
+    * **redis**		(parameter `-Dno-redis` to disable)
     * **resteasy-http**		(parameter `-Dno-resteasy-http` to disable)
     * **spark**			(parameter `-Dno-spark` to disable)
     * **spymemcached**		(parameter `-Dno-spymemcached` to disable)
--- a/extensions/cache/src/main/resources/benchmark-redis.xml
+++ b/extensions/cache/src/main/resources/benchmark-redis.xml
@@ -12,7 +12,7 @@
    <configurations>
       <config name="redis">
          <setup group="server" plugin="redis32">
-            <server xmlns="urn:radargun:plugins:redis32:3.0" file="${env.RADARGUN_DISTRO_PATH}/plugins/redis32/conf/redis.conf"
+            <server xmlns="urn:radargun:plugins:redis32:3.0" file="redis.conf"
                     distribution-dir="/tmp/redisserver${slave.index}/" distribution-zip="${env.REDIS_DISTRO_ZIP_PATH}/redisdistro.zip" port="#{7000 + ${slave.index}}">
             </server>
          </setup>
@@ -27,12 +27,11 @@
    <init/>
    <rg:scenario xmlns:rg="urn:radargun:benchmark:3.0"
                 xmlns="urn:radargun:stages:core:3.0"
-                xmlns:cache="urn:radargun:stages:cache:3.0"
-                xmlns:l="urn:radargun:stages:legacy:3.0">
+                xmlns:cache="urn:radargun:stages:cache:3.0">
       <service-start groups="server" />
 
       <!--cluster redis-->
-      <command slaves="0" cmd="sh" args="-c" non-parsed-args="echo yes | /tmp/redisserver0/src/redis-trib.rb create --replicas 1 127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002 127.0.0.1:7003 127.0.0.1:7004 127.0.0.1:7005 " exit-values="0"/>
+      <command slaves="0" cmd="sh" args="-c" non-parsed-args="echo yes | /tmp/redisserver0/src/redis-trib.rb create --replicas 1 127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002 127.0.0.1:7003 127.0.0.1:7004 127.0.0.1:7005"/>
 
       <service-start groups="client" />
 
@@ -42,20 +41,20 @@
          </cache:key-generator>
       </cache:load>
 
-      <l:basic-operations-test groups="client" test-name="reads"
+      <cache:basic-operations-test groups="client" test-name="reads"
                                duration="30s"
                                num-threads-per-node="10" put-ratio="0"
                                get-ratio="1" entry-size="10000">
-         <l:key-selector>
-            <l:concurrent-keys total-entries="100000"/>
-         </l:key-selector>
-         <l:key-generator>
-            <l:byte-array-key/>
-         </l:key-generator>
+         <cache:key-selector>
+            <cache:concurrent-keys total-entries="100000"/>
+         </cache:key-selector>
+         <cache:key-generator>
+            <cache:byte-array-key/>
+         </cache:key-generator>
          <statistics>
             <common/>
          </statistics>
-      </l:basic-operations-test>
+      </cache:basic-operations-test>
 
       <service-stop />
    </rg:scenario>
--- a/extensions/cache/src/main/resources/benchmark-redis.xml
+++ b/extensions/cache/src/main/resources/benchmark-redis.xml
@@ -1,0 +1,69 @@
+<benchmark xmlns="urn:radargun:benchmark:3.0">
+
+   <master bindAddress="${master.address:127.0.0.1}" port="${master.port:2103}"/>
+
+   <clusters>
+      <cluster size="8">
+         <group name="server" size="6" />
+         <group name="client" size="2" />
+      </cluster>
+   </clusters>
+
+   <configurations>
+      <config name="redis">
+         <setup group="server" plugin="redis32">
+            <server xmlns="urn:radargun:plugins:redis32:3.0" file="${env.RADARGUN_DISTRO_PATH}/plugins/redis32/conf/redis.conf"
+                    distribution-dir="/tmp/redisserver${slave.index}/" distribution-zip="${env.REDIS_DISTRO_ZIP_PATH}/redisdistro.zip" port="#{7000 + ${slave.index}}">
+            </server>
+         </setup>
+         <setup group="client" plugin="redis32">
+            <jedis xmlns="urn:radargun:plugins:redis32:3.0" >
+               <servers>127.0.0.1:7000</servers>
+            </jedis>
+         </setup>
+      </config>
+   </configurations>
+
+   <init/>
+   <rg:scenario xmlns:rg="urn:radargun:benchmark:3.0"
+                xmlns="urn:radargun:stages:core:3.0"
+                xmlns:cache="urn:radargun:stages:cache:3.0"
+                xmlns:l="urn:radargun:stages:legacy:3.0">
+      <service-start groups="server" />
+
+      <!--cluster redis-->
+      <command slaves="0" cmd="sh" args="-c" non-parsed-args="echo yes | /tmp/redisserver0/src/redis-trib.rb create --replicas 1 127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002 127.0.0.1:7003 127.0.0.1:7004 127.0.0.1:7005 " exit-values="0"/>
+
+      <service-start groups="client" />
+
+      <cache:load groups="client" num-entries="100000" num-threads="10" entry-size="10000">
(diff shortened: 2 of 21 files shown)


pom.xml
```diff
@@ -247,6 +249,21 @@
          </properties>
       </profile>
       <profile>
+         <id>redis</id>
+         <activation>
+            <property>
+               <name>!no-redis</name>
```
zhostasa: This parameter should get added to "Building binaries" documentation, both as parameter and as part of directory structure

zhostasa:
Plugin info should get added to documentation

pom.xml
jmarkos: Done, thanks for the reminder.

docs/getting_started/building_binaries.md
```diff
@@ -100,6 +101,7 @@ All properties have to be provided separatelly preceeded by "-D" -> property "sk
     * **jgroups**		(parameter `-Dno-jgroups` to disable)
     * **infinispan**		(parameter `-Dno-infinispan` to disable)
     * **process**		(parameter `-Dno-process` to disable)
+    * **redis**		(parameter `-Dno-redis` to disable)
```
rvansa: indentation does not look correctly here

plugins/redis32/src/main/java/org/radargun/service/RedisService.java
```diff
@@ -0,0 +1,91 @@
...
+@Service(doc = "Redis")
+public class RedisService implements Lifecycle {
+
+   protected RedisServer server;
+
+   @Property(name = Service.FILE, doc = "Configuration file.")
+   protected String config;
+
+   @Property(doc = "Redis port")
+   protected Integer port = 6379;
```
rvansa: if you set default, you can use `int`, not `Integer`

plugins/redis32/src/main/resources/plugin.properties
```diff
@@ -0,0 +1,3 @@
+service.default org.radargun.service.RedisService
+service.server org.radargun.service.RedisService
+service.redis-client org.radargun.service.RedisClientService
```
rvansa: why not call that just `client` or `jedis`?

plugins/redis32/src/main/resources/redis1.conf
```diff
@@ -0,0 +1,6 @@
+bind test1
```
rvansa: Uh uh uh, I hope it's not necessary to put in several files and set the bind address to `testX` (which is not defined in RG anywhere). If you need the servers to have separate configs, RG should copy the file and overwrite bind address, e.g. by adding a `bindInterface` property (or letting it to bind to all interfaces on given machine).

plugins/redis32/src/main/resources/plugin.properties
jmarkos: I based it on hazelcast, where we call the client hazelcast-client. Jedis looks better, I'll change it.

docs/getting_started/building_binaries.md
jmarkos: In my IDE the indentation doesn't look correctly at all for the whole paragraph. It renders fine in https://github.com/jmarkos/radargun/blob/eee318cf85d85d64edcea6ae8bbc43431c38337a/docs/getting_started/building_binaries.md

rvansa: ok, nvm then

jmarkos:
Fixed all the review comments, added an example benchmark (doesn't work out of the box, some paths need to be provided).

zhostasa:
Please add the benchmark to getting_started/example_configuration.md, also list the environment variables like this (I am presently working on this for the rest):  
Environment variables:
* *ISPN_HOME*	- pointing to Infinispan home directory

jmarkos:
Fixed the documentation, thanks for the reminder.

extensions/cache/src/main/resources/benchmark-redis.xml
```diff
@@ -0,0 +1,69 @@
+<benchmark xmlns="urn:radargun:benchmark:3.0">
+
+   <master bindAddress="${master.address:127.0.0.1}" port="${master.port:2103}"/>
+
+   <clusters>
+      <cluster size="8">
```
rvansa: Do we need such "big" setup? Example configuration should consume as few resources as possible.
Have you ran this locally using the `bin/tests.sh` script?
You could also add a comment suggesting that you run one redis per core even in real benchmark.

extensions/cache/src/main/resources/benchmark-redis.xml
```diff
@@ -0,0 +1,69 @@
...
+         <group name="server" size="6" />
+         <group name="client" size="2" />
+      </cluster>
+   </clusters>
+
+   <configurations>
+      <config name="redis">
+         <setup group="server" plugin="redis32">
+            <server xmlns="urn:radargun:plugins:redis32:3.0" file="${env.RADARGUN_DISTRO_PATH}/plugins/redis32/conf/redis.conf"
```
rvansa: Could relative path work here? `RADARGUN_DISTRO_PATH` is not set by dist.sh/slave.sh scripts (IIRC).

extensions/cache/src/main/resources/benchmark-redis.xml
```diff
@@ -0,0 +1,69 @@
...
+                    distribution-dir="/tmp/redisserver${slave.index}/" distribution-zip="${env.REDIS_DISTRO_ZIP_PATH}/redisdistro.zip" port="#{7000 + ${slave.index}}">
+            </server>
+         </setup>
+         <setup group="client" plugin="redis32">
+            <jedis xmlns="urn:radargun:plugins:redis32:3.0" >
+               <servers>127.0.0.1:7000</servers>
```
rvansa: Why aren't we using the default port here?

extensions/cache/src/main/resources/benchmark-redis.xml
```diff
@@ -0,0 +1,69 @@
...
+
+   <init/>
+   <rg:scenario xmlns:rg="urn:radargun:benchmark:3.0"
+                xmlns="urn:radargun:stages:core:3.0"
+                xmlns:cache="urn:radargun:stages:cache:3.0"
+                xmlns:l="urn:radargun:stages:legacy:3.0">
+      <service-start groups="server" />
+
+      <!--cluster redis-->
+      <command slaves="0" cmd="sh" args="-c" non-parsed-args="echo yes | /tmp/redisserver0/src/redis-trib.rb create --replicas 1 127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002 127.0.0.1:7003 127.0.0.1:7004 127.0.0.1:7005 " exit-values="0"/>
```
rvansa: `exit-values="0"` is default

extensions/cache/src/main/resources/benchmark-redis.xml
```diff
@@ -0,0 +1,69 @@
...
+            </jedis>
+         </setup>
+      </config>
+   </configurations>
+
+   <init/>
+   <rg:scenario xmlns:rg="urn:radargun:benchmark:3.0"
+                xmlns="urn:radargun:stages:core:3.0"
+                xmlns:cache="urn:radargun:stages:cache:3.0"
+                xmlns:l="urn:radargun:stages:legacy:3.0">
```
rvansa: I don't think you need legacy namespace anymore

plugins/redis32/pom.xml
```diff
@@ -0,0 +1,34 @@
...
+      <dependency>
+         <groupId>org.radargun</groupId>
+         <artifactId>radargun-cache</artifactId>
+         <version>${project.version}</version>
+         <scope>provided</scope>
+      </dependency>
+      <dependency>
+         <groupId>eu.monniot.redis</groupId>
+         <artifactId>embedded-redis</artifactId>
+         <version>1.4.0</version>
```
rvansa: Make these dependencies `optional` to avoid transitivity with future releases.

extensions/cache/src/main/resources/benchmark-redis.xml
```diff
@@ -0,0 +1,69 @@
...
+         <group name="server" size="6" />
+         <group name="client" size="2" />
+      </cluster>
+   </clusters>
+
+   <configurations>
+      <config name="redis">
+         <setup group="server" plugin="redis32">
+            <server xmlns="urn:radargun:plugins:redis32:3.0" file="${env.RADARGUN_DISTRO_PATH}/plugins/redis32/conf/redis.conf"
+                    distribution-dir="/tmp/redisserver${slave.index}/" distribution-zip="${env.REDIS_DISTRO_ZIP_PATH}/redisdistro.zip" port="#{7000 + ${slave.index}}">
```
rvansa: Just wondering if you need to unpack the zip on each slave, even if the slaves are on the same node.  Could Redis instances share the same home dir?

extensions/cache/src/main/resources/benchmark-redis.xml
jmarkos: Actually that's a limitation of Jedis that I didn't mention - they have a separate API for singlenode/cluster usecases of Redis, and I'm using the cluster one, so the plugin won't work with anything less than 3 server nodes (3 masters, no slaves = no replicas).
I never ran the tests.sh script, didn't even know it existed.

extensions/cache/src/main/resources/benchmark-redis.xml
jmarkos: The embedded redis lib that I'm using expects a full path. The environment variable is documented (some of the example benchmarks require env variables to be set before working, so I didn't think it was a huge issue).

extensions/cache/src/main/resources/benchmark-redis.xml
jmarkos: I think they could share it, as long as it's recreated between runs, but I don't see the benefit?

extensions/cache/src/main/resources/benchmark-redis.xml
jmarkos: It's a relict, the redis cluster tutorial uses 7000 https://redis.io/topics/cluster-tutorial. I don't think it's a problem, at least it shows how to change the default one.

extensions/cache/src/main/resources/benchmark-redis.xml
rvansa: Why does it need to be recreated? I thought about setting just distribution-dir (no unzip). Benefit is reduced unpack time & disk space. But nothing critical at this point.

extensions/cache/src/main/resources/benchmark-redis.xml
rvansa: Ok, then, please, try it out. The script runs all example benchmarks to confirm that these match the current codebase etc. I usually ran that just before a release.

jmarkos: I think @zhostasa is currently working on making sure the example benchmarks are working (I'm fixing only the query ones).

extensions/cache/src/main/resources/benchmark-redis.xml
rvansa: Documented?
The contents of `conf` dir are added to the classpath; RadarGun should check if it's there and resolve that for the embedded lib. Other plugins don't need absolute paths either.
`ISPN_HOME` et all are fine; there's no other place to set it. But the distro itself can detect where it is automatically, so the least you could do is setting such env var in those scripts.

extensions/cache/src/main/resources/benchmark-redis.xml
jmarkos: Redis saves some cluster information into a file configured by "cluster-config-file nodes.conf", which is located in the distribution dir. It will fail to start if the file was written to in a previous run. I didn't look into it more since we do not plan any resilience tests.

extensions/cache/src/main/resources/benchmark-redis.xml
zhostasa: Yes, he is, I will get to this eventually.

plugins/redis32/src/main/java/org/radargun/service/RedisClientService.java
```diff
@@ -0,0 +1,63 @@
...
+   @Override
+   public void start() {
+      Set<HostAndPort> jedisClusterNodes = new HashSet<>();
+      for (InetSocketAddress server : servers) {
+         jedisClusterNodes.add(new HostAndPort(server.getHostName(), server.getPort()));
+      }
+      JedisPoolConfig poolConfig = new JedisPoolConfig();
+      poolConfig.setMaxTotal(100);
+      poolConfig.setMaxIdle(100);
+      poolConfig.setMinIdle(16);
```
jmarkos: make this configurable by properties

jmarkos:
Fixed everything (I can squash the commits if you want, but it's easier for you to review with 'fix' commits).

PR #469, opened 2017-04-12
Classpath scanning from Maven build

rvansa:
* Fast classpath scanner does not recognize AntClassLoader (which is used when invoking non-forking `java` task from Ant)


rvansa:
Oops, seems I've pushed to an upstream branch instead of my private. I shall remove that when this gets in.

jmarkos:
Merged, thanks, though I didn't verify it apart from building the distro.

PR #470, opened 2017-04-12
Configuration templates

rvansa:
* allows to build a hierarchy in configuration, overriding "base" settings
 * probably not perfect (maybe won't cover more complex use cases) but does not break existing configurations


zhostasa:
Cheers for putting this into writing in documentation:-)

jmarkos:
Fixed the 1 minor comment and merged, thanks!

PR #476, opened 2017-04-21
Fix example query benchmarks

jmarkos:
This works for me, I just did a small change to your work @zhostasa .

### DIFF
--- a/plugins/infinispan90/pom.xml
+++ b/plugins/infinispan90/pom.xml
@@ -152,7 +152,7 @@
             </property>
          </activation>
          <properties>
-            <version.infinispan>9.0.0-SNAPSHOT</version.infinispan>
+            <version.infinispan>9.0.0.Final</version.infinispan>
          </properties>
       </profile>
       <profile>
--- a/plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90EmbeddedQueryable.java
+++ b/plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90EmbeddedQueryable.java
@@ -1,0 +1,164 @@
+package org.radargun.service;
+
+import org.infinispan.query.Search;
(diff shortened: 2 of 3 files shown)


plugins/infinispan90/pom.xml
```diff
@@ -152,7 +152,7 @@
             </property>
          </activation>
          <properties>
-            <version.infinispan>9.0.0-SNAPSHOT</version.infinispan>
+            <version.infinispan>9.0.1-SNAPSHOT</version.infinispan>
```
rvansa: 9.0.0.Final is out, you should set it to that one.

plugins/infinispan90/src/main/java/org/radargun/service/Infinispan90EmbeddedQueryable.java
```diff
@@ -0,0 +1,160 @@
...
+   public Infinispan90EmbeddedQueryable(Infinispan90EmbeddedService service) {
+      super(service);
+   }
+
+   @Override
+   public Query.Builder getBuilder(String containerName, Class<?> clazz) {
+      return new QueryBuilder90Impl(Search.getQueryFactory(service.getCache(containerName)), clazz);
+   }
+
+   protected static class QueryBuilder90Impl extends QueryBuilder80Impl {
```
rvansa: Could you add a comment saying which signature change this handles?

jmarkos:
Fixed the 2 comments.

jmarkos:
Also rebased, but the continuous benchmark now fails with https://gist.github.com/jmarkos/9f5cb8fba23ee3d5d0e9f4585bd5def2 . I notified Roman, seems to be related to his changes.

rvansa:
This looks good anyway, merging.

PR #478, opened 2017-04-21
Asynchronous traits and tests

rvansa:


zhostasa:
Please add benchmark-dist-async.xml and benchmark-conversation.xml to docs/getting_started/example_configurations.md

Otherwise thumbs up for the Coordinated omission document.

extensions/cache/src/main/java/org/radargun/stages/cache/test/async/Conversations.java
```diff
@@ -0,0 +1,160 @@
...
+   }
+
+   @DefinitionElement(name = "put", doc = "cache.put()")
+   public static class Put extends RecordingConversationStep {
+      public CompletableFuture<Void> exec(ConversationContext ctx) {
+         ThreadLocalRandom random = ThreadLocalRandom.current();
+         ContextProvider provider = ctx.provider();
+         Object key = provider.getRandomKey(random);
+         Object value = provider.getRandomValue(random, key);
+         return provider.cache().put(key, value);
```
jmarkos: Is there a test/use case where the caches from ContextProvider and ConversationContext can be different? If yes, you should probably get the cache from the ctx in all steps (like in Get step).

core/src/main/java/org/radargun/stages/test/async/SchedulingSelector.java
```diff
@@ -14,12 +15,15 @@
...
+   private static final long LOW_WORD = 0xFFFFFFFFL;
    private final int[] invocations;
    private final long[] intervals;
    private final T[] operations;
-   private final AtomicLongArray lastIntervals;
-   private final AtomicIntegerArray todoInvocations;
-   private volatile int offset;
+   // Higher word of count contains the truncated lowest bits of the actual interval number, lower word contains
+   // number of operations executed in this interval.
+   private final AtomicLongArray counts;
```
jmarkos: What's the benefit of doing things this way? Seems a bit complicated.

Issue #474, opened 2017-04-19
HtmlReporter does not generate histograms


rvansa:
When a test is executed with statistics providing histogram data (use `<statistics><common /></statistics>`, the report should contain histogram & percentiles charts. This is not working ATM.

rvansa:
Remove `Statistics.getOperationsStats()` and `Statistics.getOperationStatsForGroups(...)` or make them return a multimap. `Statistics.getOperationStats(String)` is also wrong, this should return a collection.

PR #480, opened 2017-04-26
#474 fix generating histograms

rmacor:


core/src/main/java/org/radargun/stats/Statistics.java
```diff
@@ -37,14 +38,14 @@
    /**
     * @return Merged operation stats for registered groups. Calculation should be based on operations registered with this group.
     */
-   Map<String, OperationStats> getOperationStatsForGroups();
+   List<Map<String, OperationStats>> getOperationStatsForGroups();
```
rvansa: It should be rather Map<String, List, but the presence of this is wrong by design.

PR #486, opened 2017-05-05
Fix issue 485 with null check

alanfx:


jmarkos:
I don't see how the line https://github.com/radargun/radargun/blob/master/core/src/main/java/org/radargun/stages/lifecycle/LifecycleHelper.java#L70 can throw a NPE? If members=null, then msg will simply be "Number of members=null is not the one expected...". Am I missing something?

alanfx:
My mistake, it's line 69. `members.size()` causes the NPE.

jmarkos:
But if members=null, the members.size() won't be evaluated because of short-circuiting, no? Did you actually get a NPE on that line in some test?

alanfx:
Sigh, you are correct. I *think* I added `members.size()` to the logged msg variable because neither `"Number of members=null is not the one expected: <correct_number_of_members>"` nor `"Number of members=<an_array_of_members_whose_size_is_less_than_ expectedNumberOfSlaves> is not the one expected: <correct_number_of_members>"` are useful messages when trying to figure out how many members you actually have that cluster. This PR and the Issue should really be about correcting that situation, and I'm sorry for my confusion on this.

jmarkos:
Right. But with your change you won't get any log message if members is actually null, which also isn't very useful. Could you log something in that case?

alanfx:
Sure, I can do that

alanfx:
I pushed another commit that fixes the failing tests

jmarkos:
Merged, thanks!

Issue #477, opened 2017-04-21
java.lang.IllegalArgumentException: No operation for ID 16


jmarkos:
The benchmark-continuous-query.xml example benchmark fails with this PR https://github.com/radargun/radargun/pull/476 with 

13:22:03,191 INFO  [org.radargun.Master] (main) Running reporter org.radargun.reporting.html.HtmlReporter@28e8dde3
PerNodeRollingFileAppender::Not using file prefix.
13:22:05,047 ERROR [org.radargun.Master] (main) Error in reporter org.radargun.reporting.html.HtmlReporter@28e8dde3
java.lang.IllegalArgumentException: No operation for ID 16
        at org.radargun.Operation.getById(Operation.java:69) ~[radargun-core-3.0.0-SNAPSHOT.jar:?]
        at org.radargun.stats.BasicStatistics.ensure(BasicStatistics.java:103) ~[radargun-core-3.0.0-SNAPSHOT.jar:?]
        at org.radargun.stats.BasicStatistics.registerOperationsGroup(BasicStatistics.java:57) ~[radargun-core-3.0.0-SNAPSHOT.jar:?]
        at org.radargun.stats.MultiStatistics.registerOperationsGroup(MultiStatistics.java:117) ~[radargun-core-3.0.0-SNAPSHOT.jar:?]
        at org.radargun.reporting.commons.TestAggregations.addIteration(TestAggregations.java:105) ~[?:?]
        at org.radargun.reporting.commons.TestAggregations.<init>(TestAggregations.java:52) ~[?:?]
        at org.radargun.reporting.html.HtmlReporter.writeTestReportDocuments(HtmlReporter.java:176) ~[?:?]
        at org.radargun.reporting.html.HtmlReporter.run(HtmlReporter.java:77) ~[?:?]
        at org.radargun.Master.run(Master.java:140) [radargun-core-3.0.0-SNAPSHOT.jar:?]
        at org.radargun.LaunchMaster.main(LaunchMaster.java:34) [radargun-core-3.0.0-SNAPSHOT.jar:?]
13:22:05,056 INFO  [org.radargun.Master] (main) Running reporter org.radargun.reporting.csv.CsvReporter@669253b7

PR #488, opened 2017-05-15
 #477 fix benchmark-continuous-query report

rmacor:


jmarkos:
Removed the random newline on line 108 and integrated, thanks!

PR #492, opened 2017-06-26
Site width fix

zhostasa:
* added min-width limit to CSS in order to prevent folding of the navbar on narrow viewports

Did also some minor formatting in CSS


jmarkos:
Merged, thanks!

PR #496, opened 2017-09-15
Added some missing Mapreduce dependencies

zhostasa:
These went unnoticed because jdg plugins are not built by default...I think


vjuranek:
Merged, thanks!

Issue #495, opened 2017-07-18
Commit d0b1cb3ce1e465 (Remove partial throughput graphs) breaks reporting


jmarkos:
This benchmark https://gist.github.com/jmarkos/366f9bfcd46993f902101ac129734e13 (infinispan configuration is here https://gist.github.com/jmarkos/ac86e4bc922539f9f5b1d3f941d9bd43 ) isn't working correctly with the mentioned commit: the 2nd html reporter doesn't report the statistics for Transactional.* operations.

PR #497, opened 2017-10-23
#495 fix reporting when there are two reporters of the same type

rmacor:


alanfx:
@rmacor do you have a benchmark file that reproduces the error?

rmacor:
@alanfx  yes, there is a link in the issue description: https://github.com/radargun/radargun/issues/495

alanfx:
@rmacor Looks good. Merging...

PR #503, opened 2018-01-08
Clustered Counters support

mgencur:


rvansa:
@mgencur 'Strong' counters is name used exclusively in Infinispan. I would prefer to call the interface in trait just 'Counter', and let the contract (strong/weak/whatever) be specified by a string (or something easily expandable).
Also, it would be nice if you could provide Hazelcast implementation along to verify that the trait feels good for them, too (shouldn't be that much effort).

mgencur:
@rvansa, thanks for the tips. What do you think now?  ...the Counter interface now includes methods that are common to Infinispan and Hazelcast. There's no parameter to choose "WeakCounter". I don't plan to test it and Hazelcast doesn't support it. So in fact the only methods that remain are those from the StrongCounter.

extensions/counter/src/main/java/org/radargun/stages/CounterTestStage.java
```diff
@@ -0,0 +1,132 @@
...
+      INCREMENT_AND_GET, DECREMENT_AND_GET, ADD_AND_GET, COMPARE_AND_SET
+   }
+
+   @InjectTrait
+   protected CounterOperations counterOperations;
+
+   @Override
+   protected OperationSelector createOperationSelector() {
+      //TODO: replace this with direct selection
+      if (operationName.equals(OperationName.INCREMENT_AND_GET)) {
```
rvansa: operationName is an `enum`, use `switch`

extensions/counter/src/main/java/org/radargun/stages/CounterTestStage.java
```diff
@@ -0,0 +1,132 @@
...
+public class CounterTestStage extends TestStage {
+
+   public static final String NAMESPACE = "urn:radargun:stages:counter:" + Version.SCHEMA_VERSION;
+
+   @Property(doc = "Counter name.", optional = false)
+   protected String counterName;
+
+   @Property(doc = "Initial value of the counter expected by this stage. The test will start" +
+      "counting from this value. Default is 0.")
+   protected int initialValue = 0;
```
rvansa: Counter uses `long` values. The same for `delta`.

extensions/counter/src/main/java/org/radargun/stages/CounterTestStage.java
```diff
@@ -0,0 +1,132 @@
...
+      public void init(Stressor stressor) {
+         super.init(stressor);
+         this.counter = counterOperations.getCounter(counterName);
+         log.warn("Transactions ignored for Counter operations!");
+         stressor.setUseTransactions(false);//transactions for counter do not make sense
+      }
+
+      @Override
+      public void run(Operation operation) throws RequestException {
+         if (operation == CounterOperations.INCREMENT_AND_GET) {
```
rvansa: switch

extensions/counter/src/main/java/org/radargun/stages/CounterTestStage.java
```diff
@@ -0,0 +1,132 @@
...
+         } else if (operation == CounterOperations.ADD_AND_GET) {
+            Invocation<Long> invocation = new CounterInvocations.AddAndGet(counter, delta);
+            long currentValue = stressor.makeRequest(invocation);
+            if (currentValue == previousValue) {
+               throw new IllegalStateException("Inconsistent counter! Expected value different from previous.");
+            } else {
+               previousValue = currentValue;
+            }
+         } else if (operation == CounterOperations.COMPARE_AND_SET) {
+            //This operation is testable only with a single thread. Running with multiple threads would cause
```
rvansa: IMO a failed invocation is valid as well, just don't check that success is always true. Implementation might allow spurious failures.

rvansa: You could record success rate and add that to the results (user can then verify that success rate is 1/numThreads).

extensions/counter/src/main/resources/benchmark-clustered-counters.xml
```diff
@@ -0,0 +1,59 @@
...
+      <!-- other options for operation name: DECREMENT_AND_GET, ADD_AND_GET, COMPARE_AND_SET -->
+      <counter:counter-test test-name="incrementAndGetTest"
+                              num-operations="1000"
+                              num-threads-per-node="1"
+                              operation-name="INCREMENT_AND_GET"
+                              counter-name="c1"
+                              initial-value="0">
+      </counter:counter-test>
+
+      <!-- Due to a bug in RG we need to add num-threads-per-node * cluster size -->
```
rvansa: what bug?

plugins/hazelcast39/src/main/java/org.radargun.service/Hazelcast39CounterOperations.java
```diff
@@ -0,0 +1,55 @@
+package org.radargun.service;
+
+import com.hazelcast.core.IAtomicLong;
+import org.radargun.traits.CounterOperations;
+
+/**
+ * @author Martin Gencur
+ */
+public class Hazelcast39CounterOperations implements CounterOperations {
```
rvansa: Hazelcast supports IAtomicLong since 3.3 (or maybe earlier), so you can include it for earlier versions as well.

extensions/counter/src/main/java/org/radargun/stages/CounterTestStage.java
mgencur: The switch is not applicable here.

extensions/counter/src/main/resources/benchmark-clustered-counters.xml
mgencur: In Stressor#runInternal the the following method is run once at the beginning and later during normal execution. ```logic.run(operation)```
I'm not sure why this behavior is there but it means that each stressor calls the operation n+1 times (the additional call is IMO a bug)

rvansa: It is not run once; there are two loops: first executes requests (but not recording them) until stage.isStart() returns true, that is until all threads start (and the main thread observes that).

So it seems that in your case the main thread manages to kick in immediately after the first wait-for-others operation, but it is not guaranteed.

mgencur: OK. What is the reason to call the logic.run on some threads before all threads are started?

rvansa: If all threads started at once, there could be convoying due to concurrent conflicts etc... We should probably do even more of this ramp-up to randomize the times when threads send their requests.

mgencur:
Wait, let me fix the compareAndSet logic.

mgencur:
OK. This is ready.

rvansa:
@mgencur what's wrong with the CI? Could you fix it?

mgencur:
@rmacor, could you please look at the CI? It's weird

PR #504, opened 2018-01-10
Do not perform operations before all threads are ready

mgencur:
Here are some performance results before/after this change using the ClusteredCounter:
Common information: warmup has 100k operations, the main test has 100k operations as well, there are 10 threads per node, 2 nodes both running on the same node
Results:
Old: 5956 reqs/s  (overall number of operations performed (measured or not): 100402)
New:  5933 reqs/s ((overall number of operations performed (measured or not): 100000)
The difference is 0.39% and the test ran just a few seconds. With longer tests the difference would be probably even smaller but anyway it falls into the "grey" area where where test results jump up and down without an obvious reason.

### DIFF
--- a/core/src/main/java/org/radargun/stages/test/Stressor.java
+++ b/core/src/main/java/org/radargun/stages/test/Stressor.java
@@ -1,7 +1,8 @@
 package org.radargun.stages.test;
 
 import java.util.Random;
-import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.BrokenBarrierException;
+import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.ThreadLocalRandom;
 
 import org.radargun.Operation;
@@ -39,9 +40,9 @@ public class Stressor extends Thread {
    private Transactional.Transaction ongoingTx;
    private Statistics stats;
    private boolean started = false;
-   private CountDownLatch threadCountDown;
+   private CyclicBarrier checkpoint;
 
-   public Stressor(TestStage stage, OperationLogic logic, int globalThreadIndex, int threadIndex, boolean logTransactionExceptions, CountDownLatch threadCountDown, long delayBetweenRequests) {
+   public Stressor(TestStage stage, OperationLogic logic, int globalThreadIndex, int threadIndex, boolean logTransactionExceptions, CyclicBarrier checkpoint, long delayBetweenRequests) {
       super("Stressor-" + threadIndex);
       this.stage = stage;
       this.threadIndex = threadIndex;
@@ -51,7 +52,7 @@ public class Stressor extends Thread {
       this.completion = stage.getCompletion();
       this.operationSelector = stage.getOperationSelector();
       this.logTransactionExceptions = logTransactionExceptions;
-      this.threadCountDown = threadCountDown;
+      this.checkpoint = checkpoint;
       this.delayBetweenRequests = delayBetweenRequests;
    }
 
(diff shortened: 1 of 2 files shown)


rvansa:
Maybe you don't need an extra option, but make this a feature of the number-of-operations based test.

mgencur:
@rvansa, I don't see any valid argument against this PR based on real observations.

mgencur:
@rvansa 
Your previous point was following: "Let's take a uniprocessor and two threads. T1 sends a request, and the server processes it as the only request at that moment. Then T2 sends a request and this request 
is processed probably faster then the other one. That's why we let some threads do some preceeding requests so that the measured ones happen together with other ones."

My changes remove the part that sends requests before all threads are ready and rather makes sure that all threads start at the same point. I still don't understand how sending requests before all threads are ready helps.

We can either use duration or numOperations and and I don't understand why the behavior should be different. This request seems unreasoning. Anyway, when I want 1000 operation I don't really want 1010 operations and not even 1035. This is a functional bug of RG, IMO.

The perf. tests showed there's not significant difference in performance and it fixes the functional problem.

rvansa:
This test is not supposed to be run as a functional test. It's not a bug but intended feature. Despite you have not observed a significant difference in your particular test, you should not be skewing other tests.

alanfx:
@rvansa I do agree with @mgencur that if the benchmark species a specific number of operations and RadarGun executes more or less operations, that is a functional bug. I also still do not understand the rationale for performing any operations before all of the threads are ready in the case of using a duration. It seems like this is the same issue and will also affect the final statistics, since the actual duration for some threads will be longer than other threads.

In the argument above, both T1 and T2 would be measured since neither request was sent before **all** of the threads were ready to start executing operations.

rvansa:
Executing exact number of operations or running exactly X seconds is not the primary objective. The primary objective is to capture correct response times during a steady-state of the test. 

While both operations from the example above are recorded in the statistics, the first operation which is not executed in parallel with other (N - 1) operations cannot be counted as executing in steady state, but during ramp-up (which should not be recorded). In fact the current code has a bug: we should change the logic in such a way that each thread executes **at least one** operation before the measured operations start. And rather more than one, to provide a proper ramp-up.

Though, I agree that executing more ops can be confusing: while I would remove that option altogether (IMO in for benchmark it is only good for warmup, making sure that we execute enough operation to fire the JIT). If you want to keep it and abuse this for functional testing, I suggest disabling the initial operations **only** when `num-operations` is set.

alanfx:
> Executing exact number of operations or running exactly X seconds is not the primary objective. The primary objective is to capture correct response times during a steady-state of the test.

I disagree with this. Most people would think it is a bug if they ran a RadarGun benchmark that was supposed to execute a specific number of operations, and the results showed that more or less operations were actually executed. This also makes comparison between benchmarks impossible, since they will probably never run the same number of operations. 

Having said this, I do remember the last time we discussed this. (Which is when this code was changed the last time!) I understand your idea that the threads should start executing operations as soon as they can, but that the statistics for the operations should not be recorded until they are all started. That was the intent of the current implementation, and this goal makes sense to me so that you don't create a storm of operations by waiting until all threads are started before executing any operations. Given this, I think the "correct" fix is to keep the current code that lets threads execute operations, but fixes the problem with counting the operations for the `num-operations` case. None of statistics for the operations executed before all of the threads are started should be recorded, though. @mgencur WDYT?

rvansa:
> None of statistics for the operations executed before all of the threads are started should be recorded.

For `num-operations` you have three options:
1. execute some ops even before measured period (therefore more than `num` in total), record only num-operations (afaik this is the current state).
2. execute exactly `num-operations` but record only some of them: I think this would be even more confusing as stats would show less ops.
3. executing & recording exactly num-ops (as @mgencur proposes), with some operations less accurate

I have already said that I would accept 3., but I don't want it to affect the `duration` variant.

rvansa:
> Most people would think it is a bug if they ran a RadarGun benchmark that was supposed to execute a specific number of operations.

The problem is not in the code but in the way how we communicate what the stats mean and why have we done it this way.

alanfx:
(It's hard to reply to two different comments here, so I'll combine and quote them!)

> 1. execute some ops even before measured period (therefore more than num in total), record only num-operations (afaik this is the current state).

That's **not** the current state according to what @mgencur has seen in his results with more operations reported on than specified in the benchmark. The results should show the statistics for the `num-operations` specified in the benchmark. If more operations were executed and not recorded, that's ok with me.

> The problem is not in the code but in the way how we communicate what the stats mean and why have we done it this way.

I'm not sure how to communicate this in a way that makes sense, so instead I'd prefer to fix the code to behave in a more understandable way. (I can't believe I'm suggesting changing code and not fixing documentation to you!)

mgencur:
> That's not the current state according to what @mgencur has seen in his results with more operations reported on than specified in the benchmark. The results should show the statistics for the num-operations specified in the benchmark. If more operations were executed and not recorded, that's ok with me.
@alanfx , the truth is that these additional operations are not recorded in the statistics. I just spotted this by seeing the value of the clustered counter, which was at the end higher than num-operations. This means that more operations are performed but they were not recorded.

@rvansa, ok I understand what you mean now that Alan mentioned the storm of operations.

What do you think about option 4) : The individual operations triggered before we start measuring would be counted in the overall number (num-operations), but the statistics would be computed only from those operations there were actually measured (their response time etc.) ?

rvansa:
@mgencur That sounds like 2). The problem with 2) is that it is even more confusing to user, because he'd set `num-operations=1000` and see only 967 in statistics. Counters are sort of special as you can easily predict the final outcome of X operations (if you don't mix them in single test).

@alanfx Looking again to the code, I don't think that you could see more ops in statistics when you set `num-operations`. If you set `duration` and `delay-between-requests`, it is possible that the numbers don't quite match, but `num-operations` should be safe.

alanfx:
@rvansa ok, I think I finally understand what is going on. (again!) @mgencur explained that he is seeing the extra operations in the counter, but in the benchmark, only `num-operations` is being measured. Given this, I also agree that this change should be rejected as the code is working as designed.

However, since this is the second time we have been through this discussion about how this stage operates, I think some comments need to be added to the code to describe the behavior, so we don't have to rely on Martin and I to remember this when it comes up next time! :-)

mgencur:
Closing.

PR #505, opened 2018-01-15
Various stuff lying around

rvansa:
I've found bunch of unrelated uncommitted changes I made over past months... Can't remember exactly when and why.


plugins/docker/src/main/java/org/radargun/service/DockerService.java
```diff
@@ -185,23 +223,85 @@ private void printContainerLog(String containerId) {
...
+         } else {
+            buffer.rewind();
+         }
+         for (int i = 0; i < payload.length; i++) {
+            byte b = payload[i];
+            if (b == 27) { // ESCAPE
+               if (++i < payload.length && (b = payload[i]) == '[') {
+                  while (++i < payload.length) {
+                     b = payload[i];
+                     if (b >= '@' && b <= '~') break;
```
mgencur: What does "color escape sequence" look like? This break statement is called even if there's [ or ] so I'm trying to see what's the goal/purpose.

core/src/main/java/org/radargun/Operation.java
```diff
@@ -28,6 +28,13 @@ private Operation(int id, String name) {
       this.name = name;
    }
 
+   // This is useful when running report.sh
+   private Object readResolve() {
```
mgencur: This method seems to be unused.

core/src/main/java/org/radargun/RemoteSlaveConnection.java
```diff
@@ -410,12 +410,6 @@ private void startServerSocket() throws IOException {
       }
       serverSocketChannel.socket().bind(address);
       log.info("Master started and listening for connection on: " + address);
-      log.info("Waiting 5 seconds for server socket to open completely");
-      try {
-         Thread.sleep(5000);
```
mgencur: Why are you removing this? Maybe I should ask why it was there before.

rvansa: Yep, the latter is the correct question :) It has been there after some early committer (back at Mircea's days) seemingly resolved some problem. Does not make much sense to me now and I was running happily without.

plugins/docker/src/main/java/org/radargun/service/DockerService.java
rvansa: It is called only when the `[` is preceded by an escape char. Subsequently there are some numbers and the sequence is terminated with a letter (included).

mgencur:
@rvansa, there's a conflict, but otherwise looks good.

diegolovison:
Please rebase

PR #507, opened 2018-01-16
Enable logs in tests

rmacor:
There are some random failures in CI job, I can't reproduce them locally so at least enabling logs on tests so that we have something to investigate next time they happen.


rvansa:
@rmacor Why don't the test use `core/src/main/resources/log4j2.xml`? And log4j**2** is preferred.

rmacor:
@rvansa PR updated to use core/src/main/resources/log4j.xml

PR #511, opened 2018-01-25
Make JDG72 plugin depend on Infinispan91

mgencur:


rvansa:
While it probably works, I am a bit hesitant to base JDG7x plugins on Infinispan9x since JDG 7 was forked off Infinispan 8.2, despite the 9x features being backported.

mgencur:
What would you suggest? Creating empty subclasses of Infinispan8x (and Infinispan9x) plugin features in JDG plugins and pointing plugin.properties to them? (I mean only subclassing features that for sure work in both Infinispan 9x and JDG 7.x).

alanfx:
@rvansa are you worried about errors or that the code from `infinispan91` will be used instead of the code from `jdg72`? IMO, we should go ahead and base off of `infinispan91` and see how it behaves in practice. Otherwise, I am not sure how to handle this either.

rvansa:
@mgencur That's the 'correct' solution, but let's postpone it until we really need it. I don't know about better solution either.

mgencur:
OK. Let's wait until we need it.

PR #512, opened 2018-01-29
Log information about pid

diegolovison:
Sometimes we are facing an issue on Jenkins related with an
"Invalid process identifier".
With this commit, we are now able to know which is the pid number.
Also, I added a single quotation marks to quickly identify an empty
string.


alanfx:
@diegolovison LGTM, merging...

PR #513, opened 2018-02-28
PREVIEW: OpenShift support - library mode

mgencur:
Please review this implementation. It works with Employee openshift but should work with any other instance.

* There are some TBDs. Please ignore them.
* WDYT about building RadarGun before creating the Docker image vs. including the build in the image itself?
* I've exposed port 7800 by default on all RG slaves. This is specific to JGroups but other plugins could re-use this port. I just wanted to make one port available for intra-cluster communication which is exposed by the Docker container and accessible from other pods. We can possible choose a different port number.
* The JVM opts needs to be tuned and described in readme (this is TBD)

I can provide config files for JGroups and ISPN that worked for me in OpenShift (upon request).

Don't integrate this yet. But thoughts are welcome.

### DIFF
--- a/core/src/main/java/org/radargun/Slave.java
+++ b/core/src/main/java/org/radargun/Slave.java
@@ -93,10 +93,6 @@ public class Slave extends SlaveBase {
       Slave slave = new Slave(new RemoteMasterConnection(ArgsHolder.getMasterHost(), ArgsHolder.getMasterPort()));
       try {
          slave.run(ArgsHolder.getSlaveIndex());
-      } catch (IOException e) {
-         System.err.println("Communication with master failed");
-         e.printStackTrace();
-         ShutDownHook.exit(127);
       } catch (Exception e) {
          System.err.println("Unexpected error in scenario");
          e.printStackTrace();
--- a/core/src/main/java/org/radargun/Slave.java
+++ b/core/src/main/java/org/radargun/Slave.java
@@ -93,7 +93,12 @@ public class Slave extends SlaveBase {
       Slave slave = new Slave(new RemoteMasterConnection(ArgsHolder.getMasterHost(), ArgsHolder.getMasterPort()));
       try {
          slave.run(ArgsHolder.getSlaveIndex());
+      } catch (IOException e) {
+         System.err.println("Communication with master failed");
+         e.printStackTrace();
+         ShutDownHook.exit(127);
       } catch (Exception e) {
+         System.err.println("Unexpected error in scenario");
          e.printStackTrace();
          ShutDownHook.exit(127);
       }
--- a/core/src/main/resources/includes.sh
+++ b/core/src/main/resources/includes.sh
@@ -108,7 +108,7 @@ tail_log() {
    tail -f ${1} | while true
    do
       # Read from stdin with a 10 min timeout
-      IFS= read -r -u 0 -t 200
+      IFS= read -r -u 0 -t 600
       if (($? == 0))
       then
          echo "${REPLY}"
--- a/core/src/main/resources/includes.sh
+++ b/core/src/main/resources/includes.sh
@@ -108,7 +108,7 @@ tail_log() {
    tail -f ${1} | while true
    do
       # Read from stdin with a 10 min timeout
-      IFS= read -r -u 0 -t 600
+      IFS= read -r -u 0 -t 200
       if (($? == 0))
       then
          echo "${REPLY}"
@@ -119,7 +119,7 @@ tail_log() {
          fi
       else
          # Kill tail if no java process spawned from the shell script is found
-         if ! (ps -ef | grep "${3}" &>/dev/null)
+         if ! (ps -ef | grep "${3}" | grep -v "grep" | grep -v "tail" &>/dev/null)
          then
             kill -9 `pgrep -P $$ tail`
             break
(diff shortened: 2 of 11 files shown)


core/src/main/resources/includes.sh
```diff
@@ -108,7 +108,7 @@ tail_log() {
    tail -f ${1} | while true
    do
       # Read from stdin with a 10 min timeout
```
vjuranek: obsolete comment

openshift/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+#ENV RADARGUN_VERSION=
+#ENV DESCRIPTION=
+
+RUN yum -y update && \
+        yum -y install unzip java-1.8.0-openjdk-devel maven openssh-server which hostname ls ps grep pwd id rsync && \
+        yum clean all -y
+
+RUN mkdir /opt/radargun /opt/radargun-data 
+COPY RadarGun-3.0.0-SNAPSHOT /opt/radargun/
```
vjuranek: where `RadarGun-3.0.0-SNAPSHOT` comes from? It's not present there yet

vjuranek:
> WDYT about building RadarGun before creating the Docker image vs. including the build in the image itself?

IMHO should be built beforehands, ideally released, so that it's clear which version image uses. Building directly in the image make sense only when we want to test with snapshot (yeah, I know we use it a lot, and it pops up again - IMHO we should do RG releases more often and use only released version for testing)

openshift/Dockerfile
mgencur: It is copied from ${RADARGUN_HOME}/target/distribution/RadarGun-3.0.0-SNAPSHOT into the current directory before building the image. This is in openshift.sh in function build_radargun_image.

mgencur:
Thanks @vjuranek for the review! It's a good point that we should only build Docker images from stable versions. They could be pushed to Dockerhub then.

alanfx:
I agree with @vjuranek that this looks good, @mgencur! I can imagine that getting the server running will be more complicated, but this is a great step forward.

core/src/main/java/org/radargun/Slave.java
```diff
@@ -93,7 +93,12 @@ public static void main(String[] args) {
       Slave slave = new Slave(new RemoteMasterConnection(ArgsHolder.getMasterHost(), ArgsHolder.getMasterPort()));
       try {
          slave.run(ArgsHolder.getSlaveIndex());
+      } catch (IOException e) {
```
slaskawi: Your printing the exception out, so I guess this could be collapsed to `catch (Exception e)` (without catching IOException).

openshift/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM centos:centos7
```
slaskawi: How about using `jboss/base-jdk:8` ? or even a CE image if that's internal?

openshift/openshift.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/bash
```
slaskawi: make...file...make...file...make..file...

I know, you guys hate me for those makefiles everywhere ;)

openshift/Dockerfile
alanfx: I think using the base-jdk image makes sense, and I'd prefer not to use an internal CE image

openshift/openshift.sh
alanfx: Make? But Bash is Turing complete, isn't it?!

core/src/main/java/org/radargun/Slave.java
mgencur: Good point!

openshift/Dockerfile
mgencur: Yeah. I will change the base image.

openshift/openshift.sh
mgencur: I'll think about it. But all other scripts in RadarGun are Bash style. So some people will object that other scripts are Bash and some will object that running this script would be easier if it was a makefile :)

openshift/Dockerfile
mgencur: Actually. RadarGun is not a JBoss project per se. It's an independent project so I'm not sure basing it on JBoss is a good step. I'd rather not.

vjuranek: agree with @mgencur, as RG is cited as "standard testing tool", we shouldn't tie it to any specific project unless necessary.

openshift/openshift.sh
vjuranek: I'd consider using makefile only for building Docker image as it gives some hint there some dependencies between tasks and Dockerfile cannot be used standalone (as I was confused in comment above).

openshift/Dockerfile
alanfx: For me, I'm not too worried about using a JBoss image. I'm more concerned if the JBoss image includes the items that are being installed in this Dockerfile. If the JBoss image doesn't include anything extra except for Java, then I agree it's not worth changing.

mgencur:
Hey, I've pushed fixed in a separate commit here.

mgencur:
Closing. Superseded by https://github.com/radargun/radargun/pull/520

PR #518, opened 2018-03-02
Add plugins for snapshot builds and use parent POMs for versions

alanfx:
Adding the Infinispan or JDG parent POM in the dependencyManagement
section means that the version is pulled from the parent POM directly.

Added snapshot build plugins for JDG and Infinispan. The Infinispan and
JDG snapshot build plugins are in separate profiles.

Move the properties and profiles from the RadarGun top-level POM to the
parent POM, so that these properties can be referenced.


plugins/infinispan91/pom.xml
```diff
@@ -156,6 +146,18 @@
```
mgencur: In this pom file there's still the clustered-counter dependency defining the version. Is this redundant?

plugins/jdg-nightly/README.txt
```diff
@@ -0,0 +1,8 @@
+Data Grid nightly build RadarGun Plug-in
+--------------------------------------
+
+This plugin is used to host the nightly Data Grid builds.
+
+after that you need to rebuild RadarGun with
```
mgencur: Is there something missing on this line? It starts with lower case.

plugins/jdg-nightly/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+   <profiles>
+      <profile>
+         <id>jdg-nightly-infinispan-version-default</id>
+         <activation>
+            <property>
+               <name>!jdg-nightly.infinispan.version</name>
+            </property>
+         </activation>
+         <properties>
+            <version.infinispan>8.5.0.ER3-redhat-1</version.infinispan>
```
mgencur: Seems like this should be a snapshot.

plugins/jdg-nightly/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+   <dependencies>
+       <dependency>
+         <groupId>org.radargun</groupId>
+         <artifactId>plugin-jdg71</artifactId>
+         <version>${project.version}</version>
+       </dependency>
+
+       <dependency>
+           <groupId>org.radargun</groupId>
+           <artifactId>plugin-infinispan91</artifactId>
```
mgencur: Do we need to depend on infinispan91 when we already depend on jdg71/jdg72 ?

plugins/jdg-nightly/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+         <properties>
+            <version.infinispan>8.5.0.ER3-redhat-1</version.infinispan>
+         </properties>
+      </profile>
+   </profiles>
+
+   <dependencies>
+       <dependency>
+         <groupId>org.radargun</groupId>
+         <artifactId>plugin-jdg71</artifactId>
```
mgencur: Depend on plugin-jdg72?

plugins/infinispan91/pom.xml
alanfx: I'm not sure what you're asking about here. Should clustered counter not be included in the `infinispan91` plugin or are you asking about the `${project.version}`? Including the parent POM means that the POMs don't need to include the `${version.infinispan}` or `${version.narayanajta}` manually.

plugins/jdg-nightly/README.txt
alanfx: Hm, yeah this looks weird

plugins/jdg-nightly/pom.xml
alanfx: OK, I'll look at the versions coming out of the nightly builds and update this

plugins/jdg-nightly/pom.xml
alanfx: Nope, there should only be one here

plugins/jdg-nightly/pom.xml
alanfx: Yep, must have been sleeping on Friday!

plugins/infinispan91/pom.xml
mgencur: The clustered-counters dependency still has `${infinispan.version}`. That's what I mean.

alanfx: OK, sorry. For some reason, the `infinispan-clustered-counter`, `infinispan-marshaller-protostuff`, and `infinispan-marshaller-kryo` dependencies are not included in the parent POM in 9.1.x. In 9.2.x, this works. I'll take a look at it again and add a comment to this file.

alanfx:
@mgencur ok, I think I have addressed all of your comments

rvansa:
These are no 'nightly' builds, it uses any snapshot lying around in the local maven repo... let's call that just snapshots.
I would prefer dropping the `parent/pom.xml` and merging everything needed to root `pom.xml` if that's possible.

alanfx:
> These are no 'nightly' builds, it uses any snapshot lying around in the local maven repo... let's call that just snapshots.

Ok, I am not wed to the `-nightly` name, and you are correct that sometimes these are built in the day!

> I would prefer dropping the parent/pom.xml and merging everything needed to root pom.xml if that's possible.

On this one, I'm not sure I agree. I like the idea that the top-level POM aggregates the plugins, extensions, and reporters into the distribution, and that the parent POM is the imported into each of these various types. At some point I'd also like to centralize all of the version information in a BOM, but that can wait. WDYT, @mgencur?

mgencur:
TBH, I think RadarGun doesn't need a parent pom in a separate module. All of the stuff could be in the top-level pom, IMHO.
When it comes to BOM it would probably have to be in a separate module so that it can be released separately. So right now I don't understand the purpose of the refactoring of the poms as it is now.

alanfx:
I think that these changes actually make the parent POM useful, since plugins, extensions, and reporters all have a POM that points to the parent and not the top-level POM. I also like the idea of splitting the concerns some between the POMs, instead of having one giant top-level POM. The BOM would just give us one place to update versions for all of the plugins, instead of having to edit each plugin specific POM. Not a huge deal, IMO.

mgencur:
I'm not particularly happy with these changes because we will have to change two pom files instead of one every time we add new plugins/extension. But it's not a big deal. Whether the new plugins are called snapshot or nightly doesn't matter either. I'm integrating this. Let's see if it causes any troubles in the future.

alanfx:
Hey @mgencur thanks for merging this, but I'm not sure which two POMs you are referring to above. When adding a new plugin, you will only need to update the parent POM now, instead of the top-level POM. The `-snapshot` plugins will depend the new values of `infinispan-latest` and `jdg-latest`. They will require updates if the `plugin.properties` changes, but I don't see any way to update that automatically. Can you explain to me the situation you are worried about?

mgencur:
In the top-level pom there's a macro called copy-plugin-links and it specifies the plugin.list. And with new plugin you need to add your plugin there, AFAIK.

alanfx:
@mgencur thanks, I can see that now. I'll try to think of a way to solve this, but the `plugin.list` will only need to be modified when a completely new group of plugins is created. When just adding a new version of an existing plugin group, the top-level POM will not need to be modified.

PR #519, opened 2018-03-12
JDG-1171 MultimapCaches test.

zhabba:


plugins/jdg71/pom.xml
```diff
@@ -156,15 +156,15 @@
...
+      <dependencies>
          <dependency>
             <groupId>org.infinispan</groupId>
             <artifactId>infinispan-parent</artifactId>
-            <version>${version.infinispan}</version>
+           <version>${version.infinispan}</version>
             <type>pom</type>
             <scope>import</scope>
-         </dependency>
-      </dependencies>
```
mgencur: This formatting is wrong (and the two lines above as well).

plugins/jdg72/pom.xml
```diff
@@ -26,21 +26,16 @@
             </property>
          </activation>
          <properties>
-            <version.infinispan>8.5.0.ER3-redhat-1</version.infinispan>
+            <version.infinispan>8.5.0.ER3-redhat-9</version.infinispan>
```
mgencur: What about switching to 8.5.0.ER4-redhat-10 right away? It's already available.

plugins/jdg72/src/main/java/org/radargun/service/JDG72MultimapCacheOperations.java
```diff
@@ -0,0 +1,83 @@
...
+
+
+public class JDG72MultimapCacheOperations extends Infinispan92MultimapCacheOperations {
+
+   public JDG72MultimapCacheOperations(InfinispanEmbeddedService service) {
+      super(service);
+   }
+
+   @Override
+   public <K, V> MultimapCache<K, V> getMultimapCache(String cacheName) {
```
mgencur: Is it necessary to implement all these methods again? I think they're inherited from the Infinispan92 plugin automatically. IMO, this class should be more or less empty.

zhabba: Unfortunately, yes. I need to implement it because of class MultimapCache<K, V> has been placed in different packages. Upstream: org.infinispan.multimap.api.embedded.MultimapCache<K, V> and  downstream: org.infinispan.multimap.api.MultimapCache<K, V> and thus I can not simply extend upstream class.

extensions/multimap/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.radargun</groupId>
+            <artifactId>radargun-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.radargun</groupId>
+            <artifactId>radargun-cache</artifactId>
+            <version>3.0.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
zhabba: Fixed

plugins/jdg71/pom.xml
zhabba: Fixed

plugins/jdg72/pom.xml
zhabba: Fixed

plugins/jdg72/src/main/java/org/radargun/service/JDG72MultimapCacheOperations.java
mgencur: I'm not talking about extending the upstream class. You're already extending Infinispan92MultimapCacheOperations so the methods should be inherited from there.

plugins/jdg72/pom.xml
```diff
@@ -26,24 +26,24 @@
             </property>
          </activation>
          <properties>
-            <version.infinispan>8.5.0.ER3-redhat-9</version.infinispan>
+            <version.infinispan>8.5.0.ER3-redhat-10</version.infinispan>
```
mgencur: Ehmm, did you mean 8.5.0.ER4-redhat-10 ?

zhabba: Fixed.

mgencur:
Looks good!. Merging

PR #528, opened 2018-03-27
Etcd Plugin

diegolovison:
Etcd Plugin - https://github.com/coreos/etcd

### DIFF
--- a/plugins/etcd/src/main/java/org/radargun/service/EtcdClientOperations.java
+++ b/plugins/etcd/src/main/java/org/radargun/service/EtcdClientOperations.java
@@ -1,0 +1,135 @@
+package org.radargun.service;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+
+import com.coreos.jetcd.KV;
+import com.coreos.jetcd.data.ByteSequence;
+import com.coreos.jetcd.data.KeyValue;
+import com.coreos.jetcd.kv.DeleteResponse;
+import com.coreos.jetcd.kv.GetResponse;
+import com.coreos.jetcd.kv.PutResponse;
+import com.coreos.jetcd.options.DeleteOption;
+import com.coreos.jetcd.options.GetOption;
+import com.coreos.jetcd.options.PutOption;
+import org.radargun.traits.BasicOperations;
+
+public class EtcdClientOperations implements BasicOperations {
+
+   public static final PutOption GET_PUT_OPTION = PutOption.newBuilder().withPrevKV().build();
+   public static final DeleteOption GET_DELETE_OPTION = DeleteOption.newBuilder().withPrevKV(true).build();
+   public static final GetOption DEFAULT_GET_OPTION = GetOption.newBuilder().withSerializable(true).build();
+
+   protected final EtcdClientService service;
+
+   public EtcdClientOperations(EtcdClientService service) {
+      this.service = service;
+   }
+
+   @Override
+   public Cache<byte[], byte[]> getCache(String cacheName) {
+      return new EtcdCacheAdapter(service.kvClient);
+   }
+
+   private class EtcdCacheAdapter implements BasicOperations.Cache<byte[], byte[]> {
+
+      private KV kvClient;
+
+      public EtcdCacheAdapter(KV kvClient) {
+         this.kvClient = kvClient;
+      }
+
+      @Override
+      public byte[] get(byte[] key) {
+         Optional<KeyValue> value = this.retrieveValueFrom(key);
+         return retrieveBytes(value);
+      }
+
+      @Override
+      public boolean containsKey(byte[] key) {
+         Optional<KeyValue> value = this.retrieveValueFrom(key);
+         return value.isPresent();
+      }
+
+      @Override
+      public void put(byte[] key, byte[] value) {
+         try {
+            kvClient.put(ByteSequence.fromBytes(key), ByteSequence.fromBytes(value)).get();
+         } catch (InterruptedException | ExecutionException e) {
+            throw new IllegalStateException("Impossible to put the cache value", e);
+         }
+      }
+
+      @Override
+      public byte[] getAndPut(byte[] key, byte[] value) {
+         PutResponse response;
+         try {
+            response = kvClient.put(ByteSequence.fromBytes(key), ByteSequence.fromBytes(value), GET_PUT_OPTION).get();
+         } catch (InterruptedException | ExecutionException e) {
+            throw new IllegalStateException("Impossible to put the cache value", e);
+         }
+
+         if (response.hasPrevKv()) {
+            return response.getPrevKv().getValue().getBytes();
+         } else {
+            return null;
+         }
+      }
+
+      @Override
+      public boolean remove(byte[] key) {
+         DeleteResponse response;
+         try {
+            response = kvClient.delete(ByteSequence.fromBytes(key)).get();
+         } catch (InterruptedException | ExecutionException e) {
+            throw new IllegalStateException("Impossible to delete the cache", e);
+         }
+         return response.getDeleted() > 0;
+      }
+
+      @Override
+      public byte[] getAndRemove(byte[] key) {
+         DeleteResponse response;
+         try {
+            response = kvClient.delete(ByteSequence.fromBytes(key), GET_DELETE_OPTION).get();
+         } catch (InterruptedException | ExecutionException e) {
+            throw new IllegalStateException("Impossible to delete the cache", e);
+         }
+
+         List<KeyValue> prevKvs = response.getPrevKvs();
+         if (prevKvs != null && prevKvs.size() > 0) {
+            return prevKvs.get(0).getValue().getBytes();
+         } else {
+            return null;
+         }
+      }
+
+      @Override
+      public void clear() {
+         throw new UnsupportedOperationException("Clearing etcd is not supported.");
+      }
+
+      private byte[] retrieveBytes(Optional<KeyValue> optionalKeyValue) {
+         if (!optionalKeyValue.isPresent()) {
+            throw new NullPointerException("Cache value cannot be null");
+         }
+         return optionalKeyValue.get().getValue().getBytes();
+      }
+
+      private Optional<KeyValue> retrieveValueFrom(byte[] key) {
+         GetResponse response;
+         try {
+            response = kvClient.get(ByteSequence.fromBytes(key), DEFAULT_GET_OPTION).get();
+         } catch (InterruptedException | ExecutionException e) {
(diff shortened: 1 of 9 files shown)


plugins/etcd/src/main/java/org/radargun/service/EtcdClientService.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.coreos.jetcd.Client;
+import com.coreos.jetcd.KV;
+import org.radargun.Service;
+import org.radargun.config.Property;
+import org.radargun.traits.Lifecycle;
+import org.radargun.traits.ProvidesTrait;
+import org.radargun.utils.RedisAddressListConverter;
+
+@Service(doc = "Etdc client")
```
mgencur: Etdc -> Etcd

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
```diff
@@ -0,0 +1,79 @@
...
+
+   @ProvidesTrait
+   public EtcdService getSelf() {
+      return this;
+   }
+
+   @Override
+   public void start() {
+
+      try {
```
mgencur: We should probably assert that distributionDir and distributionZip are not both set by the user. The behaviour would be unexpected.

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
```diff
@@ -0,0 +1,79 @@
...
+   /**
+    * Stop the member etcd process
+    *
+    * https://coreos.com/etcd/docs/latest/v2/admin_guide.html
+    * ssh 10.0.1.11
+    * kill `pgrep etcd`
+    */
+   @Override
+   public void stop() {
+
```
mgencur: I can see that you're using this blank line in some methods but not all. Probably good to unify the approach.

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
diegolovison: In this case, the `distributionZip` will be extracted into the `distributionDir`.
The `distributionDir` it will be `servers/server-0`, `servers/server-1`, `servers/server-2`. It will change in each slave. This idea comes from the `redis` plugin.

In the HotRod we have different attributes values but the same behavior.
`home` is `distributionDir`
`server-zip` is `distributionZip`

plugins/etcd/src/main/java/org/radargun/service/EtcdClientOperations.java
```diff
@@ -0,0 +1,151 @@
...
+
+      @Override
+      public void put(byte[] key, byte[] value) {
+
+         sync((kvClient) -> kvClient.put(ByteSequence.fromBytes(key), ByteSequence.fromBytes(value)));
+      }
+
+      @Override
+      public byte[] getAndPut(byte[] key, byte[] value) {
+
```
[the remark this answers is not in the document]
diegolovison: No problem. Maybe you can see blank lines in a future PR because I use them everywhere :)

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
mgencur: OK. I get it now. The distributionDir is mandatory. The zip is optional. Cool.

diegolovison:
I kept the commit separated but if you would like to have then squashed please let me know.

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
```diff
@@ -17,17 +25,59 @@
...
+ * The documentation about the the configuration is available at https://coreos.com/etcd/docs/latest/op-guide/configuration.html
+ *
  */
 @Service(doc = "Etcd")
 public class EtcdService implements Lifecycle {
 
    private Process server;
 
-   @Property(doc = "Directory of the etcd distribution.", optional = false)
+   @Property(doc = "Distribution zip. If set, the zip will be extracted to the 'distributionDir' directory.", optional = false)
```
mgencur: I think the doc is outdated because the parameter is now mandatory.

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
```diff
@@ -17,17 +25,59 @@
...
+    * It is related with the --listen-peer-urls member flag. In this case the hostname will be dynamically full filled.
+    */
+   @Property(doc = "The listen peer port port number. Default is 2380")
+   protected int listenPeerPort = 2380;
+
+   /**
+    * It is realted with the --name member flag.
+    */
+   @Property(doc = "The server name. Each server must have an unique name")
+   protected String name;
```
mgencur: Is name really optional? You're later passing it to the command and it could be null at that point.

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
```diff
@@ -0,0 +1,164 @@
...
+ * Alternatively you can zip the compiled distribution, point to the zip with the distributionZip property and RG will unzip it to distributionDir for you.
+ *
+ * Note that etcd saves some data in the distribution directory, and you need to delete it between runs. Using the distributionZip
+ * property is preferred, so the distribution is 'clean' with each run.
+ *
+ * The documentation about the the configuration is available at https://coreos.com/etcd/docs/latest/op-guide/configuration.html
+ *
+ */
+@Service(doc = "Etcd")
+public class EtcdService implements Lifecycle {
```
rvansa: Why don't you inherit from ProcessLifecycle?

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
```diff
@@ -17,17 +25,59 @@
...
 
-   @Property(doc = "Directory of the etcd distribution.", optional = false)
+   @Property(doc = "Distribution zip. If set, the zip will be extracted to the 'distributionDir' directory.", optional = false)
+   protected String distributionZip;
+
+   /**
+    * It is realted with the --discovery member flag.
+    */
+   @Property(doc = "Prefix flags need to be set when using discovery service.", optional = false)
+   protected String discovery;
```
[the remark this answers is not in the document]
diegolovison: We are using public discovery service https://github.com/coreos/etcd/blob/master/Documentation/op-guide/clustering.md#public-etcd-discovery-service

In this case, we need to execute the following script before start the test
ETCD_DISCOVERY_URL=$(curl https://discovery.etcd.io/new?size=3)

In the RadarGun config file we need to set: discovery="${env.ETCD_DISCOVERY_URL}"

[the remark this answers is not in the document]
mgencur: Oh. This looks like something that could be done automatically.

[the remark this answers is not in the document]
mgencur: But, to be honest. I don't insist on that.

plugins/etcd/src/main/java/org/radargun/service/EtcdClientService.java
```diff
@@ -0,0 +1,58 @@
...
+
+@Service(doc = "Etcd client")
+public class EtcdClientService implements Lifecycle {
+
+   private Client client;
+
+   KV kvClient;
+
+   @Property(doc = "List of server addresses (in host:port format) the clients should connect to, separated by semicolons (;).", converter = RedisAddressListConverter.class)
+   protected List<InetSocketAddress> servers;
```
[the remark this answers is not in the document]
diegolovison: https://github.com/radargun/radargun/issues/529

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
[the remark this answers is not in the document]
rvansa: The public discovery service is nice for demos, but could we start the discovery service directly in the cluster?

plugins/etcd/src/main/java/org/radargun/service/EtcdClientOperations.java
```diff
@@ -0,0 +1,135 @@
...
+   public EtcdClientOperations(EtcdClientService service) {
+      this.service = service;
+   }
+
+   @Override
+   public <K, V> Cache<K, V> getCache(String cacheName) {
+      return new EtcdCacheAdapter(service.kvClient);
+   }
+
+   private class EtcdCacheAdapter<K, V> implements BasicOperations.Cache<byte[], byte[]> {
```
rvansa: The generic parameters are not used.

plugins/etcd/src/main/java/org/radargun/service/EtcdService.java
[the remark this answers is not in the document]
diegolovison: Per my understanding we have 3 options:
Static - Need to know all the clusters members before start each node
etcd Discovery - We can create a private or use the public
DNS Discovery - We need to create DNS SRV records

We can built our private etcd discovery using the source code: https://github.com/coreos/discovery.etcd.io

IMO, I will prefer use the public discovery service and if for some reason it goes down or stop working we can create our server.

I was not able to find an way where we can start a discovery service directly in the cluster.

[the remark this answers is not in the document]
rvansa: OK, if it's not easy to set it up yourselves, let's forget about it.

Note that rather than setting env vars manually (you need call the curl on single node and pass that to all slaves, don't you?) you can run the command from benchmark itself using `<command cmd="curl" args="https://discovery.etcd.io/new?size=3" var="discovery.url"/>`. That would run it from all slaves, though; you could enhance that stage to distribute the result to all nodes overriding `processAckOnMaster`. That should go to different PR, though.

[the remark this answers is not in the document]
diegolovison: Sounds good to me. It will fix the @mgencur suggestion as well.

https://github.com/radargun/radargun/issues/531

plugins/etcd/src/main/java/org/radargun/service/EtcdClientOperations.java
```diff
@@ -0,0 +1,135 @@
...
+         }
+      }
+
+      @Override
+      public void clear() {
+         throw new UnsupportedOperationException("Clearing etcd is not supported.");
+      }
+
+      private byte[] retrieveBytes(Optional<KeyValue> optionalKeyValue) {
+         if (!optionalKeyValue.isPresent()) {
```
rvansa: When you use `Optional` you should rather use `.orElseThrow(...).getValue().getBytes()`.

rvansa:
Integrated, thanks.

PR #534, opened 2018-04-10
Rss Monitor

diegolovison:
Rss Monitor

### DIFF
--- a/core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
+++ b/core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
@@ -14,10 +14,12 @@ import org.radargun.sysmonitor.MasterMonitors;
 import org.radargun.sysmonitor.MemoryUsageMonitor;
 import org.radargun.sysmonitor.NetworkBytesMonitor;
 import org.radargun.sysmonitor.OpenFilesMonitor;
+import org.radargun.sysmonitor.RssMonitor;
 import org.radargun.sysmonitor.SlaveMonitors;
 import org.radargun.traits.InjectTrait;
 import org.radargun.traits.InternalsExposition;
 import org.radargun.traits.JmxConnectionProvider;
+import org.radargun.utils.SystemUtils;
 import org.radargun.utils.TimeConverter;
 
 /**
@@ -75,6 +77,9 @@ public class MonitorStartStage extends AbstractDistStage {
       monitors.addMonitor(new MemoryUsageMonitor(jmxConnectionProvider, timeline));
       monitors.addMonitor(new GcMonitor(jmxConnectionProvider, timeline));
       monitors.addMonitor(new OpenFilesMonitor(jmxConnectionProvider, timeline));
+      if (SystemUtils.IS_LINUX) {
+         monitors.addMonitor(new RssMonitor(timeline));
+      }
       if (interfaceName != null) {
          monitors.addMonitor(NetworkBytesMonitor.createReceiveMonitor(interfaceName, timeline));
          monitors.addMonitor(NetworkBytesMonitor.createTransmitMonitor(interfaceName, timeline));
--- a/core/src/main/java/org/radargun/sysmonitor/RssMonitor.java
+++ b/core/src/main/java/org/radargun/sysmonitor/RssMonitor.java
@@ -1,0 +1,60 @@
+package org.radargun.sysmonitor;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+
+import org.radargun.reporting.Timeline;
+import org.radargun.utils.Utils;
+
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/radargun/sysmonitor/RssMonitor.java
```diff
@@ -0,0 +1,69 @@
...
+         timeline.addValue(Timeline.Category.sysCategory(RSS_MEMORY_USAGE), new Timeline.Value(rssUsage));
+      } finally {
+         if (process != null) process.destroy();
+      }
+   }
+
+   private Process createProcess() {
+      Process process;
+      try {
+         String command = "ps --no-headers -o rss " + getCurrentPid();
```
alanfx: Do all versions of `ps` support these options? Just wondering since this monitor will only work on Linux.

core/src/main/java/org/radargun/sysmonitor/RssMonitor.java
```diff
@@ -0,0 +1,69 @@
...
+         ProcessBuilder builder = new ProcessBuilder().command("bash", "-c", command);
```
alanfx: Just wondering if this works without a full path to `bash`? Adding the path will make it even less portable though!

core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
```diff
@@ -75,6 +76,7 @@ private void addMonitors(AbstractMonitors monitors, Timeline timeline) {
       monitors.addMonitor(new MemoryUsageMonitor(jmxConnectionProvider, timeline));
       monitors.addMonitor(new GcMonitor(jmxConnectionProvider, timeline));
       monitors.addMonitor(new OpenFilesMonitor(jmxConnectionProvider, timeline));
+      monitors.addMonitor(new RssMonitor(timeline));
```
alanfx: You should probably add a platform check here before creating the monitor. RadarGun doesn't run on Windows yet, but that's in the plans.

core/src/main/java/org/radargun/sysmonitor/RssMonitor.java
diegolovison: It is working on Fedora and Ubuntu.

core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
```diff
@@ -6,18 +6,11 @@
...
-import org.radargun.sysmonitor.AbstractMonitors;
-import org.radargun.sysmonitor.CpuUsageMonitor;
-import org.radargun.sysmonitor.GcMonitor;
-import org.radargun.sysmonitor.InternalsMonitor;
-import org.radargun.sysmonitor.MasterMonitors;
-import org.radargun.sysmonitor.MemoryUsageMonitor;
-import org.radargun.sysmonitor.NetworkBytesMonitor;
-import org.radargun.sysmonitor.OpenFilesMonitor;
-import org.radargun.sysmonitor.SlaveMonitors;
+import org.radargun.sysmonitor.*;
```
rvansa: Could you configure your IDE to not use wildcard imports, please?

core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
```diff
@@ -75,6 +68,11 @@ private void addMonitors(AbstractMonitors monitors, Timeline timeline) {
       monitors.addMonitor(new MemoryUsageMonitor(jmxConnectionProvider, timeline));
       monitors.addMonitor(new GcMonitor(jmxConnectionProvider, timeline));
       monitors.addMonitor(new OpenFilesMonitor(jmxConnectionProvider, timeline));
+      if (SystemUtils.IS_LINUX) {
+         monitors.addMonitor(new RssMonitor(timeline));
+      } else {
+         log.warn("Rss monitor only works on Linux.");
```
rvansa: User did not ask for RSS explicitly, don't bother him with warnings about running wrong OS :)

core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
diegolovison: Should we update the IDE settings? Today it is set equals 10.
https://github.com/radargun/radargun/blob/3359ca7f6174116b1844719685fe8ee7a5123c85/ide-settings/intellij/RadarGunCodeStyle.xml#L17

rvansa: Hmm, I didn't know about these settings. Yes, no wildcards is better.

diegolovison: https://github.com/radargun/radargun/pull/541/files

sparkoo:
does this need any more changes? I think all review comments were resolved.

rvansa:
Merged, thanks.

PR #535, opened 2018-04-10
Class loading count

diegolovison:


alanfx:
@diegolovison you need to resolve conflicts on this one

PR #538, opened 2018-04-17
don't try to monitor open files on Windows

sparkoo:
Hi,
I've been running radargun on Windows and many exceptions occurred with OpenFilesMonitor (https://gist.github.com/sparkoo/d26f359f20683e2b8bde4a218965bbe0). Issue is that java on windows can't do that. There is no windows version of unix's MXBean `UnixOperatingSystemMXBean`. So I've disabled this monitor for windows.
The if condition is not very nice. I can put it to `Utils` class, or `commons.lang` has nicer method `SystemUtils.IS_OS_WINDOWS`, but it's not dependency of `core` module at the moment.

### DIFF
--- a/core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
+++ b/core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
@@ -76,9 +76,12 @@ public class MonitorStartStage extends AbstractDistStage {
       monitors.addMonitor(new CpuUsageMonitor(jmxConnectionProvider, timeline));
       monitors.addMonitor(new MemoryUsageMonitor(jmxConnectionProvider, timeline));
       monitors.addMonitor(new GcMonitor(jmxConnectionProvider, timeline));
-      monitors.addMonitor(new OpenFilesMonitor(jmxConnectionProvider, timeline));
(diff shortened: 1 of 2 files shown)


diegolovison:
There is an utility class that is coming to address OS issues.
https://github.com/diegolovison/radargun/blob/7f6ff8188e2e6ab4ab084781752e2f6eeae97727/core/src/main/java/org/radargun/utils/SystemUtils.java

It is pending the approval of https://github.com/radargun/radargun/pull/534

sparkoo:
ah, I haven't notice that. Once #534 is merged, I'll update this one.

sparkoo:
I updated PR so now isWindows condition is taken from SystemUtils. `OpenFileMonitor` is disabled only for windows.

rvansa:
@sparkoo The branch needs a rebase.

sparkoo:
Can you use "Squash and merge"?

mgencur:
Michal, we're trying to keep linear history of commits. Hence the rebase is needed. Thanks

sparkoo:
I understand, that's why I suggested to use "Squash and merge" which creates one commit in parent repo, regardless to fork repo. It's one button operation.

rvansa:
Thanks, integrated.

PR #541, opened 2018-04-18
Increased the number. We cannot disabled this option

diegolovison:


mgencur:
The commit message could be more descriptive but otherwise looks good. Integrating.

PR #543, opened 2018-04-20
Fix for issue 434

alanfx:
https://github.com/radargun/radargun/issues/434

I've addressed this issue with three changes:

1) Increase the timeout in `Utils.shutdownAndWait()` to 30 seconds

2) Call `Utils.shutdownAndWait()` on the `ExecutorService` used by
statistics and monitors

3) Restructure the monitors code to add no more entries to the
TimelineDocument after `stop()` is called

I've verified that this fix stops the
`java.util.ConcurrentModificationException` from happening

### DIFF
--- a/core/src/main/java/org/radargun/sysmonitor/AbstractMonitor.java
+++ b/core/src/main/java/org/radargun/sysmonitor/AbstractMonitor.java
@@ -1,0 +1,31 @@
+package org.radargun.sysmonitor;
+
+/**
+ * 
+ * AbstractMonitor is the base class for all system monitors. It manages the lifecycle of the
+ * monitor, so that no values are written to the reports before the benchmark starts or ends.
+ * 
+ */
+public abstract class AbstractMonitor implements Monitor {
+   protected volatile boolean shouldRun = false;
+
+   @Override
+   public void run() {
+      if (shouldRun) {
+         runMonitor();
+      }
+   }
+
+   @Override
+   public void start() {
+      shouldRun = true;
+   }
+
+   @Override
+   public void stop() {
+      shouldRun = false;
+   }
+
+   public abstract void runMonitor();
+
+}
--- a/extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStatisticsManager.java
+++ b/extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStatisticsManager.java
@@ -16,9 +16,11 @@ import org.radargun.state.SlaveState;
 import org.radargun.stats.Statistics;
 import org.radargun.stats.representation.OperationThroughput;
 import org.radargun.utils.TimeService;
+import org.radargun.utils.Utils;
 
 /**
- * Coordinator for collecting statistics from background stressor threads started by {@link org.radargun.stages.cache.background.BackgroundOpsManager}.
+ * Coordinator for collecting statistics from background stressor threads started by
+ * {@link org.radargun.stages.cache.background.BackgroundOpsManager}.
  *
  * @author Matej Cimbora &lt;mcimbora@redhat.com&gt;
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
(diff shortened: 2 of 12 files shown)


alanfx:
Marking as preview until I can try it to see if it fixes the issue.

rvansa:
@alanfx And *why* should that help? The SO post above replaces direct invocation of `Collection.remove` with `Iterator.remove`; you don't use the iteration-removal anywhere.

rvansa:
The code inside the loop does not modify the collection, so it seems that some thread was not stopped properly and keeps adding data to the collection. My guess would be that `BackgroundStatisticsManager.StatsTask` was not stopped properly. Or some of the monitors.

alanfx:
@rvansa yeah, adding a few Iterators has stopped the issue from happening in the CSV reporter, but the HTML and Serialized reporters are still getting errors. I'll take a thread dump before starting the reporters to see if you're correct. Thanks

alanfx:
@rvansa I have been able to see that the `RSSMonitor` seems to have some threads running, so I am working on a fix for that now. I think the `ExecutorService` is not being terminated as well.

alanfx:
@rvansa updated with an actual fix!

extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStatisticsManager.java
```diff
@@ -82,8 +83,19 @@ public synchronized void startStats() {
...
       if (statsTask != null) {
-         statsTask.cancel(true);
-         statsTask = null;
+
+         if (statsTask.cancel(true)) {
+            statsTask = null;
+         } else {
+            log.error("Statistics task was not cancelled");
+         }
+         Utils.shutdownAndWait(executor);
```
rvansa: If you terminate the executor, please make sure that a new one is created in `startStats`. The stats can be started/stopped several times during a scenario.

core/src/main/java/org/radargun/sysmonitor/AbstractMonitor.java
```diff
@@ -0,0 +1,31 @@
+package org.radargun.sysmonitor;
+
+/**
+ * 
+ * AbstractMonitor is the base class for all system monitors. It manages the lifecycle of the
+ * monitor, so that no values are written to the reports before the benchmark starts or ends.
+ * 
+ */
+public abstract class AbstractMonitor implements Monitor {
+   protected boolean shouldRun = false;
```
rvansa: IIUC the start/stop methods are invoked from main thread, so this should be `volatile`.

extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStatisticsManager.java
rvansa: @alanfx Then, what does `startStats()` do when `executor` is null?

alanfx: @rvansa I see what you're saying now, sorry. Now `BackgroundStatisticsManager.stopStats()` does not set `executor` to `null`, so it is still available to schedule tasks in `BackgroundStatisticsManager.startStats()`.

rvansa: @alanfx I am afraid that since you shutdown the executor here, task scheduled in `startStats` will be rejected, though. 

Maybe a test case would get this correct?

alanfx: @rvansa fair point, I'll see what I can find out

alanfx: @rvansa The [JavaDocs](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html) for `ExecutorService` are pretty clear that shutting down prevents new tasks from being submitted, so I updated the code to re-initialize `executor` in `BackgroundStatisticsManager.stopStats()`. Thanks for catching this.

rvansa:
Integrated, thanks.

PR #545, opened 2018-05-22
Coordinated omission using sync mode

diegolovison:
**I am asking for a review in the following topics:**

1.  Testing modes
- When `opsPerSec` is equals to 0. It means that the current behavior won't change. We will continue having the maximum throughput [1a]
- When `opsPerSec` is greater than 0. This mode is for benchmarking response time under load. [1b]

2. Notion of schedule
- Basically, the thread will sleep. In this case if we set `opsPerSec` equals 1, only one request per second will be generated.

[1a]
Throughput: latency == service time, response/wait time not recorded. Maximum throughput is an important test mode for flushing out CPU bottlenecks and contention. It may help in exposing IO configuration issues as well. It is not however a good predictor of response time distribution as no attempt is made to estimate independent request timing (i.e uncoordinated measurement). The maximum throughput number is valuable to batch processing etc, but I'd caution against using it for the sizing of a responsive system. If a system has responsiveness SLAs it is better to use the 'fixed' test mode and run for significant periods to determine the load under which response times SLA can be met.

[1b]
Fixed  ("-rate fixed=1000/s"): latency == response time, service/response/wait time recorded. This mode is for benchmarking response time under load. The partitions/second parameter is used to determine the fixed rate scheduled of the requests. The hdr log can be used to visualize latencies over time or aggregate latencies for any segment down to the logging frequency. The logs contain response/wait/ service components that can be extracted and handled separately.

All the information comes from: https://psy-lob-saw.blogspot.com.br/2016/07/fixing-co-in-cstress.html

### DIFF
--- a/core/src/main/java/org/radargun/stages/test/Stressor.java
+++ b/core/src/main/java/org/radargun/stages/test/Stressor.java
@@ -3,6 +3,8 @@ package org.radargun.stages.test;
 import java.util.Random;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.LockSupport;
 
 import org.radargun.Operation;
 import org.radargun.logging.Log;
@@ -11,6 +13,7 @@ import org.radargun.stats.Request;
 import org.radargun.stats.RequestSet;
 import org.radargun.stats.Statistics;
 import org.radargun.traits.Transactional;
+import org.radargun.utils.TimeService;
 
 /**
  * Each stressor operates according to its {@link OperationLogic logic} - the instance is private to each thread.
@@ -31,7 +34,7 @@ public class Stressor extends Thread {
    private final OperationSelector operationSelector;
    private final Completion completion;
    private final boolean logTransactionExceptions;
-   private long delayBetweenRequests;
+   private long thinkTime;
 
    private boolean useTransactions;
    private int txRemainingOperations = 0;
@@ -40,8 +43,9 @@ public class Stressor extends Thread {
    private Statistics stats;
    private boolean started = false;
    private CountDownLatch threadCountDown;
+   private UniformRateLimiter uniformRateLimiter;
 
-   public Stressor(TestStage stage, OperationLogic logic, int globalThreadIndex, int threadIndex, boolean logTransactionExceptions, CountDownLatch threadCountDown, long delayBetweenRequests) {
+   public Stressor(TestStage stage, OperationLogic logic, int globalThreadIndex, int threadIndex, CountDownLatch threadCountDown) {
       super("Stressor-" + threadIndex);
       this.stage = stage;
       this.threadIndex = threadIndex;
@@ -50,9 +54,10 @@ public class Stressor extends Thread {
       this.random = ThreadLocalRandom.current();
       this.completion = stage.getCompletion();
       this.operationSelector = stage.getOperationSelector();
-      this.logTransactionExceptions = logTransactionExceptions;
+      this.logTransactionExceptions = stage.logTransactionExceptions;
       this.threadCountDown = threadCountDown;
-      this.delayBetweenRequests = delayBetweenRequests;
+      this.thinkTime = stage.thinkTime;
+      this.uniformRateLimiter = new UniformRateLimiter(TimeUnit.MILLISECONDS.toNanos(stage.cycleTime));
    }
 
    private boolean recording() {
@@ -100,8 +105,8 @@ public class Stressor extends Thread {
                Operation operation = operationSelector.next(random);
                try {
                   logic.run(operation);
-                  if (delayBetweenRequests > 0)
-                    sleep(delayBetweenRequests);
+                  if (thinkTime > 0)
+                    sleep(thinkTime);
                } catch (OperationLogic.RequestException e) {
                   // the exception was already logged in makeRequest
                } catch (InterruptedException e) {
@@ -112,6 +117,7 @@ public class Stressor extends Thread {
          }
 
          stats.begin();
+         uniformRateLimiter.start();
          this.started = true;
          completion.start();
          int i = 0;
@@ -120,8 +126,8 @@ public class Stressor extends Thread {
             if (!completion.moreToRun()) break;
             try {
                logic.run(operation);
-               if (delayBetweenRequests > 0)
-                  sleep(delayBetweenRequests);
+               if (thinkTime > 0)
+                  sleep(thinkTime);
             } catch (OperationLogic.RequestException e) {
                // the exception was already logged in makeRequest
             } catch (InterruptedException e) {
@@ -166,7 +172,7 @@ public class Stressor extends Thread {
 
       T result = null;
       Exception exception = null;
-      Request request = recording() ? stats.startRequest() : null;
+      Request request = uniformRateLimiter.next(recording() ? stats.startRequest(uniformRateLimiter.acquire()) : null);
       try {
          result = invocation.invoke();
          succeeded(request, invocation.operation());
@@ -307,6 +313,40 @@ public class Stressor extends Thread {
 
       public Request getRequest() {
          return request;
+      }
+   }
+
+   static class UniformRateLimiter {
+      long start = Long.MIN_VALUE;
+      final long opsPerNano;
+      long opIndex = 0;
+
+      UniformRateLimiter(long opsPerNano) {
+         this.opsPerNano = opsPerNano;
+      }
+
+      void start() {
+         start = TimeService.nanoTime();
+      }
+
+      Request next(Request request) {
+         if (opsPerNano > 0 && request != null) {
+            long intendedTime = request.getRequestStartTime();
+            long now;
+            while ((now = System.nanoTime()) < intendedTime)
+               LockSupport.parkNanos(intendedTime - now);
+         }
+         return request;
+      }
+
+      /* @return intended start time in ns for the operation */
+      long acquire() {
+         if (opsPerNano > 0) {
+            long currOpIndex = opIndex++;
+            return start + currOpIndex * opsPerNano;
+         } else {
+            return 0;
+         }
       }
    }
 }
--- a/core/src/main/java/org/radargun/stages/test/TestStage.java
+++ b/core/src/main/java/org/radargun/stages/test/TestStage.java
@@ -61,11 +61,14 @@ public abstract class TestStage extends BaseTestStage {
    @Property(doc = "Delay to let all threads start executing operations. Default is 0.", converter = TimeConverter.class)
    public long rampUp = 0;
 
-   @Property(converter = TimeConverter.class, doc = "Time between consecutive requests of one stressor thread. Default is 0.")
-   protected long delayBetweenRequests = 0;
+   @Property(converter = TimeConverter.class, doc = "Time between consecutive requests of one stressor thread. Default is 0.", deprecatedName = "delayBetweenRequests")
+   protected long thinkTime = 0;
 
    @Property(doc = "Whether an error from transaction commit/rollback should be logged as error. Default is true.")
    public boolean logTransactionExceptions = true;
+
+   @Property(converter = TimeConverter.class, doc = "Intended time between each request. Default is 0.")
+   protected long cycleTime = 0;
 
    @InjectTrait
    protected Transactional transactional;
@@ -90,6 +93,7 @@ public abstract class TestStage extends BaseTestStage {
       if (totalThreads > 0 && numThreadsPerNode > 0)
          throw new IllegalStateException("You have to set only one ot total-threads, num-threads-per-node");
       if (totalThreads < 0 || numThreadsPerNode < 0) throw new IllegalStateException("Number of threads can't be < 0");
+      if (cycleTime > 0 && thinkTime > 0) throw new IllegalStateException("We cannot mix cycleTime and thinkTime");
    }
 
    public DistStageAck executeOnSlave() {
@@ -252,7 +256,7 @@ public abstract class TestStage extends BaseTestStage {
 
       List<Stressor> stressors = new ArrayList<>();
       for (int threadIndex = stressors.size(); threadIndex < myNumThreads; threadIndex++) {
-         Stressor stressor = new Stressor(this, getLogic(), myFirstThread + threadIndex, threadIndex, logTransactionExceptions, threadCountDown, delayBetweenRequests);
+         Stressor stressor = new Stressor(this, getLogic(), myFirstThread + threadIndex, threadIndex, threadCountDown);
          stressors.add(stressor);
          stressor.start();
       }
--- a/core/src/main/java/org/radargun/stats/Request.java
+++ b/core/src/main/java/org/radargun/stats/Request.java
@@ -22,8 +22,12 @@ public final class Request {
    private boolean successful = true;
 
    public Request(Statistics statistics) {
+      this(statistics, TimeService.nanoTime());
+   }
+
+   public Request(Statistics statistics, long requestStartTime) {
       this.statistics = statistics;
-      this.requestStartTime = TimeService.nanoTime();
+      this.requestStartTime = requestStartTime;
    }
 
    public void exec(Operation operation, Runnable runnable) {
@@ -62,7 +66,7 @@ public final class Request {
          successful = false;
          throw t;
       } finally {
-         statistics.record(this, operation);
+         record(operation);
       }
       return value;
    }
@@ -73,14 +77,12 @@ public final class Request {
 //   public void responseStarted() {}
 
    public void succeeded(Operation operation) {
-      this.responseCompleteTime = TimeService.nanoTime();
-      statistics.record(this, operation);
+      record(operation);
    }
 
    public void failed(Operation operation) {
-      this.responseCompleteTime = TimeService.nanoTime();
       this.successful = false;
-      statistics.record(this, operation);
+      record(operation);
    }
 
    public void discard() {
@@ -108,5 +110,10 @@ public final class Request {
     */
    public long duration() {
       return responseCompleteTime - requestStartTime;
+   }
+
+   private void record(Operation operation) {
+      this.responseCompleteTime = TimeService.nanoTime();
+      statistics.record(this, operation);
    }
 }
--- a/core/src/main/java/org/radargun/stats/Statistics.java
+++ b/core/src/main/java/org/radargun/stats/Statistics.java
@@ -74,6 +74,10 @@ public interface Statistics extends Serializable {
       return new Request(this);
    }
 
+   default Request startRequest(long requestStartTime) {
+      return new Request(this, requestStartTime);
+   }
+
    /**
     * Create an object for tracking non-rpc-like operations.
     */
--- a/core/src/main/java/org/radargun/stats/Message.java
+++ b/core/src/main/java/org/radargun/stats/Message.java
@@ -46,4 +46,8 @@ public final class Message {
    public long totalTime() {
       return TimeUnit.MILLISECONDS.toNanos(receiveCompleteTime - sendStartTime);
    }
+
+   public long getReceiveCompleteTime() {
+      return receiveCompleteTime;
+   }
 }
--- a/core/src/main/java/org/radargun/stats/RequestSet.java
+++ b/core/src/main/java/org/radargun/stats/RequestSet.java
@@ -58,4 +58,8 @@ public final class RequestSet {
    public boolean isSuccessful() {
       return successful;
    }
+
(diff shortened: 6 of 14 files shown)


core/src/main/java/org/radargun/stats/Request.java
```diff
@@ -109,4 +111,12 @@ public long getResponseCompleteTime() {
    public long duration() {
       return responseCompleteTime - requestStartTime;
    }
+
+   public long getIntendedTime() {
+      return intendedTime;
+   }
+
+   public void setIntendedStartNs(long intendedTime) {
```
rvansa: let's match getter and setter name (if using `getFoo()`/`setFoo(long f)` style and not just `long foo()`/`void foo(long f)`)

core/src/main/java/org/radargun/stats/BasicStatistics.java
```diff
@@ -28,6 +28,9 @@
    @Property(name = "operationStats", doc = "Operation statistics prototype.", complexConverter = OperationStats.Converter.class)
    protected OperationStats prototype = new BasicOperationStats();
 
+   @Property(doc = "Operation per second.")
+   public int opsPerSec = 0;
```
rvansa: I don't think this should be configured in statistics; it should be a property of the benchmark stage and this class would have just the setter.
Ideally, `opsPerSec` would be computed in the stage based on the benchmark setup (which usually sets per-operation rate or ratio).

core/src/main/java/org/radargun/stats/IntervalStatistics.java
```diff
@@ -53,6 +57,16 @@ public void merge(Statistics otherStats) {
       endTime = Math.max(otherStats.getEnd(), endTime);
    }
 
+   // TODO I would like to understand why some stages are calling startRequest without calling the begin first
```
rvansa: begin() should be called only when the warm-up/ramp-up phase finishes, just before we enter the steady-state. We switch to steady-state usually just flipping a volatile boolean and we want to record requests that *finished* within the steady-state; that's why the Requests are created even during warm-up.
Alternative would be recording all requests that *started* within steady-state but that requires some ramp-down waiting for all the in-progress requests to finish.

core/src/main/java/org/radargun/stats/Statistics.java
```diff
@@ -173,9 +175,59 @@ default Statistics with(Statistics otherStats) {
...
    class Converter extends ReflexiveConverters.ObjectConverter {
       public Converter() {
          super(Statistics.class);
       }
    }
+
+   class UniformRateLimiter {
+      long start = Long.MIN_VALUE;
+      final long intervalNs;
+      final AtomicLong opIndex = new AtomicLong();
```
rvansa: Besides `SynchronizedStatistics` all statistics classes are intentionally single-threaded - no need to use AtomicLong.

core/src/main/java/org/radargun/stats/Statistics.java
```diff
@@ -173,9 +175,59 @@ default Statistics with(Statistics otherStats) {
...
+
+      UniformRateLimiter(int opsPerSec) {
+         intervalNs = 1000000000 / opsPerSec;
+      }
+
+      /* @return intended start time in ns for the operation */
+      long acquire() {
+         long currOpIndex = opIndex.getAndAdd(1);
+         return start + currOpIndex * intervalNs;
```
rvansa: As `intervalNs` is be rounded down the time the operation was supposed to start will slowly drift down. You need to use `start + currOpIndex * 1000_000_000 / opsPerSec` to multiply first.

core/src/main/java/org/radargun/stats/Statistics.java
```diff
@@ -173,9 +175,59 @@ default Statistics with(Statistics otherStats) {
...
+
+      Request nextRequest(Statistics statistics) {
+
+         Request req = new Request(statistics);
+
+         if (rateLimiter != null) {
+            long intendedTime = rateLimiter.acquire();
+            req.setIntendedStartNs(intendedTime);
+            long now;
+            while ((now = System.nanoTime()) < intendedTime)
```
rvansa: The purpose of `Statistics` is recording data. While I would be ok with adding the intended start to the request, `nextRequest` must not block the invocation - move this to the benchmark stage.

It might be better to move the `StreamOfRequest` stuff to benchmark stages completely, not sure ATM.

core/src/main/java/org/radargun/stats/Statistics.java
```diff
@@ -173,9 +175,59 @@ default Statistics with(Statistics otherStats) {
...
+      }
+
+      /* @return intended start time in ns for the operation */
+      long acquire() {
+         long currOpIndex = opIndex.getAndAdd(1);
+         return start + currOpIndex * intervalNs;
+      }
+
+      void start() {
+         start = System.nanoTime();
```
rvansa: Use `TimeService.nanoTime()` instead (that allows mocking in tests).

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
```diff
@@ -33,24 +23,15 @@
    @Property(doc = "Number of significant digits. Default is 2.")
    private int digits = 2;
 
-   private AbstractHistogram histogram;
-   private SoftReference<AbstractHistogram> soft;
-   private long errors = 0;
-   private Histogram compacted;
+   private Timer timer;
```
rvansa: Why are you replacing this with another instance rather than embedding the three histograms directly?
I suppose this is one of the 'gaps' you mentioned.

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
```diff
@@ -33,24 +23,15 @@
    @Property(doc = "Number of significant digits. Default is 2.")
    private int digits = 2;
 
-   private AbstractHistogram histogram;
-   private SoftReference<AbstractHistogram> soft;
```
rvansa: Fyi, the soft reference stuff et all is here because we had problems when the whole histogram was moved to master - as several slaves sent the histograms for hundreds of threads the master blew up running out of memory.

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
diegolovison: Yes. It is a gap yet.

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
diegolovison: Thanks for letting me know the reason for SoftReference. I will address this later.

diegolovison:
The PR is ready for review.
The gaps mentioned https://github.com/radargun/radargun/pull/545#issue-189753587 are the same.

I decided to put all the logic in the Stressor. In this way, the dev that will implement something don't need to worry about the `'compensation of results`

The short path that I found to send `intendedTime` between objects were:
https://github.com/radargun/radargun/pull/545/files#diff-0af01b139d10eb9062e807f68d1a09b6R117

diegolovison:
Performance tests run successfully. Link to the results http://perfrepo.mw.lab.eng.bos.redhat.com/reports/tableComparisonReport/10712

Additional info:
Branch: https://github.com/diegolovison/infinispan/tree/ISPN-5451_data_container_striping
Commit: 6d5311a
Build number: #250
Comment body: run performance tests please\r\nlibrary repl reads

-----

Why we have better response time and worst throughput?
I would like to have this comment here because maybe there is something that we reproduce later to test against this PR

rvansa:
I haven't checked the code if it's actually possible, but my guess is that response time mean does not lie, but the throughput it just # of ops/ duration of the test. So if the duration was substantially longer than expected (one thread got stuck and delayed everything), we could have observed good response times but the throughput would drop. Needs to be verified.

diegolovison:
It is ready for review:
**Notes**
1. There is one missing implementation: https://github.com/radargun/radargun/pull/545/files#diff-7efe75c97e0c7c4a741155b99cba18b8R79
1. Maybe a better name for the PR and for the commit: Compensation ??
1. There are TODO in the source code that I need your opinion. It could be address now or later in another PR ( I prefer now, if possible ).

rvansa:
@diegolovison WIthout being able to merge histograms from different threads you can't produce report (with histograms/percentiles). Have you tested that?

Not sure which TODO you mean... name for `StreamOfRequest`?

core/src/main/java/org/radargun/Operation.java
```diff
@@ -101,6 +103,14 @@ public Operation derive(String variant) {
       return operation;
    }
 
+   public long getIntendedTime() {
```
rvansa: Ouch, there should be only a single instance of this class for each `name` :-/ It works like an enum, and for local uses it should support reference-equality comparison. And yes, I haven't written this in the javadoc...

core/src/main/java/org/radargun/stages/test/Stressor.java
```diff
@@ -309,4 +314,54 @@ public Request getRequest() {
...
+         return start + currOpIndex * 1_000_000_000 / opsPerSec;
+      }
+
+      void start() {
+         start = TimeService.nanoTime();
+      }
+   }
+
+   // TODO maybe a better name ?
+   static class StreamOfOperation {
```
rvansa: Why are this and `UniformRateLimiter` separate classes? You could easily inline the functionality.

core/src/main/java/org/radargun/stages/test/TestStage.java
```diff
@@ -67,6 +67,9 @@
    @Property(doc = "Whether an error from transaction commit/rollback should be logged as error. Default is true.")
    public boolean logTransactionExceptions = true;
 
+   @Property(doc = "Operation per second.")
+   private int opsPerSec = 0;
```
rvansa: User should not calculate this by hand; you already have duration, thread number and delayBetweenRequests (which is incorrectly named as it's a cycle-time - period between starting requests)

core/src/main/java/org/radargun/stats/Statistics.java
```diff
@@ -51,6 +51,12 @@
     * Mark this moment as start of the measurement.
     * No operations should be recorded before this call.
     */
+   /**
+    * TODO
+    * We are using TimeService.currentTimeMillis() or TimeService.nanoTime()
```
rvansa: Begin and end are wall clock times that should tell you when was the test running/when were the statistics collected.
Note that `nanoTime()` is unsuitable for measuring extended periods of time as it can skew with CPU throttling and such.

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
```diff
@@ -250,4 +249,62 @@ private void readObject(ObjectInputStream s) throws IOException, ClassNotFoundEx
       }
       compacted = new Histogram(ranges, counts);
    }
+
+   public final class InternalHistogram {
```
rvansa: I still don't see why this is not inlined to `HistogramOperationStats` class.

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
```diff
@@ -250,4 +249,62 @@ private void readObject(ObjectInputStream s) throws IOException, ClassNotFoundEx
...
+         serviceTime = new org.HdrHistogram.Histogram(maxValue, digits);
+         waitTime = new org.HdrHistogram.Histogram(maxValue, digits);
+      }
+
+      public void recordValue(long startTimeNs, long endTimeNs, long intendedTimeNs) {
+         serviceTime.recordValue(endTimeNs - startTimeNs);
+
+         if (intendedTimeNs != 0) { // if intended time is set we can compute response/wait
+            responseTime.recordValue(endTimeNs - intendedTimeNs);
+            waitTime.recordValue(startTimeNs - intendedTimeNs);
```
rvansa: waitTime is never presented, and by definition it's just difference between the other two histograms.

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
```diff
@@ -250,4 +249,62 @@ private void readObject(ObjectInputStream s) throws IOException, ClassNotFoundEx
...
+         }
+      }
+
+      private org.HdrHistogram.Histogram getLatencyHistogram() {
+         if (responseTime.getTotalCount() == 0)
```
rvansa: If you always present just one of the histograms there's no reason to occupy memory by collecting the other histogram.
It seems to me that you could completely drop `intendedTime` and fill `Request.requestStart` with the value you pass to `indendedTime`, if the test is configured for such compensation, and that would save you the hassle with rewriting statistics.

rvansa:
I hope that you've already noticed that I am a pesky reviewer and it won't discourage you :)

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
diegolovison: I don't believe that we can drop `intendedTime`.
The missing part in RadarGun according with https://psy-lob-saw.blogspot.com.br/2016/07/fixing-co-in-cstress.html is the notion of schedule.
For example:
intendedTime = 10:30
requestStartTime = 10:32
responseCompleteTime = 10:33
Per my understanding, we will have 3 minutes stored in InternalHistogram.responseTime instead of 1 minute.

Well, this was what I understood.

core/src/main/java/org/radargun/Operation.java
diegolovison: Well, I will wait an answer https://github.com/radargun/radargun/pull/545#discussion_r191424876 before go and fix this

core/src/main/java/org/radargun/stages/test/TestStage.java
diegolovison: My intention here with `opsPerSec` is:
If it is 0, RadarGun will have the same behavior that we have today.
If it is greater than 0, RadarGun will start reporting in the response histogram. For sure, who will implement the benchmark file will need to start doing the formula that you mentioned by hand.

For the next PR, I was thinking about remove `thread` and `delayBetweenRequests` and just keep one thread using `CompletableFuture`.

Now that RadarGun have the notion of schedule ( intention of this PR), we know when the request should be started and when it was completed. This still pending about an answer in https://github.com/radargun/radargun/pull/545#discussion_r191424876

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
rvansa: If you never *use* requestStartTime, you don't need to record it at all, and I suggest using this field to store `intededTIme`.

diegolovison: At this time, I would like to have both because I would like to have a chance to compare the values if we found something weird.
`intendedTime` for `responseTime`
`requestStartTime` for `serviceTime`

If this is not needed, I can remove `serviceTime` and `requestStartTime`

core/src/main/java/org/radargun/Operation.java
diegolovison: What should be the best place for `intendedTime` ?

core/src/main/java/org/radargun/stages/test/TestStage.java
diegolovison: Now we have only one histogram
```java
      if (intendedTimeNs != 0) { // if intended time is set we can compute response/wait
         histogram.recordValue(endTimeNs - intendedTimeNs);
      } else {
         histogram.recordValue(endTimeNs - startTimeNs);
      }
```

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
```diff
@@ -96,15 +96,17 @@ public void record(Request request) {
    @Override
    public void record(Message message) {
       if (message.isValid()) {
-         histogram.recordValue(message.totalTime());
+         // TODO need to investigate what should be the value in intendedTimesNs
```
diegolovison: What should be the value for `intendedTimeNs` ?

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
```diff
@@ -96,15 +96,17 @@ public void record(Request request) {
...
       } else {
          errors++;
       }
    }
 
    @Override
    public void record(RequestSet requestSet) {
-      histogram.recordValue(requestSet.sumDurations());
+      // TODO need to investigate what should be the value in intendedTimesNs
+      recordValue(requestSet.getBegin(), requestSet.getEnd(), 0L);
```
diegolovison: What should be the value for `intendedTimeNs` ?

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
```diff
@@ -250,4 +252,12 @@ private void readObject(ObjectInputStream s) throws IOException, ClassNotFoundEx
       }
       compacted = new Histogram(ranges, counts);
    }
+
+   private void recordValue(long startTimeNs, long endTimeNs, long intendedTimeNs) {
+      if (intendedTimeNs != 0) { // if intended time is set we can compute response/wait
+         histogram.recordValue(endTimeNs - intendedTimeNs);
```
diegolovison: If intendedTimeNs is equals 5 and endTimeNs is equals 4.
What should we do in this case?

diegolovison:
Hi @rvansa 

Don't worry. I appreciate all your reviews.

I just pushed more source code that resulted in more questions :)

Thanks

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
rvansa: It's up to stage to set `sendStartTime` to the time when it intended to send the message instead of the time when it actually sent it.

extensions/hdrhistogram/src/main/java/org/radargun/stats/HistogramOperationStats.java
rvansa: How could the response be received before you intended to send it? If you intend to send the message later, you simply don't send it yet (regrettably it's not as easy when you're already past the time :)), so if this happens it's a bug.

core/src/main/java/org/radargun/Operation.java
diegolovison: I was looking for usage in `stats.startRequest()` and found in Stage/Logic/Stressor
What should we do in https://github.com/radargun/radargun/blob/aa57b7de437c360f5e863c65eb7a9c6b2cffa022/extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStressorLogic.java#L129 ?

rvansa: Just add a `startRequest(long intendedTime)` overload.

diegolovison:
Moved intendedTime to Request. 
See: https://github.com/radargun/radargun/pull/545/commits/8ee57cba4e924601753137450fc46212f7c964bc

Let me know what do you think

core/src/main/java/org/radargun/stages/test/TestStage.java
rvansa: So is there any reason to calculate this by hand? If you want just report intended start time instead of actual start time (ok for me), this could be a boolean property.

core/src/main/java/org/radargun/stats/Request.java
```diff
@@ -21,9 +21,16 @@
    private long responseCompleteTime = Long.MIN_VALUE;
    private boolean successful = true;
 
+   private final long intendedTime;
```
rvansa: So I still don't see why do you need to store `intendedTime` as a separate field, and then have a special handling in HistogramOperationStats. I would just let stage provide this 'fake' time and keep statistics as-is.

rvansa:
@diegolovison Sorry, some of my comments were left unaddressed so I didn't know it's up to me now.

rvansa:
Not sure if you wanted anything else but looks good to me now. Thanks for all the patience :)

mgencur:
Nice! @diegolovison

diegolovison:
Thank you @rvansa for your patience :)

alanfx:
Woo hoo! Thank you @diegolovison and @rvansa!

Issue #551, opened 2018-06-13
JGroupsProbeStage has duplicated code JGroups::Probe


diegolovison:
We can decouple Probe::main method in a new class called ProbeBuilder.
In this case, JGroupsProbeStage can delegate a call to Probe instead of have duplicated code

PR #552, opened 2018-06-13
Probe monitor

diegolovison:
Probe monitor

### DIFF
--- a/core/src/main/java/org/radargun/stages/AbstractJGroupsProbeStage.java
+++ b/core/src/main/java/org/radargun/stages/AbstractJGroupsProbeStage.java
@@ -7,18 +7,23 @@ import java.net.MulticastSocket;
 import java.util.ArrayList;
 import java.util.List;
 
-import org.radargun.DistStageAck;
 import org.radargun.config.DefinitionElement;
 import org.radargun.config.Property;
 import org.radargun.utils.ReflexiveConverters;
 import org.radargun.utils.TimeConverter;
+import org.radargun.utils.TimeService;
 
 /**
  * Inspired by org.jgroups.tests.Probe.
  *
+ * TODO https://github.com/radargun/radargun/issues/551
+ *
  * @author Matej Cimbora
+ * @author Diego Lovison &lt;dlovison@redhat.com&gt;
  */
 public abstract class AbstractJGroupsProbeStage extends AbstractDistStage {
+
+   private static final String[] EMPTY_RESPONSE = new String[0];
 
    @Property(doc = "List of queries to be performed.", optional = false, complexConverter = JGroupsProbeStage.ListConverter.class)
    private List<Query> queries = new ArrayList<>();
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/radargun/stages/AbstractJGroupsProbeStage.java
```diff
@@ -0,0 +1,113 @@
...
+         InetAddress targetAddress = InetAddress.getByName(address);
+         byte[] payload = getQuery().getBytes();
+         DatagramPacket packet = new DatagramPacket(payload, 0, payload.length, targetAddress, port);
+         socket.send(packet);
+
+         // receive responses
+         if (expectedResponseCount == -1) {
+            return successfulResponse();
+         }
+         long start = System.currentTimeMillis();
```
rvansa: Use TimeService

core/src/main/java/org/radargun/stages/AbstractJGroupsProbeStage.java
```diff
@@ -0,0 +1,113 @@
...
+               String errorMessage = "Timed out waiting for query responses";
+               log.error(errorMessage);
+               return errorResponse(errorMessage);
+            }
+            byte[] buffer = new byte[1024 * 64];
+            DatagramPacket result = new DatagramPacket(buffer, buffer.length);
+            socket.receive(result);
+            received++;
+            if (printResultsAsInfo) {
+               log.infof("Received response %s from %s:%d", new String(result.getData(), 0, result.getLength()), result.getAddress(), result.getPort());
```
rvansa: Preformat the message on one place?

core/src/main/java/org/radargun/stages/monitor/JgroupsProbeMonitorStartStage.java
```diff
@@ -0,0 +1,55 @@
...
+ * Starts collecting probe statistics on each slave node.
+ * {@link ProbeSlaveMonitor}
+ *
+ * @author Diego Lovison &lt;dlovison@redhat.com&gt;
+ */
+@Stage(name="jgroups-probe-monitor-start", doc = "Starts collecting jgroups statistics locally in each slave node.")
+public class JgroupsProbeMonitorStartStage extends AbstractJGroupsProbeStage {
+
+   @Property(doc = "Group where probe will be executed")
+   private String groupName;
```
rvansa: `AbstractDistStage` already has `groups`

core/src/main/java/org/radargun/stages/monitor/JgroupsProbeMonitorStartStage.java
```diff
@@ -0,0 +1,55 @@
...
+         slaveMonitors.addMonitor(new Monitor() {
+
+            @Override
+            public void run() {
+               JgroupsProbeMonitorStartStage.this.run();
+            }
+
+            @Override
+            public void start() {
+               JgroupsProbeMonitorStartStage.this.run();
```
rvansa: Why do you execute extra run in start() and stop()?

diegolovison: I would like to have a chance to see the "probe" response before and after the test.
For example, in ISPN-9198 we don't know why the node leave the cluster.
In this way we can compare the results: before, during, after the test.
If this won't help so much I am happy to remove.

diegolovison:
1. TextEvent
We have the issue https://github.com/radargun/radargun/issues/553 This is the reason why I kept log.info or log.trace

1. Probe
I would like to remove the code into `AbstractJGroupsProbeStage::run` fixing https://github.com/radargun/radargun/issues/551

core/src/main/java/org/radargun/stages/monitor/JgroupsProbeMonitorStartStage.java
```diff
@@ -0,0 +1,78 @@
...
+@Stage(name="jgroups-probe-monitor-start", doc = "Starts collecting jgroups statistics locally in each slave node.")
+public class JgroupsProbeMonitorStartStage extends AbstractJGroupsProbeStage {
+
+   @Property(doc = "Period of statistics collection. The default is 60 seconds.", converter = TimeConverter.class)
+   private long period = 60_000;
+
+   @Override
+   public DistStageAck executeOnSlave() {
+
+      if (groups == null || (slaveState.getGroupName() != null && groups.contains(slaveState.getGroupName()))) {
```
rvansa: This condition (and more) is already covered in `AbstractDistStage.shouldExecute()`

core/src/main/java/org/radargun/stages/monitor/JgroupsProbeMonitorStartStage.java
```diff
@@ -0,0 +1,78 @@
...
+         ProbeSlaveMonitor slaveMonitors = slaveState.get(ProbeSlaveMonitor.MONITORS) == null ? new ProbeSlaveMonitor(slaveState, period)
+            : (ProbeSlaveMonitor) slaveState.get(ProbeSlaveMonitor.MONITORS);
+         slaveMonitors.addMonitor(new Monitor() {
+
+            @Override
+            public void run() {
+               try {
+                  monitor(JgroupsProbeMonitorStartStage.this.run());
+               } catch (IOException e) {
```
rvansa: neither `run` nor `monitor` seem to declare `IOException`

diegolovison: I decided to throw `IOException` because `JGroupsProbeStage` needs to return `errorResponse`. See https://github.com/radargun/radargun/pull/552/files#diff-d5a5226c6f119fd7bcbc7f9a3980afeaR31

I decided to call the method `run`. Maybe we can change to another one to avoid confusion. What do you think?

diegolovison:
Closing because I would like to integrate this PR into master instead of regression2

PR #560, opened 2018-07-30
Etcd fixes and avoid cache manager exception

diegolovison:
For etcd, allow String and byte[] as a cache key


alanfx:
Looks good to me, @diegolovison. Merging

PR #561, opened 2018-08-02
Remove default rest easy http headers and allow custom http headers

diegolovison:
Remove default rest easy http headers and allow custom http headers


rvansa:
LGTM, integrated, thanks.

PR #567, opened 2018-08-30
Improve HTML report and statistics

diegolovison:
**Overall Operation - e2d32dd**

The following changes were applied in the HTML report sections

The following sections were removed:
- Operation: BasicOperations.Total - It was showing only information for non-tx cache
- Operation: Transactional.Duration - It was showing only information for tx cache
- Operation: BasicOperations.Total.TX - Each metric in the chart was ( ( Transactional.Begin + BasicOperation + Transactional.Commit + Transactional.Duration ) / Total Operation ) This can be misinterpreted. 

The following section was added:
- Operation: OverallOperation.Duration -In this section, we can now compare a tx and non-tx cache. The statistics are now accurate.

Here is an example:
[Test writes10 - pr.pdf](https://github.com/radargun/radargun/files/2337558/Test.writes10.-.pr.pdf)

---

**Allow different envs for each test stage - eaf1396**
I would like to keep only one benchmark file with different cycle-time configuration.
Here is the xml example:
```xml
<benchmark xmlns="urn:radargun:benchmark:3.0">
   <master bindAddress="${master.address:127.0.0.1}" port="${master.port:2103}"/>
   <clusters>
      <scale from="4" to="4">
         <cluster />
      </scale>
   </clusters>
   <configurations>

      <config name="dist">
         <setup plugin="${env.PLUGINNAME}">
            <vm-args>
              <properties>
                  <property name="foo" value="bar" />
              </properties>
            </vm-args>
            <environment>
               <var name="CYCLE_TIME_10" value="5l" />
               <var name="CYCLE_TIME_100" value="30l" />
            </environment>
         </setup>
      </config>

     <config name="dist-tx">
         <setup plugin="${env.PLUGINNAME}">
            <vm-args>
              <properties>
                  <property name="foo" value="bar" />
              </properties>
            </vm-args>
            <environment>
               <var name="CYCLE_TIME_10" value="5l" />
               <var name="CYCLE_TIME_100" value="30l" />
            </environment>
         </setup>
      </config>
   </configurations>
   <init/>
   <rg:scenario xmlns:rg="urn:radargun:benchmark:3.0"
                xmlns="urn:radargun:stages:core:3.0"
                xmlns:cache="urn:radargun:stages:cache:3.0"
                xmlns:l="urn:radargun:stages:legacy:3.0">
               
      <l:basic-operations-test test-name="writes10" duration="2m" num-threads-per-node="10" entry-size="${entry.size}" get-ratio="0" put-ratio="1" cycle-time="${CYCLE_TIME_10}">
         <l:key-selector>
            <l:concurrent-keys total-entries="${num.entries}" />
         </l:key-selector>
         <statistics>
            <common/>
         </statistics>
      </l:basic-operations-test>

      <l:basic-operations-test test-name="writes100" duration="2m" num-threads-per-node="100" entry-size="${entry.size}" get-ratio="0" put-ratio="1" cycle-time="${CYCLE_TIME_100}">
         <l:key-selector>
            <l:concurrent-keys total-entries="${num.entries}" />
         </l:key-selector>
         <statistics>
            <common/>
         </statistics>
      </l:basic-operations-test>

   </rg:scenario>

</benchmark>
```

---

I believe that the other commits are explained by their commit message but if you have any question please let me know.


core/src/main/java/org/radargun/stages/test/Stressor.java
```diff
@@ -157,13 +158,15 @@ private void runInternal() {
    }
 
    public <T> T makeRequest(Invocation<T> invocation, boolean countForTx) throws OperationLogic.RequestException {
+
+      if (recording()) {
+         requests = stats.requestSet();
```
rvansa: The requestSet should be created only when we're about to start a transaction - when there's ongoing transaction with multiple writes it shouldn't do it on each request.

core/src/main/java/org/radargun/Master.java
```diff
@@ -218,6 +218,11 @@ private int executeStage(Configuration configuration, Cluster cluster, int stage
       }
       extras.put(Properties.PROPERTY_SLAVE_INDEX, "-1");
       extras.put(Properties.PROPERTY_PROCESS_ID, String.valueOf(Utils.getProcessID()));
+      configuration.getSetups().forEach(setup ->
```
rvansa: I don't understand this change: envs should be referenced using `${env.MY_VAR}`

extensions/hdrhistogram/src/main/java/org/radargun/stats/CommonStatistics.java
```diff
@@ -23,7 +23,7 @@
    private int digits = 2;
 
    public CommonStatistics() {
-      super(new Statistics[] {new BasicStatistics(), new PeriodicStatistics(), new BasicStatistics(new HistogramOperationStats())});
+      super(new Statistics[] {new PeriodicStatistics(), new BasicStatistics(new HistogramOperationStats())});
```
rvansa: You won't get precise values for mean or stddev with histogram stats; buckets introduce imprecision. Ofc it's questionable whether you need more than 2 digits (the default histogram precision)....

core/src/main/java/org/radargun/stages/test/Stressor.java
diegolovison: Removed the commit

core/src/main/java/org/radargun/Master.java
diegolovison: Removed the commit.
We can do the following:
```xml
....
            <environment>
               <var name="CYCLE_TIME" value="1s" />
            </environment>
.....
cycle-time="${env.CYCLE_TIME:0}"
```
Master will compile and set the value equals 0
When master starts the slave, it will send 1s as an environment variable

extensions/hdrhistogram/src/main/java/org/radargun/stats/CommonStatistics.java
diegolovison: My motivation behind this commit is that I would like to work with only one data store when looking for statistical values.

Using 2 digits from histogram, 3 digits from histogram and our implementation we have the following values respectively:
5.000021402104251E8
5.000000307106532E8
4.99999999E8

This commit is not mandatory. I would like to follow your advice. 
If you believe that we could have any kind of problem in the future I am fine to revert.

PR #569, opened 2018-09-21
Hot Rod Transaction

diegolovison:
The cache configuration looks like:

```xml
                <distributed-cache name="hotrodDistTx">
                    <locking isolation="REPEATABLE_READ" />
                    <transaction mode="NON_XA" locking="PESSIMISTIC" />
                </distributed-cache>
```


plugins/infinispan4/src/main/java/org/radargun/service/InfinispanEmbeddedService.java
```diff
@@ -33,7 +33,7 @@
 import static java.util.concurrent.TimeUnit.MINUTES;
 
 @Service(doc = InfinispanEmbeddedService.SERVICE_DESCRIPTION)
-public class InfinispanEmbeddedService {
+public class InfinispanEmbeddedService implements InfinispanTransactionalService {
    protected static final String SERVICE_DESCRIPTION = "Service hosting Infinispan in embedded (library) mode.";
```
alanfx: @diegolovison I think this approach is wrong. You should add the `Transactional` trait to the implementation of the HotRodService for 9.3 or 9.4. (I can't remember which version this was added in.) @rvansa isn't that correct?

rvansa: @diegolovison did expose the `Transactional` trait but rather than creating his own implementation he reused the embedded one, and added some helpers through the InfinispanTransactionalService. I haven't undersood the changes completely but I must say that I dislike the `if (... instanceof HotRodService)` in Infinispan94Transactional. Hot Rod and Embedded implementations should not share the implemenation in RG unless the implementation in Infinispan is really common.

diegolovison: Hot Rod Transaction is supported since 9.3. We should create a new plugin `infinispan93` and move the Hot Rod transaction support into `infinispan93`

Hot Rod and Embedded Transaction is the same from the client perspective.
https://github.com/radargun/radargun/blob/387c90a53adf83b99993b3ba7b3fbfec535b836a/plugins/infinispan4/src/main/java/org/radargun/service/InfinispanTransactional.java#L46

diegolovison:
**infinispan93 - plugin**
The plugin `infinispan93` was added because Hot Rod Transactional support was finalized in ISPN 9.3

**InfinispanEmbeddedService.java**
Infinispan51EmbeddedService extends InfinispanEmbeddedService
InfinispanEmbeddedService implements InfinispanTransactionalService

The transactional support exists from the begging of Infinispan, this is why InfinispanTransactionalService implements InfinispanTransactionalService

**Infinispan93HotrodService.java**
Transaction support was introduce in 93. This is why Infinispan93HotrodService implements InfinispanTransactionalService

**InfinispanTransactional.java**
The code to handle transactional cache in the client side, it is the same for Embedded and Hot Rod mode.

This is why we created the InfinispanTransactionalService interface and the construtor of InfinispanTransactional receive the interface instead of InfinispanEmbeddedService.

The embedded mode and Hot Rod has different ways to interact with the transactional cache but they use a common patther ( methods into the interface )

**ProvidesTrait**
For Hot Rod we can:
```java
   @ProvidesTrait
   public InfinispanTransactional createTransactional() {
      return new Infinispan93Transactional(this);
   }
```
For Embedded Mode we can:
```java
   @ProvidesTrait
   @Override
   public InfinispanTransactional createTransactional() {
      return new Infinispan51Transactional(this);
   }
```

alanfx:
Closing in favor of https://github.com/radargun/radargun/pull/571

PR #571, opened 2018-09-28
Hot Rod Transaction

diegolovison:
Hot Rod Transaction


plugins/infinispan93/src/main/java/org/radargun/service/Infinispan93HotRodTransactional.java
```diff
@@ -0,0 +1,89 @@
...
+   @Override
+   public Transaction getTransaction() {
+      return new Infinispan93HotRodTransactional.Tx();
+   }
+
+   protected <T> TransactionManager getTransactionManager(T resource) {
+      if (resource instanceof HotRodOperations.HotRodCache) {
+         return ((HotRodOperations.HotRodCache) resource).noReturn.getTransactionManager();
+      } else {
+         throw new IllegalArgumentException("Invalid resource: " + String.valueOf(resource));
```
diegolovison: @alanfx we should use `resource instanceof HotRodOperations.HotRodCache` to avoid a ClassCastException.
In this case, should we return null?

alanfx: @diegolovison Can you make this happen or are you just worried it might happen? I'm trying to understand how the wrong type of cache would be passed into the HotRod service.

diegolovison: I consider the cast check as a good practice and prefer an error instead of null.
Removed.

alanfx: I feel like if this error occurs, we have bigger problems!

diegolovison: In this case, it means that the dev implemented something wrong. IMO, it is ok keep as is because `ClassCastException` or `IllegalArgumentException` will stop the code execution.

alanfx:
@diegolovison Looks good, merging...

PR #572, opened 2018-09-28
 Openshift client in PATH and remove sudo from docker

diegolovison:
`myproject` is the default project in Openshift. It will prevent removing the default.


rvansa:
I wasn't sure about removing the sudo and forcing user to grant priviledges related to `docker` group but using sudo here forces user to be in `wheel` anyway, so probably good enough.

PR #573, opened 2018-10-02
Flight Recorder for Java Process

diegolovison:
Dynamically enable Flight Recorder using an environment variable.
Here is the example:

```xml
      <config name="dist">
         <setup group="server" plugin="${env.PLUGINNAME}">
            <server>
               <flight-recorder>
                  <enabled>${env.ENABLE_SERVER_FLIGHT_RECORDER:false}</enabled>
                  <settings>/${WORKSPACE}/jdg-qe-profile.jfc</settings>
                  <filename>${WORKSPACE}/jfr_server_${slave.index}.jfr</filename>
               </flight-recorder>
            </server>
         </setup>
         <setup group="client" plugin="${env.PLUGINNAME}">
            <hotrod xmlns="urn:radargun:plugins:${env.PLUGINNAME}:3.0" cache="hotrodDist">
               <servers>127.0.0.1:11222</servers>
            </hotrod>
         </setup>
      </config>
```

### DIFF
--- a/plugins/process/src/main/java/org/radargun/service/JavaProcessService.java
+++ b/plugins/process/src/main/java/org/radargun/service/JavaProcessService.java
@@ -4,8 +4,10 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


plugins/process/src/main/java/org/radargun/service/JavaProcessService.java
```diff
@@ -152,12 +158,27 @@ public String getJavaPIDs() {
...
+
+      // the default implementation is using setArgs. I would like to avoid doing this refactor
+      if (flightRecorder.isEnabled()) {
+         List<String> flightRecorderArgs = new ArrayList<>(1);
+         flightRecorder.setArgs(flightRecorderArgs);
+         StringBuilder flightRecorderArgBuilder = new StringBuilder();
+         flightRecorderArgs.forEach(arg -> {
+            flightRecorderArgBuilder.append(arg).append(" ");
+         });
+         overwriteJavaOpts(envs, flightRecorderArgBuilder.toString());
```
rvansa: This looks like adding flight recorder overrides `javaOpts` parameter. Please add it instead (don't worry about duplicate args here). The `javaOpts` directly overwrites `JAVA_OPTS` env var because it that's just a different way to do the same thing.

rvansa:
OK, integrated, thanks.

PR #576, opened 2018-10-04
JDG73

diegolovison:
JDG73


plugins/jdg73/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+      <dependency>
+         <groupId>org.radargun</groupId>
+         <artifactId>radargun-mapreduce</artifactId>
+         <version>${project.version}</version>
+         <scope>provided</scope>
+      </dependency>
+
+      <dependency>
+         <groupId>org.radargun</groupId>
+         <artifactId>plugin-${plugins.infinispan-latest}</artifactId>
```
alanfx: @diegolovison I just saw this, but `jdg73` should depend on `jdg72`, and an infinispan plugin version. This is correct in the `jdg71` pom.xml https://github.com/radargun/radargun/blob/master/plugins/jdg71/pom.xml, but it is wrong in the `jdg72` pom file. Can you fix this?

diegolovison:
Hey @alanfx I fixed that issue. Please let me know what do you think. Thanks

PR #578, opened 2018-10-12
Update Hazelcast versions and add 3.10.x plugin

alanfx:

PR #579, opened 2018-10-24
Hazelcast 3.10, Hazelcast Enterprise 3.10, Redis Open-source and Redi…

Holmistr:
…sLabs Enterprise support

I understand that there might be quite a few code duplications. However, I'm not a Radargun expert and the plugins inheritance was really messing with me, thus I took the easy road.


plugins/hazelcast310/src/main/resources/dist-sync.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<hazelcast xsi:schemaLocation="http://www.hazelcast.com/schema/config hazelcast-config-3.10.xsd"
+           xmlns="http://www.hazelcast.com/schema/config"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+    <group>
+        <name>dev</name>
```
alanfx: Can you add a comment that explains what this username and password are used for?

plugins/hazelcast39/pom.xml
```diff
@@ -13,7 +13,7 @@
    <name>Hazelcast 3.9.x plugin for RadarGun</name>
 
    <properties>
-      <version.hazelcast>3.9.2</version.hazelcast>
+      <version.hazelcast>3.9</version.hazelcast>
```
alanfx: Why a less specific version? In https://github.com/radargun/radargun/pull/578 I set this to `3.9.4`, and made another change to depend on `plugin-hazelcast38` instead of `plugin-hazelcast37`.

plugins/hazelcast39/src/main/resources/dist-sync.xml
```diff
@@ -10,32 +10,19 @@
...
         </join>
     </network>
 
-    <executor-service>
-        <pool-size>16</pool-size>
-        <queue-capacity>0</queue-capacity>
-    </executor-service>
-
     <map name="default">
-        <backup-count>1</backup-count>
```
alanfx: Can you say why you're removing these configuration options?

plugins/redis-enterprise/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+        <groupId>org.radargun</groupId>
+        <artifactId>radargun-plugin-parent</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+
+    <name>Redis Enterprise plugin for RadarGun</name>
+
+    <artifactId>plugin-redis-enterprise</artifactId>
```
alanfx: I don't understand why you're adding a new plugin here, since it seems to be exactly the same as `plugin-redis`. Was it just to add the hack for clearing the `JedisPool`? @rvansa any thoughts on where this could be hooked in instead? I was think that you could stop the client service, but that might effect the benchmark too much.

plugins/hazelcast39/src/main/resources/dist-sync.xml
rvansa: My guess is since it's the default. We used to keep even the defaults to make obvious that this is 1:1 to our `numOwners="2"`. No strong feelings about the removal.

plugins/redis-enterprise/pom.xml
rvansa: Looking at the code, I'd rather do `<stop /><start />` than hijacking the `clear` method. Feel free to add `RestartStage` or use `<stop restart="true">` if you long for single stage.

extensions/cache/src/main/resources/benchmark-redis.xml
```diff
@@ -1,68 +1,73 @@
...
-         <group name="server" size="6" />
-         <group name="client" size="2" />
+      <cluster size="3">
+         <group name="client" size="3"/>
       </cluster>
    </clusters>
 
    <configurations>
       <config name="redis">
-         <setup group="server" plugin="redis32">
```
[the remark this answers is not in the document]
rvansa: Yep, the example benchmarks should be self-contained. I think that since @Holmistr was using some Enterprise version, he could not add that one to this configuration. I don't think we should call the plugin itself `Enterprise` if it's using community client to connect to `Enterprise` server, though, but I might misunderstood.

Holmistr:
Hi @alanfx @rvansa , I completely forgot about this one :) Apparently all my notifications got moved to some other folder. I will be reviving this and hopefully get it merged one day :)

alanfx:
@Holmistr yeah sure, I'll believe it when I see it! You could also take a look at 
https://github.com/radargun/radargun/pull/588 which is an attempt to consolidate the work you started. I may even buy you a beer! 🍺

Holmistr:
@alanfx Hehe, thanks for the trust :-D ...
One question - is it possible to add a stage just to a plugin? I mean if the product has some specific functionality and you would like to test it - example: Redis pipelining. 
I don't see how you could do it other that hacking some of the existing stages. But than you cannot compile it for other plugins. Am I missing something?

alanfx:
@Holmistr I think you'd have to add a `Trait` that only the Redis plugin supported that is required by a `Stage`. Isn't this the correct approach, @rvansa?

Holmistr:
Ok, great, so I went through the right way. Getting into the architecture is a bit difficult, but I'm glad that I understood it correctly :) Thanks

alanfx:
You're a smart guy, @Holmistr I knew you could figure it out!

rvansa:
Similar requirement came up a couple of times, but I have always decided to not promote something that cannot be abstracted into a trait.
ATM I am actually not sure what will happen now if you put some stage just into one plugin; RG3 uses flat classpath and loads the stages via reflection, so it could work. Ofc, you won't get all the fancy schemes as plugin modules are not scraped for stages.

alanfx:
I like using the `Trait`, because then if another plugin adds the ability, the `Stage` can be reused which is the point of the architecture

alanfx:
Replaced by https://github.com/radargun/radargun/pull/596 Closing...

PR #580, opened 2018-10-31
Narayana is required for Infinispan 94x

diegolovison:
After https://github.com/infinispan/infinispan/pull/6361 narayana is required when building the infinispan-plugin for the version 94x


alanfx:
OK, looks good.

PR #583, opened 2018-11-19
ISPN 10x Internal thread states exposition

diegolovison:
The commit https://github.com/infinispan/infinispan/commit/2b68c55bcb94fc42fad5e2ee3b63eef9d912dd90 removed the `delegate` field.
The goal with this PR is make RadarGun compatible with ISPN10x


parent/pom.xml
```diff
@@ -430,10 +430,11 @@
             <module>../plugins/infinispan92</module>
             <module>../plugins/infinispan93</module>
             <module>../plugins/infinispan94</module>
+            <module>../plugins/infinispan100</module>
```
alanfx: Not sure why you're picking `infinispan100` instead of `infinispan10`?

diegolovison: `infinispan50` is ISPN5.0
`infinispan100` is ISPN 10.0
I followed that pattern

alanfx: I guess for me, the plugin version never changes for a micro release, so I think `infinispan10` is better.

diegolovison: I believe that we will have 10.1, 10.2, 10.3 in a short period of time instead of 11.0, 12.0 and 13.0

alanfx: OK, you are completely correct on this! Sorry, I have having Monday morning brain. I'll merge this

PR #591, opened 2019-02-19
Java11 Support

diegolovison:


plugins/jdg66/pom.xml
```diff
@@ -166,20 +166,8 @@
...
-       </dependency>     
-       
-   </dependencies>
+       </dependency>
 
-   <dependencyManagement>
-      <dependencies>
-         <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-parent</artifactId>
```
alanfx: My intent with adding this dependency was to avoid needing the `property` settings above. This avoid us needing to maintain the versions manually.

parent/pom.xml
```diff
@@ -648,50 +648,139 @@
...
+            <jdk>1.8</jdk>
+         </activation>
+         <dependencies>
+            <!-- This dependency is not loaded in plugin classloader, we have to include
+                 that directly on classpath. The reason is that it loads native shared
+                 libraries and you can't have that loaded in multiple classloaders. -->
+            <dependency>
+               <groupId>com.sun</groupId>
+               <artifactId>tools</artifactId>
+               <version>1.7.0</version>
```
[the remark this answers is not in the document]
diegolovison: The scope is system, in this case, we can set any version.
I copied and pasted from here https://github.com/radargun/radargun/pull/591/files#diff-676a3f01f4fe2cef9b6dbc49f05d3dafL21 and decided to keep as is.

plugins/jdg66/pom.xml
```diff
@@ -83,64 +85,60 @@
...
+            <exclusion>
+               <groupId>org.jboss.logging</groupId>
+               <artifactId>jboss-logging-spi</artifactId>
+            </exclusion>
+         </exclusions>
+      </dependency>
 
-       <dependency>
-           <groupId>org.jboss.spec.javax.transaction</groupId>
-           <artifactId>jboss-transaction-api_1.1_spec</artifactId>
```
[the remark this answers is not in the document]
diegolovison: The new dependency `<artifactId>jbossjta</artifactId>` has `<artifactId>jboss-transaction-api_1.1_spec</artifactId>`

plugins/jdg66/pom.xml
diegolovison: Exactly. The main issue is that those old poms or a plugin have a dependency for `${java.home}/../lib/tools.jar`. Instead of drop JDK8 support, I decided that it is better to allow to build with JDK11 and keep JDK8 support adding a property value.

parent/pom.xml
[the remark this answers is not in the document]
alanfx: I think it makes sense to use `1.8.0` since that is the base version we'll need now. WDYT?

plugins/jdg66/pom.xml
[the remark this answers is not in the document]
alanfx: ok

plugins/jdg66/pom.xml
alanfx: Well, the `tools.jar` file is only used by the `JavaProcessService` class, and I'm still not clear why removing this dependency helps?

parent/pom.xml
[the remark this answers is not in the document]
diegolovison: Done

plugins/jdg66/pom.xml
diegolovison: Because there are plugins from `infinispan-parent` that use it.
For example: in this PR you will see that the plugins are using the latest version because of the same issue.

alanfx: Sorry, @diegolovison I am still confused by this statement. Can you just explain what you are fixing by removing the `infinispan-parent` dependency from the plugin POM file? If a plugin expects to be running in a JDK during build time, that is fine. Setting `java.home` for the `JavaProcessService` is a runtime concern, so I don't see how those two are related?

diegolovison: Plugins from `<artifactId>infinispan-parent</artifactId>` are using `tools.jar`.
With the `<artifactId>infinispan-parent</artifactId>` the build with JDK11 will fail.

diegolovison: Well, I don't tell that `JavaProcessService` is related with `<artifactId>infinispan-parent</artifactId>` :)

alanfx:
@diegolovison explained the last bits of this to me, and I have been able to build with JDK 8 and 11. Merging...

PR #592, opened 2019-02-21
Remove transactionStarted and transactionEnded

diegolovison:
According with my researches we don't need the class `Delegates.java` and we could start the transaction in https://github.com/radargun/radargun/compare/master...diegolovison:remove_delegate?expand=1#diff-9158f9969a715eda006055cd03e939d4R163

This PR is important because it is a step forward to remove the cycled dependency from `Stressor` and `OperationLogic`

Once we have removed the cycled dependency, we can work and make Stressor Stateless. After that, we can create Asynchronous traits without creating new Tests.

I am open for discussion and questions as always :)

Cheers


extensions/cache/src/main/java/org/radargun/stages/iteration/IterateStage.java
```diff
@@ -255,7 +255,12 @@ public Operation operation() {
 
       @Override
       public Operation txOperation() {
-         return Iterable.GET_ITERATOR;
+         throw new UnsupportedOperationException("There is no transactional operation");
+      }
+
+      @Override
+      public Object getTxResource() {
+         throw new UnsupportedOperationException("getTxResource is supposed to be used when we need transactional support");
```
rvansa: Inheritance/default methods are usually better than copy-paste.

core/src/main/java/org/radargun/stages/test/Stressor.java
```diff
@@ -160,7 +160,7 @@ private void runInternal() {
       if (useTransactions && txRemainingOperations <= 0) {
          try {
             ongoingTx = stage.transactional.getTransaction();
-            logic.transactionStarted();
+            ongoingTx.wrap(invocation.getTxResource());
```
rvansa: This doesn't look right; you're not using the return value. You're supposed to call the operation methods on the return value (wrapper), not on the original cache. If you start a transaction but run the method (e.g. put()) on the original cache, it won't run transactionally in Hazelcast.

diegolovison: Thanks for catching this @rvansa 
I am closing this PR

PR #595, opened 2019-03-14
Fix Unknown argument 'compress' in diagnostic for JDK11

diegolovison:


alanfx:
Looks painless. Merging!

PR #597, opened 2019-04-05
Avoid penalizing the first request

diegolovison:
Avoid penalizing the first request


alanfx:
Looks good @diegolovison Merging

PR #600, opened 2019-05-01
Add Couchbase cache plugin

MikeGoldsmith:
- based on Couchbase Java SDK 2.7.4

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -85,6 +85,8 @@
    <properties>
       <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
       <plugins.chm/>
+      <plugins.couchbase/>
+      <plugins.couchbase-latest/>
       <plugins.docker/>
       <plugins.openshift/>
       <plugins.jcache/>
@@ -303,6 +305,21 @@
          </modules>
          <properties>
             <plugins.chm>chm</plugins.chm>
+         </properties>
+      </profile>
+      <profile>
+         <id>couchbase</id>
+         <activation>
+            <property>
+               <name>!no-couchbase</name>
+            </property>
+         </activation>
+         <modules>
+            <module>../plugins/couchbase</module>
+         </modules>
+         <properties>
+            <plugins.couchbase-latest>couchbase</plugins.couchbase-latest>
+            <plugins.couchbase>${plugins.couchbase-latest}</plugins.couchbase>
          </properties>
       </profile>
       <profile>
--- a/plugins/couchbase/src/main/resources/plugin.properties
+++ b/plugins/couchbase/src/main/resources/plugin.properties
@@ -1,0 +1,3 @@
+service.default org.radargun.service.CouchbaseClientService
+service.server org.radargun.service.CouchbaseClientService
+service.couchbase-client org.radargun.service.CouchbaseClientService
--- a/docs/getting_started/building_binaries.md
+++ b/docs/getting_started/building_binaries.md
@@ -34,6 +34,7 @@ This operation might take a few minutes, as it will download the dependencies ne
     lib
     plugins(*)
         chm
+        couchbase
         docker
         ehcache25
         hazelcast2/36/37
@@ -95,6 +96,7 @@ All properties have to be provided separatelly preceeded by "-D" -> property "sk
 
 * **Enabled by default**:
     * **chm**			(parameter `-Dno-chm` to disable)
+    * **couchbase**		(parameter `-Dno-couchbase` to disable)
     * **ehcache**		(parameter `-Dno-ehcache` to disable)
     * **hazelcast**		(parameter `-Dno-hazelcast` to disable)
     * **jbosscache**		(parameter `-Dno-jbosscache` to disable)
--- a/plugins/couchbase/README.md
+++ b/plugins/couchbase/README.md
@@ -1,0 +1,21 @@
+# Couchbase Client Plugin
+
+This plugin allows Radargun to run performance tests against Couchbase server using the Couchbase Java SDK (version 2.7.4).
+
+## Creating a Cluster
+
+### Docker
+
+Couchbase server can be setup and configured using Docker, with official images available on [DockerHub](https://hub.docker.com/_/couchbase). Instructions on how to start the container are available on there too.
+
+### Manual install
+
+Couchbase server can also be installed using a direct download available from [Couchbase.com/downloads](https://www.couchbase.com/downloads). Download and install instructions are available on the website.
+
+## Setting up the Cluster
+
+Once you have a running cluster, you need to setup the cluster before it can service requests. This is done by opening the web UI at [localhost:8091](http://localhost:8091) and following create new cluster forms (only the KV service is required to run Radargun tests).
+
+## Creating the 'default' bucket
+
+By default, the plugin expects to use a bucket called _default_ to store and retrieve values from. You can create a bucket using the *bucket* tab and selecting _new bucket_.
--- a/plugins/couchbase/pom.xml
+++ b/plugins/couchbase/pom.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.radargun</groupId>
+        <artifactId>radargun-plugin-parent</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>plugin-couchbase</artifactId>
+    <name>Couchbase plugin for RadarGun</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.radargun</groupId>
+            <artifactId>radargun-cache</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.couchbase.client</groupId>
+            <artifactId>java-client</artifactId>
+            <version>2.7.4</version>
+        </dependency>
+    </dependencies>
+</project>
--- a/plugins/couchbase/src/main/java/org/radargun/service/CouchbaseClientOperations.java
+++ b/plugins/couchbase/src/main/java/org/radargun/service/CouchbaseClientOperations.java
@@ -1,0 +1,86 @@
+package org.radargun.service;
+
+import java.nio.charset.StandardCharsets;
+
+import com.couchbase.client.java.document.ByteArrayDocument;
+import org.radargun.traits.BasicOperations;
+
+public class CouchbaseClientOperations implements BasicOperations {
+
+   protected final CouchbaseClientService service;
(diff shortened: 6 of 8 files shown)


parent/pom.xml
```diff
@@ -85,6 +85,8 @@
    <properties>
       <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
       <plugins.chm/>
+      <plugins.couchbase/>
+      <plugins.couchbase-latest/>
```
daschl: why is this needed?

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseService.java
```diff
@@ -0,0 +1,45 @@
...
+
+   @Property(doc = "Password")
+   protected String password = "password";
+
+   @Property(doc = "Bucket name")
+   protected String bucketName = "default";
+
+   @Override
+   public void start() {
+      cluster = CouchbaseCluster.create(nodes);
```
daschl: I think since this is just doing sync kv ops, we can pass in the environment and set the following:

.callbacksOnIoPool(true)
.runtimeMetricsCollectorConfig(DefaultMetricsCollectorConfig.disabled())
.networkLatencyMetricsCollectorConfig(DefaultLatencyMetricsCollectorConfig.disabled())

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
```diff
@@ -0,0 +1,70 @@
...
+      @Override
+      public byte[] getAndRemove(byte[] key) {
+         byte[] doc = get(key);
+         remove(key);
+         return doc;
+      }
+
+      @Override
+      public void clear() {
+         throw new UnsupportedOperationException("Clearing Couchbase is not implemented yet.");
```
daschl: does this compare to the flush() command on the bucket manager?

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
```diff
@@ -0,0 +1,70 @@
...
+
+      @Override
+      public void put(byte[] key, byte[] value) {
+         ByteArrayDocument doc = ByteArrayDocument.create(new String(key), value);
+         service.bucket.upsert(doc);
+      }
+
+      @Override
+      public byte[] getAndPut(byte[] key, byte[] value) {
+         byte[] document = get(key);
```
daschl: I think this should use the cas on put (replace) to be more realistic.

Might make sense to have specialized impls for this to avoid creating a string key more than once

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
```diff
@@ -0,0 +1,70 @@
...
+            service.bucket.remove(new String(key));
+            return true;
+         } catch(Exception e) {
+            return false;
+         }
+      }
+
+      @Override
+      public byte[] getAndRemove(byte[] key) {
+         byte[] doc = get(key);
```
daschl: I think this should use the cas on remove to be more realistic.

Might make sense to have specialized impls for this to avoid creating a string key more than once

parent/pom.xml
MikeGoldsmith: I believe this is used to discover the plugin projects and is used by the build scripts when creating the distributable ZIP.

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
MikeGoldsmith: Yes, it does - I didn't implement as the Redis plugin provider I used as a basis didn't implement it. I'm not sure if it's used but can update to do the flush.

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
MikeGoldsmith: Sure, will look to do that.

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
MikeGoldsmith: Sure, will look to do that.

parent/pom.xml
alanfx: First of all, thank you for this contribution. This property allows you to build a RadarGun distribution that only includes the latest version of each plugin, instead of every version of the plugin.

MikeGoldsmith:
@alanfx  @diegolovison I think this is ready for initial review. Let me know if there is anything else you would like to see in this PR.

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
```diff
@@ -0,0 +1,92 @@
...
+   public CouchbaseOperations(CouchbaseService service) {
+      this.service = service;
+   }
+
+   @Override
+   public <K, V> Cache<K, V> getCache(String cacheName) {
+      return adapter;
+   }
+
+   private class CouchbaseCacheAdapter<K, V> implements BasicOperations.Cache<byte[], byte[]> {
```
diegolovison: You should use something like: `private class CouchbaseCacheAdapter <K, V> implements BasicOperations.Cache<K, V>`. See as an example: https://github.com/radargun/radargun/commit/373b353317d0a05b9a7e9e7009a3c376500b8d24#diff-df93a8f6e3d2569df9283e049b55bfbc

RadarGun allows different keys and values generators. Here are some that we support https://github.com/radargun/radargun/tree/master/extensions/cache/src/main/java/org/radargun/stages/cache/generators

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
```diff
@@ -0,0 +1,92 @@
+package org.radargun.service;
+
+import com.couchbase.client.java.document.ByteArrayDocument;
+import org.radargun.traits.BasicOperations;
+
+import java.nio.charset.StandardCharsets;
+
+public class CouchbaseOperations implements BasicOperations {
```
diegolovison: I prefer to see the name of the class like `CouchbaseClientOperations`

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseService.java
```diff
@@ -0,0 +1,56 @@
...
+import com.couchbase.client.java.Bucket;
+import com.couchbase.client.java.CouchbaseCluster;
+import com.couchbase.client.java.env.CouchbaseEnvironment;
+import com.couchbase.client.java.env.DefaultCouchbaseEnvironment;
+import org.radargun.Service;
+import org.radargun.config.Property;
+import org.radargun.traits.Lifecycle;
+
+@Service(doc = "Couchbase")
+public class CouchbaseService implements Lifecycle {
```
diegolovison: Same here: CouchbaseClientService

plugins/couchbase/src/main/resources/plugin.properties
```diff
@@ -0,0 +1,3 @@
+service.default org.radargun.service.CouchbaseService
```
diegolovison: For me, it is missing the `CouchbaseServerService` class implementation. For example, I would like to run RadarGun and compare the project X with Couchbase in a perf lab. Without `CouchbaseServerService` I will need to setup the server everytime. I saw https://docs.couchbase.com/server/4.1/getting-started/installing.html that we can download and install the server. A good entry point is https://github.com/radargun/radargun/blob/master/plugins/etcd/src/main/java/org/radargun/service/EtcdServerService.java

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseOperations.java
```diff
@@ -0,0 +1,92 @@
...
+      private void put(String key, byte[] value) {
+         ByteArrayDocument doc = ByteArrayDocument.create(key, value);
+         service.bucket.upsert(doc);
+      }
+
+      @Override
+      public byte[] getAndPut(byte[] keyBytes, byte[] value) {
+         String key = createKey(keyBytes);
+         byte[] document = get(key);
+         put(key, value);
```
diegolovison: @rvansa should it be in a unique operation or doesn't support it, right?

MikeGoldsmith:
@diegolovison I've updated the class names / files and refactored `CouchbaseClientOperations` to utilize **K** and **V** type params in the inner `CouchbaeCacheAdapter`.

I'm not sure downloading and installing Couchbase as part of a test run is a good idea. It is a fairly large process / download. There are docker images and/or vagrant scripts that we may be able to utilize.

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseClientOperations.java
```diff
@@ -0,0 +1,105 @@
...
+            return true;
+         } catch(Exception e) {
+            return false;
+         }
+      }
+
+      @Override
+      public V getAndRemove(K key) {
+         String keyStr = keyAsString(key);
+         V document = get(keyStr);
```
diegolovison: It should be only one operation in the server. In this case, we are calling the server twice. If Couchbase doesn't support returning the previous value when removing the key we need to throw an exception.

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseClientOperations.java
```diff
@@ -0,0 +1,105 @@
...
+      private void put(String key, V value) {
+         ByteArrayDocument doc = ByteArrayDocument.create(key, valueAsByteArray(value));
+         service.bucket.upsert(doc);
+      }
+
+      @Override
+      public V getAndPut(K key, V value) {
+         String keyStr = keyAsString(key);
+         V document = get(keyStr);
+         put(keyStr, value);
```
diegolovison: It should be only one operation in the server. In this case, we are calling the server twice. If Couchbase doesn't support returning the previous value when putting the key we need to throw an exception.

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseClientOperations.java
```diff
@@ -0,0 +1,105 @@
+package org.radargun.service;
```
diegolovison: I just saw that the folder structure is using (dot) instead of a folder separator.
The name of the file is: `org.radargun.service/CouchbaseClientOperations.java` and it should be `org/radargun/service/CouchbaseClientOperations.java`

plugins/couchbase/src/main/java/org.radargun.service/CouchbaseClientService.java
```diff
@@ -0,0 +1,56 @@
+package org.radargun.service;
```
diegolovison: Same here about the folder separator

diegolovison:
@MikeGoldsmith I am going to squash and merge using the GitHub button. IIRC, the commit will be in my name. I would like to see the commit with who developed that. Would you like to squash or can I move forward?

diegolovison:
@MikeGoldsmith I was going to use the "Squash and merge" button in the GitHub interface and I just remembered that last time that I did it, I was the commit author. I prefer to see your team as an author of this commit. Could you please squash?

MikeGoldsmith:
@diegolovison commits have been squashed

MikeGoldsmith:
@diegolovison commits have been squashed 👍

diegolovison:
Thanks @MikeGoldsmith and @daschl

PR #602, opened 2019-06-07
Allow the Loader report the service time when benchmarking the SUT

diegolovison:
Allow the Loader report the service time when benchmarking the SUT with the Compensation of Coordinate Omission in place.

This change allows the following:

```xml
      .......
      <config name="dist">
         <setup plugin="fooPlugin">
            <environment>
               <var name="LATENCY_AS_SERVICE_TIME" value="true" />
            </environment>
           ......
           <embedded file="library.xml" cache="distCache"/>
         </setup>
      </config>
     <config name="dist-co-fix">
         <setup plugin="fooPlugin">
            <environment>
               <var name="LATENCY_AS_SERVICE_TIME" value="false" />
            </environment>
            .....
           <embedded file="library.xml" cache="distCache"/>
         </setup>
      </config>
   .....
      <l:basic-operations-test test-name="reads10" ..... cycle-time="10L" report-latency-as-service-time="${env.LATENCY_AS_SERVICE_TIME:false}">
      ........
      </l:basic-operations-test>
    ......
   </rg:scenario>
```

The report will be:

Screen Shot 2019-06-07 at 08 06 42

In this case, both tests are running in the same scenario but times are reported different:
dist: it is reporting service time
dist-co-fix: it is reporting latency time

### DIFF
--- a/core/src/main/java/org/radargun/stages/test/Stressor.java
+++ b/core/src/main/java/org/radargun/stages/test/Stressor.java
@@ -49,6 +49,8 @@ public class Stressor extends Thread {
    long uniformRateLimiterOpIndex = 0;
    long uniformRateLimiterStart = Long.MIN_VALUE;
 
+   final boolean reportLatencyAsServiceTime;
+
    public Stressor(TestStage stage, OperationLogic logic, int globalThreadIndex, int threadIndex, CountDownLatch threadCountDown) {
       super("Stressor-" + threadIndex);
       this.stage = stage;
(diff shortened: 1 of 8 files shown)


rvansa:
I think that 'compensate for CO' would be more precise than 'don't suffer from CO'.

diegolovison:
@rvansa Done!

alanfx:
@diegolovison this looks good to me. You have any other comments @rvansa?

alanfx:
Merging

PR #604, opened 2019-06-20
Fix NoSuchMethodError: ParserRegistry.parse

diegolovison:


alanfx:
@diegolovison I'm guessing this is the new way, so I'll merge! Thanks

PR #605, opened 2019-06-24
Removed duplicated infinispan-multimap

diegolovison:
When running `mvn ....  -pl \!plugins/infinispan100` it will fail because of the duplicated dependency. This PR fix that issue


alanfx:
@diegolovison I'll merge this, but it since it seems like the `infinispan92` plugin includes this dependency already.

PR #610, opened 2019-09-09
Add systemProperty parameter on @Property

gustavolira:
Parameters can be passed via System Environment, on this way we can use the same benchmark file with any ispn version


diegolovison:
Wildfly was removed from master
See: https://github.com/infinispan/infinispan/commit/306f37a5edadf1a96adff85e00f61429e262ff95 , https://github.com/radargun/radargun/pull/611#event-2666095092

PR #611, opened 2019-09-26
Legacy Wildfly server was removed from ISPN10

diegolovison:


alanfx:
@diegolovison This looks good. Merging...

PR #612, opened 2019-09-27
ISPN removed deprecated TcpTransportFactory

diegolovison:


plugins/infinispan52/src/main/java/org/radargun/service/InfinispanHotrodService.java
```diff
@@ -71,7 +71,10 @@ public boolean isRunning() {
       return managerNoReturn != null && managerNoReturn.isStarted();
    }
 
-   protected TcpTransportFactory getTransportFactory(RemoteCacheManager manager) {
+   /*
+    * ISPN10 removed TcpTransportFactory and a class not found exception will throws.
+    */
+   protected Object getTransportFactory(RemoteCacheManager manager) {
```
alanfx: Why is the return type `Object` here?

plugins/infinispan60/src/main/java/org/radargun/service/Infinispan60HotrodService.java
```diff
@@ -71,6 +72,27 @@ public void init() {
    protected ConfigurationBuilder getDefaultHotRodConfig() {
       ConfigurationBuilder builder = new ConfigurationBuilder();
       builder.connectionPool().maxActive(maxConnectionsServer).maxTotal(maxConnectionsTotal);
+      parseServerAddresses().forEach((address) -> builder.addServer().host(address.getHost()).port(address.getPort()));
+      createQueryConfiguration(builder);
+      return builder;
+   }
+
+   protected ConfigurationBuilder createQueryConfiguration(ConfigurationBuilder builder) {
```
alanfx: Have you verified that older plugins are still working with this change?

plugins/infinispan52/src/main/java/org/radargun/service/InfinispanHotrodService.java
diegolovison: ISPN10 removed TcpTransportFactory and a class not found exception will throws.

plugins/infinispan60/src/main/java/org/radargun/service/Infinispan60HotrodService.java
diegolovison: Yes. It was copied and pasted

plugins/infinispan52/src/main/java/org/radargun/service/InfinispanHotrodService.java
```diff
@@ -97,8 +100,8 @@ protected TcpTransportFactory getTransportFactory(RemoteCacheManager manager) {
    @Override
    public Map<String, Number> getValues() {
       Map<String, Number> values = new HashMap<>();
-      TcpTransportFactory nrFactory = getTransportFactory(managerNoReturn);
-      TcpTransportFactory frFactory = getTransportFactory(managerForceReturn);
+      TcpTransportFactory nrFactory = (TcpTransportFactory) getTransportFactory(managerNoReturn);
+      TcpTransportFactory frFactory = (TcpTransportFactory) getTransportFactory(managerForceReturn);
```
gustavolira: Why in this context a class not found exception will not be thrown?

diegolovison: Because of https://github.com/radargun/radargun/blob/master/core/src/main/java/org/radargun/config/InitHelper.java#L86

gustavolira: Just a nit, you can cast on this linehttps://github.com/radargun/radargun/blob/62f349c805206ff7ea35cf296e5934814363e860/plugins/infinispan52/src/main/java/org/radargun/service/InfinispanHotrodService.java#L84 and keep the return as TcpTransportFactory instead of Object, on this way you can avoid cast every time you call this method like https://github.com/radargun/radargun/blob/62f349c805206ff7ea35cf296e5934814363e860/plugins/infinispan52/src/main/java/org/radargun/service/InfinispanHotrodService.java#L103

diegolovison: If a method is using a class that is not present in the plugin classpath `infinispan-100` a class not found will throw. The method is recursive and will check all [getSuperclass](https://github.com/radargun/radargun/blob/master/core/src/main/java/org/radargun/config/InitHelper.java#L102)

gustavolira:
LGTM, thanks @diegolovison

gustavolira:
merged #612 into master

PR #614, opened 2019-10-01
Support ISPN10 bind address

diegolovison:


diegolovison:
@alanfx @gustavolira please review the PR. I will suggest looking for each commit.
Jenkins and Infinispan regression tests are green.
Once, you have been reviewed, I am going to squash and merge.
Tks

plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
```diff
@@ -23,46 +22,74 @@
...
-   public CacheManagerInfo getCacheManager() {
-      String url = String.format("http://%s:%s/rest/%s", serverIp, this.defaultPort, CACHE_MANAGER_RESOURCE);
-      HttpRequest get = HttpRequest.newBuilder()
-            .uri(URI.create(url))
-            .GET()
-            .build();
+   private String lookupServerIp() {
+      // the following are predefined values in the INFINISPAN10 XML Files
+      String interfaceName = System.getProperty("radargun.infinispan.interfaceName", "public");
+      String bindAddress = System.getProperty("infinispan.bind.address", "127.0.0.1");
```
alanfx: @diegolovison I'm worried this may not work for all allowable JGroups properties. There are a couple of issues I see:

1. The JGroups XML file can use a different variable name instead of `infinispan.bind.address`
1. JGroups allows you to set the bind address value to non IP address values.

Is it possible to get this information from the Infinispan server directly? Maybe we can talk to Tristan about this?

plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
```diff
@@ -23,46 +22,74 @@
...
    private Integer defaultPort;
    private static final String CACHE_MANAGER_RESOURCE = "v2/cache-managers/clustered";
    private String serverIp;
-   private final Log log = LogFactory.getLog(getClass());
 
    public InfinispanRestAPI(Integer defaultPort) {
-      mapper = new ObjectMapper();
-      mapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
-      mapper.setPropertyNamingStrategy(PropertyNamingStrategy.SNAKE_CASE);
-      httpClient = HttpClient.newBuilder().version(HttpClient.Version.HTTP_1_1).build();
```
alanfx: @gustavolira sorry about removing this code, but I didn't think about the consequences of relying on JDK 11 library code when it went into the codebase.

plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
```diff
@@ -23,46 +22,74 @@
...
       CacheManagerInfo cacheManagerInfo = null;
-      try {
-         response = httpClient.send(get, HttpResponse.BodyHandlers.ofString());
-         String json = response.body();
-         cacheManagerInfo = mapper.readValue(json, CacheManagerInfo.class);
-      } catch (IOException e) {
-         e.printStackTrace();
-      } catch (InterruptedException e) {
-         e.printStackTrace();
+      while (cacheManagerInfo == null && count++ < 5) {
```
alanfx: Would it make more sense to add this retry logic to the `doGet` method directly? It seems like this will be useful for any REST API calls.

parent/pom.xml
```diff
@@ -665,117 +675,43 @@
...
+            <executions>
+               <execution>
+                  <id>enforce-java</id>
+                  <goals>
+                     <goal>enforce</goal>
+                  </goals>
+                  <configuration>
+                     <rules>
+                        <requireJavaVersion>
+                           <version>[1.8.0-0,1.8.0-999]</version>
```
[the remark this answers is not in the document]
diegolovison: No. It will be greater than `1.8.0-0`. In this case, we will be able to build JDK11

plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
diegolovison: It is based on Infinispan. See https://github.com/infinispan/infinispan/blob/master/server/runtime/src/main/server/server/conf/infinispan.xml

plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
diegolovison: Yeah. I am sorry as well.

plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
diegolovison: Done

plugins/infinispan-snapshot/pom.xml
```diff
@@ -161,4 +161,33 @@
       </profile>
    </profiles>
 
+   <build>
+      <plugins>
+         <plugin>
+            <artifactId>maven-resources-plugin</artifactId>
+            <version>3.1.0</version>
+            <executions>
+               <execution>
```
diegolovison: @alanfx we don't need to maint the resources in snapshot. it is dynamic now based on `${plugins.infinispan-latest}`

diegolovison:
@alanfx comments addressed. 

@alanfx we don't need to maintain the resources in the snapshot. It is dynamic now based on ${plugins.infinispan-latest}.

See: https://github.com/radargun/radargun/pull/614/files#diff-ea895097697ffb4b652309a70e46af40R170-R188

plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
alanfx: Yes, but a different XML file can use a different variable name. That's my point.

parent/pom.xml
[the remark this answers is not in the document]
alanfx: OK, I think you can use `requireJavaVersion` instead https://maven.apache.org/enforcer/enforcer-rules/requireJavaVersion.html

plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
gustavolira: np

alanfx:
Merging, even though I'd prefer that the Maven Enforcer message happened before the `tools. jar` dependency error.

PR #615, opened 2019-10-07
Fix retrieve cache manager

diegolovison:


plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
```diff
@@ -52,17 +52,13 @@ private String lookupServerIp() {
    }
 
    public CacheManagerInfo getCacheManager() {
-      // network is not reliable
-      int count = 0;
+      String url = String.format("http://%s:%s/rest/%s", serverIp, this.defaultPort, CACHE_MANAGER_RESOURCE);
```
alanfx: Should `"http://%s:%s/rest/%s"` be a constant, in case any other commands are added in this class? Not a huge deal

diegolovison: I am going to drop this class and use the Infinispan10 rest client :)

alanfx:
@diegolovison looks ok, but I will ask for a constant with the next update!

diegolovison:
I am was looking to fix in this PR. A new one will come tomorrow morning. Thank you.

PR #617, opened 2019-10-18
Return empty CacheManagerInfo if serverNG is busy

diegolovison:
With this PR I am going to fix 4 situations:

1) When serverNG is busy we should return an empty ISPN Cache Manager Info because when we have Cluster Split Verify enabled, a long timeout, in this case, 10secs ( 1 sec X 10 times ) "means" that something is wrong in the cluster. GC is busy or another kind of situation.

2) The regex for the server startup was wrong. See https://github.com/infinispan/infinispan/pull/7449

3) When `scheduleAtFixedRate` is called and throw an exception a subsequent call won't be executed

4) Only report data if all stages have a custom return code.
Example
```
<perfrepo xmlns="urn:radargun:reporters:reporter-perfrepo:3.0" master-return-codes="0">
```

### DIFF
--- a/core/src/main/java/org/radargun/reporting/AbstractReporter.java
+++ b/core/src/main/java/org/radargun/reporting/AbstractReporter.java
@@ -1,0 +1,30 @@
+package org.radargun.reporting;
+
+import java.util.Collection;
+import java.util.List;
+
+import org.radargun.config.MasterConfig;
+import org.radargun.config.Property;
+import org.radargun.logging.Log;
+import org.radargun.logging.LogFactory;
+
+public abstract class AbstractReporter implements Reporter {
+
+   private static Log log = LogFactory.getLog(AbstractReporter.class);
+
+   @Property(doc = "Which master return code should make the report run. Default is all. Comma separated.")
+   private List<Integer> masterReturnCodes;
+
+   public void run(MasterConfig masterConfig, Collection<Report> reports, int masterReturnCode) {
+      String className = this.getClass().getSimpleName();
+      if (masterReturnCodes == null || masterReturnCodes.isEmpty() || masterReturnCodes.contains(masterReturnCode)) {
+         log.info("Running reporter " + className);
+         run(masterConfig, reports);
+      } else {
+         log.info(String.format("Skipping report %s. Return codes are %s and the master return code is %d",
+               className, masterReturnCodes, masterReturnCode));
+      }
+   }
+
+   public abstract void run(MasterConfig masterConfig, Collection<Report> reports);
+}
--- a/core/src/main/java/org/radargun/reporting/AbstractReporter.java
+++ b/core/src/main/java/org/radargun/reporting/AbstractReporter.java
@@ -1,28 +1,42 @@
 package org.radargun.reporting;
 
 import java.util.Collection;
-import java.util.List;
 
 import org.radargun.config.MasterConfig;
 import org.radargun.config.Property;
 import org.radargun.logging.Log;
 import org.radargun.logging.LogFactory;
 
+/**
+ * Control the report generation based on the stage failure
+ *
+ * @author Diego Lovison &lt;dlovison@redhat.com&gt;
+ */
 public abstract class AbstractReporter implements Reporter {
 
    private static Log log = LogFactory.getLog(AbstractReporter.class);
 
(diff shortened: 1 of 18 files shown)


plugins/infinispan100/src/main/java/org/radargun/service/ServerLogPattern.java
```diff
@@ -9,7 +9,7 @@
 public enum ServerLogPattern {
 
    START_OK(".*\\[org\\.infinispan\\.SERVER\\].*started in.*"),
-   START_ERROR(".*\\[org\\.infinispan\\.SERVER\\].*Infinispan Server stopping.*"),
+   START_ERROR(".*\\[org\\.infinispan\\.SERVER\\].*Server failed to start.*"),
```
gustavolira: @diegolovison How did you get this message?
If you start the server and try to start another instance without change the port, the server will not start, but you will not see `Server failed to start`

diegolovison: You need https://github.com/infinispan/infinispan/pull/7449

gustavolira: @diegolovison lol it's not fair, this PR was merged minutes ago! I'm testing again

gustavolira: After pull and compile upstream with the last modifications I still receive the same messages.
- `pull ispn master`
- `mvn clean install ... `
- run `server/runtime/target/infinispan-server-10.0.0-SNAPSHOT/bin/server.sh`
- run `server/runtime/target/infinispan-server-10.0.0-SNAPSHOT/bin/server.sh` again

alanfx:
Hey @diegolovison, the code looks ok to me, but I would like to dig into the first issue more 

> When serverNG is busy we should return an empty ISPN Cache Manager Info because when we have Cluster Split Verify enabled, a long timeout, in this case, 10secs ( 1 sec X 10 times ) "means" that something is wrong in the cluster. GC is busy or another kind of situation.

and understand the last issue as well!

> Only report data if all stages have a custom return code.

The fact that ServerNG isn't responding for a long time is not good and probably a bug. I think the last issue it trying to make sure each stage completes successfully before reporting to PerfRepo? I think this needs some more documentation, since it will affect every stage?

diegolovison:
> The fact that ServerNG isn't responding for a long time is not good and probably a bug.
Yes. And the code in RadarGun now is able to deal with this situation. Before that change, the RadarGun was passing the test when it shouldn't.

> each stage completes successfully before reporting to
It is a configuration. By default, all data will be reported. If you would like to customize you need to add `master-return-codes="0"`

> since it will affect every stage
It is documented here https://github.com/radargun/radargun/pull/617/files#diff-d2f544a284d57c18ed2c8ecb87a549ebR15

core/src/main/java/org/radargun/reporting/AbstractReporter.java
```diff
@@ -0,0 +1,30 @@
...
+import org.radargun.config.MasterConfig;
+import org.radargun.config.Property;
+import org.radargun.logging.Log;
+import org.radargun.logging.LogFactory;
+
+public abstract class AbstractReporter implements Reporter {
+
+   private static Log log = LogFactory.getLog(AbstractReporter.class);
+
+   @Property(doc = "Which master return code should make the report run. Default is all. Comma separated.")
```
alanfx: @diegolovison from our discussion, I thought this property would be a `boolean` named `skipOnStageFailures`?

diegolovison: @alanfx yes. it is a boolean. https://github.com/radargun/radargun/pull/617/files#diff-d2f544a284d57c18ed2c8ecb87a549ebR20

alanfx: @diegolovison OK, I guess I was confused. I thought that this would be the only property required, and that if `skipOnStageFailures` was `true` then the reporter would check for failures in the stages. Is that possible?

diegolovison: @alanfx master run all stages and if one failed the data won't be reported when `skipOnStageFailures = true`

alanfx:
@diegolovison looks good and thanks for bearing with me! Merging...

PR #620, opened 2019-11-01
Fixed jgroups config that was missing

gustavolira:


diegolovison:
LGTM. Thanks @gustavolira

PR #621, opened 2019-11-07
Fix Infinispan Server stopped message

diegolovison:
After the fix https://github.com/radargun/radargun/commit/cd72ce3668d37742b370d1735d1c6cb5958410da#diff-e0fc040510f958a1c18589dea7e1fcdfR48-R51 this stop working.
Fixed

### DIFF
--- a/plugins/infinispan52/src/main/java/org/radargun/service/InfinispanServerLifecycle.java
+++ b/plugins/infinispan52/src/main/java/org/radargun/service/InfinispanServerLifecycle.java
@@ -23,11 +23,11 @@ public class InfinispanServerLifecycle extends ProcessLifecycle<InfinispanServer
    }
 
    protected Pattern getStartError() {
-      return Pattern.compile(".*\\[org\\.jboss\\.as\\].*started in.*");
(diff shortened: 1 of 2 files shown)


plugins/infinispan52/src/main/java/org/radargun/service/InfinispanServerLifecycle.java
```diff
@@ -23,11 +23,11 @@ protected Pattern getStartOK() {
    }
 
    protected Pattern getStartError() {
-      return Pattern.compile(".*\\[org\\.jboss\\.as\\].*started in.*");
+      return Pattern.compile(".*\\[org\\.jboss\\.as\\].*stopped in.*");
```
andyuk1986: Hi @diegolovison , I think the message here should be not `stopped` but:
Pattern.compile(".*\\[org\\.jboss\\.as\\].*started \\(with errors\\) in.*");
(at least this is how it was before Gustavo's changes)
Would be great if you will fix that too. Then I will rebase the PR as I am facing this issue too.

andyuk1986:
LGTM, thanks for the fix.

PR #623, opened 2019-11-29
Support ISPN 10 logging configuration file

diegolovison:


plugins/infinispan100/src/main/java/org/radargun/service/Infinispan100ServerService.java
```diff
@@ -42,6 +42,9 @@
    @Property(doc="security config. If set, it will replace the default <security> config")
    private String securityConfig;
 
+   @Property(doc="logging config. If set, it will setup the server with the logging configuration file provided")
+   private String loggingConfig;
```
alanfx: Suggested change:

A full path to a logging configuration file. If set, the server will use the provided logging configuration file. If not set, the default server logging configuration is used.

diegolovison:
@alanfx done

alanfx:
Looks good. Merging

PR #635, opened 2020-02-05
Aggregate the tx operation in the non-tx operation using Duration

diegolovison:
@alanfx the change is explained in https://github.com/radargun/radargun/pull/635/files#diff-9158f9969a715eda006055cd03e939d4R239-R250

### DIFF
--- a/core/src/main/java/org/radargun/stages/test/Stressor.java
+++ b/core/src/main/java/org/radargun/stages/test/Stressor.java
@@ -68,7 +68,7 @@ public class Stressor extends Thread {
    }
 
    private boolean recording() {
-      return this.started && !stage.isFinished();
+      return this.started;
    }
 
    @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/radargun/stages/test/Stressor.java
```diff
@@ -68,7 +68,7 @@ public Stressor(TestStage stage, OperationLogic logic, int globalThreadIndex, in
    }
 
    private boolean recording() {
-      return this.started && !stage.isFinished();
+      return this.started;
```
alanfx: @diegolovison curious why you changed this line?

diegolovison: Because it will generate different numbers of requests.
The fix is the line that you mentioned and https://github.com/radargun/radargun/pull/635/files#diff-9158f9969a715eda006055cd03e939d4R147

For example, a request could be reported in statistics in https://github.com/radargun/radargun/pull/635/files#diff-9158f9969a715eda006055cd03e939d4R189 but not in https://github.com/radargun/radargun/pull/635/files#diff-9158f9969a715eda006055cd03e939d4R251

alanfx:
Thanks @diegolovison Merging

PR #636, opened 2020-02-07
Skip chart serie if there is no data

diegolovison:


alanfx:
Looks good to me @diegolovison! Merging

PR #638, opened 2020-02-19
Changes to FlightRecorder and JVM arguments

alanfx:
Remove -XX:+UnlockCommercialFeatures from FlightRecorder options
Update the logging and GC logging options
Update the version for the jdg73 plugin to JDG 7.3.5

### DIFF
--- a/core/src/main/java/org/radargun/config/FlightRecorder.java
+++ b/core/src/main/java/org/radargun/config/FlightRecorder.java
@@ -17,6 +17,8 @@ public class FlightRecorder implements VmArg {
 
    @Property(doc = "Flight Recorder Options")
    private String flightRecorderOptions;
+   
+   private boolean isOracle = !System.getProperty("java.runtime.name").contains("OpenJDK");
 
    @Override
    public void setArgs(List<String> args) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/radargun/config/FlightRecorder.java
```diff
@@ -17,6 +17,9 @@
 
    @Property(doc = "Flight Recorder Options")
    private String flightRecorderOptions;
+   
+   @Property(doc = "Is this an Oracle JVM? Default is false.")
+   private boolean isOracle = false;
```
rvansa: This is something you can autodetect, see `java.vendor` (regrettably Oracle Corporation for both commercial and OpenJDK) and `java.runtime.name` (that differs).

alanfx: Thanks, @rvansa! Missed the `java.runtime.name` property

alanfx: Updated!

diegolovison:
Thanks @alanfx

PR #639, opened 2020-02-24
Fix jmx

diegolovison:


andyuk1986:
Looks good to me. Thanks, Diego :)

PR #645, opened 2020-03-26
Fix infinispan jaxb-core missing dependency

diegolovison:
dependencyManagement for org.infinispan:infinispan-parent overwrite the core/pom.xml jaxb-core dependency
Infinisnpan jaxb-core version 2.3.1 doesn't exists
Once fixed in infinispan/master, remove this workaround


gustavolira:
Merged #645 into master

PR #646, opened 2020-03-27
Throw exception if an error occurred during the report generation

diegolovison:


gustavolira:
Merged #646 into master

PR #649, opened 2020-04-09
Replace infinispan-parent dependency to infinispan-bom

gustavolira:


plugins/etcd/pom.xml
```diff
@@ -27,7 +27,6 @@
       <dependency>
          <groupId>commons-io</groupId>
          <artifactId>commons-io</artifactId>
-         <version>2.4</version>
```
diegolovison, line 30: `commons-io` is only required for `etcd` plugin.
Why we should move to `plugins/pom.xml` ?

plugins/infinispan-snapshot/dev.properties
```diff
@@ -0,0 +1 @@
+version.jboss.narayana=5.9.8.Final
```
diegolovison, line 1: I didn't see where we are using this ?

plugins/jdg62/pom.xml
```diff
@@ -75,19 +75,6 @@
          </exclusions>
       </dependency>
 
-      <dependency>
```
diegolovison, line 78: This pom.xml doesn't use infinispan-bom. I would like to keep as yes.

plugins/jdg63/pom.xml
```diff
@@ -88,19 +88,6 @@
          <scope>provided</scope>
       </dependency>
 
-      <dependency>
```
diegolovison, line 91: This pom.xml doesn't use infinispan-bom. I would like to keep as yes.

plugins/jdg64/pom.xml
```diff
@@ -87,19 +87,6 @@
          </exclusions>
       </dependency>
 
-      <dependency>
```
diegolovison, line 90: This pom.xml doesn't use infinispan-bom. I would like to keep as yes.

plugins/jdg65/pom.xml
```diff
@@ -86,19 +86,6 @@
            </exclusions>
        </dependency>
 
-       <dependency>
```
diegolovison, line 89: This pom.xml doesn't use infinispan-bom. I would like to keep as yes.

plugins/jdg70/pom.xml
```diff
@@ -86,19 +86,6 @@
            </exclusions>
        </dependency>
 
-       <dependency>
```
diegolovison, line 89: This pom.xml doesn't use infinispan-bom. I would like to keep as yes.

plugins/pom.xml
```diff
@@ -31,5 +31,32 @@
...
+            <exclusion>
+               <groupId>org.jboss.logging</groupId>
+               <artifactId>jboss-logging-spi</artifactId>
+            </exclusion>
+         </exclusions>
+      </dependency>
+
    </dependencies>
+
+   <dependencyManagement>
```
diegolovison, line 52: See my comment above

plugins/infinispan-snapshot/dev.properties
alanfx, line 1: Hey @gustavolira is there any way we could get this version from this property file?: https://github.com/infinispan/infinispan/blob/5e1864da273a448f111629c2029f89c6fe5bf07f/build-configuration/pom.xml

plugins/pom.xml
```diff
@@ -31,5 +31,32 @@
          <artifactId>jboss-logging</artifactId>
          <version>3.3.0.Final</version>
       </dependency>
+
+      <!-- needed to avoid receive NoClassDefFoundError: com/arjuna/ats/arjuna/common/arjPropertyManager
```
diegolovison, line 35: What is the plugin that requires this dependency? Could you share the stack trace ?

plugins/pom.xml
```diff
@@ -31,5 +31,32 @@
...
+           during build -->
+      <dependency>
+         <groupId>org.jboss.narayana.jta</groupId>
+         <artifactId>narayana-jta</artifactId>
+         <version>5.10.4.Final</version>
```
alanfx, line 40: I think this should either use `${version.jboss.narayana}` or get the property from here: https://github.com/infinispan/infinispan/blob/5e1864da273a448f111629c2029f89c6fe5bf07f/build-configuration/pom.xml#L168

plugins/pom.xml
```diff
@@ -31,5 +31,32 @@
...
+      </dependency>
+
    </dependencies>
+
+   <dependencyManagement>
+      <dependencies>
+         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.6</version>
```
alanfx, line 57: Same here. This version should be in dev.properties or retrieved from here: https://github.com/infinispan/infinispan/blob/7de021c034851e56efaad8abc253422543a8b8cf/pom.xml#L1117

plugins/infinispan-snapshot/dev.properties
gustavolira: @alanfx actually this file should be deleted. Good catch, thanks!

plugins/etcd/pom.xml
gustavolira: @diegolovison  Actually I didn't move, I only added a `dependencyManagement` just to be able to remove the `version` from `plugins/etcd/pom.xml`. IMO is more organised to avoid add versions spread into the project

plugins/jdg62/pom.xml
gustavolira: I only removed `narayana-jta` because is not being used, WDYT?

alanfx:
@gustavolira this Maven plugin lets you reference external properties files: http://www.mojohaus.org/properties-maven-plugin/

plugins/pom.xml
gustavolira: Unfortunately we can't use the property from here https://github.com/infinispan/infinispan/blob/5e1864da273a448f111629c2029f89c6fe5bf07f/build-configuration/pom.xml#L168` because we are not using `infinispan-bom` as parent, only as imported.
I'll create `version.jboss.narayana` in the properties

gustavolira:
> @gustavolira this Maven plugin lets you reference external properties files: http://www.mojohaus.org/properties-maven-plugin/

@alanfx I tried used this plugin before, that's why I created `dev.properties` and forgot to delete before my commit.
I can try to include this plugin in another PR. I spent some time trying to do this plugin work to replace the properties but didn't work as expected and even if it works during the compilation the IDE will always complain warning that this property doesn't exist.

plugins/pom.xml
gustavolira:
```
  [echo] Packaging plugin infinispan90...
     [copy] Copying 90 files to /Users/gustavo/git/radargun/target/distribution/RadarGun-3.0.0-SNAPSHOT/plugins/infinispan90
     [java] Exception in thread "main" java.lang.NoClassDefFoundError: com/arjuna/ats/arjuna/common/arjPropertyManager
     [java] 	at org.radargun.service.InfinispanEmbeddedService.<clinit>(InfinispanEmbeddedService.java:41)
     [java] 	at java.lang.Class.forName0(Native Method)
     [java] 	at java.lang.Class.forName(Class.java:264)
     [java] 	at org.radargun.ServiceHelper.loadServices(ServiceHelper.java:104)
     [java] 	at org.radargun.config.PluginSchemaGenerator.main(PluginSchemaGenerator.java:61)
     [java] Caused by: java.lang.ClassNotFoundException: com.arjuna.ats.arjuna.common.arjPropertyManager
     [java] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
     [java] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
     [java] 	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
     [java] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
```

PR #650, opened 2020-05-11
Useful changes / additions

andyuk1986:


plugins/infinispan93/src/main/java/org/radargun/service/Infinispan93HotrodService.java
```diff
@@ -43,6 +57,20 @@ protected ConfigurationBuilder getDefaultHotRodConfig() {
...
+
+      config.maxRetries(maxRetries).socketTimeout(connectionTimeout).connectionTimeout(connectionTimeout);
+
+      if (propertiesPath != null) {
+         Properties p = new Properties();
+         try (Reader r = new FileReader(propertiesPath)) {
+            p.load(r);
+            config.withProperties(p);
+         } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
diegolovison, line 69: Should we throw IllegalStateException?

plugins/infinispan93/src/main/java/org/radargun/service/Infinispan93HotrodService.java
```diff
@@ -43,6 +57,20 @@ protected ConfigurationBuilder getDefaultHotRodConfig() {
...
+         } catch (IOException e) {
+            e.printStackTrace();
```
diegolovison, line 71: Should we throw IllegalStateException?

gustavolira: nit: if the `catch` are doing the same stuff, probability would be better combine them with `FileNotFoundException | IOException`

plugins/infinispan93/src/main/java/org/radargun/service/Infinispan93HotrodService.java
andyuk1986: Done!

plugins/infinispan93/src/main/java/org/radargun/service/Infinispan93HotrodService.java
andyuk1986: @gustavolira @diegolovison it was not possible to combine them, as FileNotFoundException extends IOException (java didn't allow it), so I am just catching IOException and throwing IllegalStateException with message. Hope this ok for both of you :)

gustavolira: Sorry Anna I didn't notice that

diegolovison:
Thanks @andyuk1986

PR #655, opened 2020-05-21
Fix NPE when the slaves are down for BackgroundStatisticsStopStage

diegolovison:


extensions/cache/src/main/java/org/radargun/stages/cache/background/BackgroundStatisticsStopStage.java
```diff
@@ -66,8 +66,11 @@ public StageResult processAckOnMaster(List<DistStageAck> acks) {
          long min = Long.MAX_VALUE, max = Long.MIN_VALUE;
          for (Map.Entry<Integer, Long> iterationData : cacheSizes.getColumn(iteration).entrySet()) {
             slaveResults.put(iterationData.getKey(), new Report.SlaveResult(String.valueOf(iterationData.getValue()), false));
-            min = Math.min(min, iterationData.getValue());
-            max = Math.max(max, iterationData.getValue());
+            // it could null if the slave is down
```
alanfx, line 69: it could null if the slave is down -> iterationData could be null if the slave is down

gustavolira:
Merged #655 into master

PR #656, opened 2020-06-01
Fix ISPN NoSuchMethodError: DistributionManager.getConsistentHash

diegolovison:


gustavolira:
Merged #656 into master

PR #659, opened 2020-06-12
ISPN 11 Query support

diegolovison:


extensions/query/src/test/java/org/radargun/util/QueryTraitRepository.java
```diff
@@ -61,93 +61,93 @@ public Builder(ConcurrentHashMap cache) {
          }
 
          @Override
-         public Query.Builder subquery() {
+         public org.radargun.traits.Query.Builder subquery() {
```
alanfx, line 64: Why is this needed? Can't you import the class with the name `Query` to avoid this on line 3 above?

diegolovison: It should be good now

alanfx:
Looks good, @diegolovison! Merging

PR #662, opened 2020-06-22
Added jdg81 plugin.

andyuk1986:


plugins/infinispan100/src/main/java/org/radargun/service/Infinispan100ServerService.java
```diff
@@ -42,6 +42,15 @@
    @Property(doc="A full path to a logging configuration file. If set, the server will use the provided logging configuration file. If not set, the default server logging configuration is used.")
    private String loggingConfig;
 
+   @Property(doc="Boolean property for enabling user creation on server. If set to true, the new user will be added with provided username and password. Default is false.")
+   private boolean addUser=false;
```
gustavolira, line 46: Since ispn10 and jdg8.0 authentication is not mandatory, shouldn't those modifications be included only on jdg81 or infinispan11 plugin?

plugins/jdg81/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+
+      <!--
+      TODO
+      dependencyManagement for org.infinispan:infinispan-parent which overwrite the core/pom.xml jaxb-core dependency
+      Infinisnpan jaxb-core version 2.3.1 doesn't exists
+      Once fixed in infinispan/master, remove this workaround
+      -->
+      <dependency>
+         <groupId>org.glassfish.jaxb</groupId>
+         <artifactId>jaxb-core</artifactId>
```
gustavolira, line 158: Did you try remove `jaxb-core`, I think `infinispan-parent` is no more overriding our jaxb-core dependency

andyuk1986: @gustavolira honestly, I didn't touch any dependencies. I have just copy/pasted this pom.xml from jdg80 plugin and only changed the jdg81.infinispan.version property part.

plugins/infinispan100/src/main/java/org/radargun/service/Infinispan100ServerService.java
andyuk1986: @gustavolira as there is no Infinispan110ServerService.java class, and the server service is extended from jdg80 plugin, I thought to add it here and added addUser property so that if jdg80 plugin is used it is false by default and no need to add username/password to it.

plugins/jdg81/pom.xml
gustavolira: yes I got it, I think for jdg81 this dependency is no more necessary

andyuk1986: I have removed it Gustavo. The build on my local succeeded.

gustavolira:
LGTM, thanks a lot @andyuk1986

Issue #666, opened 2020-07-01
InfinispanServerClustered should not add an empty node during checkAndUpdateMembership


diegolovison:
It will prevent this kind of error
```
40579, edg-perf04-47890, edg-perf06-25622, edg-perf07-42361, edg-perf08-23107], clusterSize=8}
14:23:22,179 INFO  [org.radargun.Slave] (sc-main) Starting stage ClusterSplitVerify
14:23:22,180 ERROR [org.radargun.stages.monitor.ClusterSplitVerifyStage] (sc-main) Cluster size at the beginning of the test was 8 but changed to 1 during the test! Perhaps a split occured, or a new node joined?
14:23:22,180 INFO  [org.radargun.Slave] (sc-main) Finished stage ClusterSplitVerify
```

PR #667, opened 2020-07-01
Increase the ISPN Rest timeout

diegolovison:
Fix for #666


gustavolira:
Merged #667 into master

PR #669, opened 2020-07-07
Load first the property file

diegolovison:
Load first the property file

Related with ISPN-12096


gustavolira:
Merged #669 into master

PR #675, opened 2020-07-31
Show exception

diegolovison:


plugins/infinispan100/src/main/java/org/radargun/service/InfinispanRestAPI.java
```diff
@@ -79,6 +78,18 @@ CacheManagerInfo getCacheManager() throws RestException {
       return cacheManagerInfo;
    }
 
+   public void stopCluster(long stopTimeout) {
+      try {
+         restClientOkHttp.cluster().stop().toCompletableFuture().get(stopTimeout, TimeUnit.MILLISECONDS);
+      } catch (InterruptedException | ExecutionException | TimeoutException e) {
+         // the stage should be responsible to check if the node still alive
```
gustavolira, line 85: I think you should at least log the exception or propagate the exception

diegolovison: Not needed. We should completely ignore. The errors will be raised when the nodes are down.

gustavolira: I was just trying to avoid https://stackoverflow.com/questions/1234343/why-are-empty-catch-blocks-a-bad-idea

gustavolira:
LGTM, thanks Diego

gustavolira:
Merged #675 into master

PR #676, opened 2020-08-17
Insecure rest

diegolovison:
Insecure rest

### DIFF
--- a/plugins/resteasy-http/src/main/java/org/radargun/http/service/AbstractRESTEasyService.java
+++ b/plugins/resteasy-http/src/main/java/org/radargun/http/service/AbstractRESTEasyService.java
@@ -1,7 +1,9 @@
 package org.radargun.http.service;
 
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.net.InetSocketAddress;
+import java.net.URLEncoder;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -61,6 +63,12 @@ public abstract class AbstractRESTEasyService implements Lifecycle {
    @Property(doc = "Http headers for request. Default is null", complexConverter = KeyValueProperty.KeyValuePairListConverter.class)
    private List<KeyValueProperty> httpHeaders;
 
+   @Property(doc = "The HTTP protocol. Default is http")
+   private String protocol = "http";
+
+   @Property(doc = "Trust all https certificates. Default is false")
+   private boolean trustAll = false;
+
    @ProvidesTrait
    public Lifecycle getLifecycle() {
       return this;
@@ -71,6 +79,10 @@ public abstract class AbstractRESTEasyService implements Lifecycle {
       if (httpClient != null) {
          log.warn("Service already started");
          return;
+      }
+
+      if (trustAll) {
+         HttpsHelper.trustAll();
       }
 
       httpClient = new ResteasyClientBuilder().httpEngine(new URLConnectionEngine())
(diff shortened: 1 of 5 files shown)


plugins/resteasy-http/src/main/java/org/radargun/http/service/RESTEasyCacheInfo.java
```diff
@@ -74,11 +77,11 @@ public long getTotalSize() {
                size = Long.parseLong(responseStr);    //if the call was done to nodejs app, then the long size is returned
 
             } catch (NumberFormatException ex) {
-               //If the call was done to Infinispan Server, then the response is the list of all entries
-               //WARNING: As this way the REST API retrieves all keys from the cache and returns them as a response,
-               //this part of block may have an impact on the performance of the stage which rely on this trait.
-               String[] entries = responseStr.split(System.getProperty("line.separator"));
-               size = entries.length;
+               Matcher matcher = Pattern.compile("\"size\":(\\d+)").matcher(responseStr);
```
gustavolira, line 80: Nitpick: It's more readable and easier to understand if you create a private method with a name which explain what you are doing ou desire here, it's more readable than analyse the pattern, find, parse ...

plugins/resteasy-http/src/main/java/org/radargun/http/service/AbstractRESTEasyService.java
```diff
@@ -61,6 +63,12 @@
    @Property(doc = "Http headers for request. Default is null", complexConverter = KeyValueProperty.KeyValuePairListConverter.class)
    private List<KeyValueProperty> httpHeaders;
 
+   @Property(doc = "If the REST client will ignore the https certificates")
```
Crumby, line 66: Ignoring https certificates and using insecure http protocol are two different things. From my understanding insecure property does the latter while for ignoring https certificates there is `trustAll` bellow.

diegolovison: Maybe change to:
* protocol=(http|https)
* trust-all=(true|false)

WDYT?

Crumby: Sounds good

plugins/resteasy-http/src/main/java/org/radargun/http/service/AbstractRESTEasyService.java
```diff
@@ -73,6 +81,10 @@ public synchronized void start() {
          return;
       }
 
+      if (trustAll) {
+         HttpsHelper.trustAll();
```
Crumby, line 85: Setting trustall globally as default if one of the clients sets it isn't a good design. But probably shouldn't be issue for RadarGun as you are probably not going to run performance tests against cluster with valid and invalid certificats.

plugins/resteasy-http/src/main/java/org/radargun/http/service/RESTEasyCacheInfo.java
```diff
@@ -74,11 +77,11 @@ public long getTotalSize() {
...
+               if (matcher.find()) {
+                  // size:100
```
Crumby, line 82: Is this comment an example of what size can be returned or is 100 magic number that isn't clear by default?

plugins/resteasy-http/src/main/java/org/radargun/http/service/AbstractRESTEasyService.java
diegolovison: It is false by default

plugins/resteasy-http/src/main/java/org/radargun/http/service/RESTEasyCacheInfo.java
diegolovison: Removed the comment

Crumby (approved):

gustavolira:
Merged #676 into master. Thanks Diego

plugins/resteasy-http/src/main/java/org/radargun/http/service/AbstractRESTEasyService.java
Crumby: I didn't mean the property itself but the fact that one client setting the property to true sets static variables in `HttpsURLConnection`:
```
HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());
HttpsURLConnection.setDefaultHostnameVerifier(validHosts);
```

PR #680, opened 2020-11-20
Property file must overwrite what was previous configured.

diegolovison:
The property file has a high priority over other configurations


alanfx:
Looks good, @diegolovison! Merging

PR #685, opened 2021-01-28
Infinispan 12 plugin

diegolovison:


gustavolira:
Merged #685 into master. Thanks Diego

PR #687, opened 2021-03-11
checkAndUpdateMembership should be executed after the start

diegolovison:


gustavolira (approved):

gustavolira:
Merged #687 into master

PR #689, opened 2021-03-12
Thread dump

diegolovison:


core/src/main/java/org/radargun/sysmonitor/ThreadDumpMonitor.java
```diff
@@ -0,0 +1,84 @@
...
+   }
+
+   @Override
+   public synchronized void stop() {
+      super.stop();
+   }
+
+   public void threadDump() {
+      log.info("Generating thread dump for: " + this.workerIndex);
+      String fileName = "thread-dump-worker-" + this.workerIndex + "-" + DATE_FORMATTER.format(LocalDateTime.now()) + ".dump";
```
gustavolira, line 64: String.format

core/src/main/java/org/radargun/sysmonitor/ThreadDumpMonitor.java
```diff
@@ -0,0 +1,84 @@
...
+      BufferedWriter writer = null;
+      try {
+         writer = new BufferedWriter(new FileWriter(fileName, true));
```
gustavolira, line 67: You should use try-with-resources it's more safe and you can delete all the finally block

docs/stages/core.md
```diff
@@ -165,6 +165,9 @@ Allows to invoke JMX-exposed methods and attributes.
 
 ### jvm-monitor-start
 Starts collecting statistics locally on main and each worker node.
+> dump-thread (**optional**) - Dump Threads. The default is false.
```
gustavolira, line 168: can you change all `dump-thread` to `thread-dump`?

core/src/main/java/org/radargun/stages/monitor/MonitorStartStage.java
```diff
@@ -45,6 +46,14 @@
    @InjectTrait
    private InternalsExposition internalsExposition;
 
+   // related to thread dump
+   @Property(doc = "Dump Threads. The default is false.")
```
gustavolira, line 50: nitpick: You can remove all `The default` and use just `Default`

gustavolira (changes requested):

gustavolira:
Merged #689 into master. Thanks Diego!

PR #691, opened 2021-03-17
Added a java class using ProtoStream to be encode and decode into into Protocol Buffers

gustavolira:


core/pom.xml
```diff
@@ -64,6 +65,11 @@
          <artifactId>javax.activation-api</artifactId>
          <version>1.2.0</version>
       </dependency>
+      <dependency>
+         <groupId>org.infinispan.protostream</groupId>
```
diegolovison, line 69: Should it go into the Infinispan plugin?
The same applies to the other changes. We can start the support in `infinispan121` plugin.

diegolovison (changes requested):

core/pom.xml
gustavolira: IMHO this makes more sense to go into extensions/cache since this was only necessary mainly because of `cache:value-generator->cache:single-word`

PR #692, opened 2021-03-25
Bump Data Grid

diegolovison:


Crumby (approved):

plugins/jdg73/pom.xml
```diff
@@ -26,7 +26,7 @@
             </property>
          </activation>
          <properties>
-            <version.infinispan>9.4.18.Final-redhat-00001</version.infinispan>
+            <version.infinispan>9.4.19.Final-redhat-00001</version.infinispan>
```
gustavolira, line 29: The last DG 7.3 is `9.4.21.Final-redhat-00002` which is 7.3.8.GA

PR #704, opened 2021-09-13
Increase timeout to 15 seconds

diegolovison:
Increase timeout to 15 seconds


gustavolira:
Merged #704 into master
Thanks Diego!

PR #706, opened 2021-09-27
Fix path for DG83 resources

diegolovison:


alanfx:
Looks fine, @diegolovison Merging

PR #714, opened 2022-02-22
Infinispan 14 plugin

diegolovison:


gustavolira:
Merged #714 into master. Thanks Diego

PR #715, opened 2022-02-24
UUID keys

diegolovison:


alanfx:
Looks good, @diegolovison!

extensions/cache/src/main/java/org/radargun/stages/cache/test/UUIDKeysSelector.java
```diff
@@ -0,0 +1,29 @@
...
+import org.radargun.stages.test.TestStage;
+
+public class UUIDKeysSelector implements KeySelector {
+
+   public UUIDKeysSelector() {
+   }
+
+   @Override
+   public long next() {
+      return UUID.randomUUID().getMostSignificantBits() & Long.MAX_VALUE;
```
rvansa, line 17: Oh my, folks - don't use `UUID.randomUUID()` in performance tests! This uses secure source of randomness and consumes entropy. Eventually you could run out and get blocked in this call.

Check out https://www.baeldung.com/java-security-egd

diegolovison: How about
```java
long next = ThreadLocalRandom.current().nextLong() & Long.MAX_VALUE;
```

rvansa: Yes, please do that.

PR #719, opened 2022-04-08
Fix source directory to copy JDG Snapshot files

diegolovison:


gustavolira:
Merged #719 into master. Thanks @diegolovison

PR #724, opened 2022-05-11
added lifecycle provider to cacheservice

rigazilla:
this seems to be needed to actually run echo service


diegolovison:
Thanks @rigazilla

PR #726, opened 2022-05-26
Baseline: Java11

diegolovison:


gustavolira:
Merged #726 into master, Thanks Diego

PR #727, opened 2022-07-13
Fix Log

diegolovison:


gustavolira:
Merged #727 into master. Thanks Diego

PR #730, opened 2022-08-02
Update JGroups plugins

diegolovison:


gustavolira:
Merged #730 into Master. Thanks Diego

PR #736, opened 2022-08-26
Shared continue running

diegolovison:


gustavolira:
Merged #736 into Master. Thanks Diego!

PR #747, opened 2022-10-28
Drop outdated plugins

diegolovison:


alanfx:
OK, looks good @diegolovison! Merging

PR #748, opened 2022-11-01
Custom memcached bind port

diegolovison:


gustavolira:
Merged #748 into master. Thanks Diego!

PR #755, opened 2022-12-06
Remove Infinispan82 plugin

diegolovison:


plugins/infinispan90/src/main/java/org/radargun/service/Infinispan70CacheInfo.java
```diff
@@ -0,0 +1,47 @@
...
+import org.infinispan.AdvancedCache;
+import org.infinispan.container.DataContainer;
+import org.infinispan.distribution.ch.ConsistentHash;
+import org.radargun.traits.CacheInformation;
+
+/**
+ * @author Radim Vansa &lt;rvansa@redhat.com&gt;
+ * @author Alan Field &lt;afield@redhat.com&gt;
+ */
+public abstract class Infinispan70CacheInfo extends Infinispan53CacheInfo {
```
alanfx, line 15: It seems kind of confusing to use the `Infinispan70CacheInfo` name in the `infinispan90` plugin code?

plugins/infinispan90/src/main/java/org/radargun/service/Infinispan81EmbeddedContinuousQuery.java
```diff
@@ -0,0 +1,20 @@
+package org.radargun.service;
+
+import org.infinispan.Cache;
+import org.radargun.traits.ContinuousQuery;
+
+/**
+ * @author Vojtech Juranek &lt;vjuranek@redhat.com&gt;
+ */
+public abstract class Infinispan81EmbeddedContinuousQuery implements ContinuousQuery {
```
alanfx, line 9: Same comment here about naming. I know the code is coming from that plugin, but it could be weird for users.

